PR #9, opened 2020-09-08
grafana - add support for prometheus serviceMonitor

smcavallo:
add support for optional prometheus serviceMonitor

This is a pattern that is pretty standard across helm charts.  Some examples:
https://github.com/helm/charts/blob/master/stable/prometheus-redis-exporter/templates/servicemonitor.yaml
https://github.com/helm/charts/blob/master/stable/prometheus-cloudwatch-exporter/templates/servicemonitor.yaml
https://github.com/helm/charts/blob/master/stable/coredns/templates/servicemonitor.yaml
It would be helpful to have this built into the grafana helm chart as well for ease of monitoring.

Signed-off-by: S.Cavallo <smcavallo@hotmail.com>

### DIFF
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -124,6 +124,19 @@ service:
   labels: {}
   portName: service
 
+serviceMonitor:
+  ## If true, a ServiceMonitor CRD is created for a prometheus operator
+  ## https://github.com/coreos/prometheus-operator
+  ##
+  enabled: false
+  path: /metrics
+  #  namespace: monitoring
+  labels: {}
+  interval: 5m
+  scrapeTimeout: 30s
+  scheme: http
+  relabelings: []
+
 extraExposePorts: []
  # - name: keycloak
  #   port: 8080
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -130,7 +130,7 @@ serviceMonitor:
   ##
   enabled: false
   path: /metrics
-  #  namespace: monitoring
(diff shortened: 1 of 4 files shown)


charts/grafana/values.yaml
```diff
@@ -124,6 +124,19 @@ service:
   labels: {}
   portName: service
 
+serviceMonitor:
+  ## If true, a ServiceMonitor CRD is created for a prometheus operator
+  ## https://github.com/coreos/prometheus-operator
+  ##
+  enabled: false
+  path: /metrics
+  #  namespace: monitoring
```
torstenwalter, line 133: Can we set this to `namespace: ""` by default and maybe add a comment above what it does? Or which namespace is used by default.

zanhsieh:
@smcavallo 
This PR need to rebase.

smcavallo:
> 
> 
> @smcavallo
> This PR need to rebase.

rebased, thank you

charts/grafana/values.yaml
smcavallo: > 
> 
> Can we set this to `namespace: ""` by default and maybe add a comment above what it does? Or which namespace is used by default.

@torstenwalter - good catch, thank you.  I've updated the readme to reflect the additions to values.yaml with a description and default value.  If the documentation doesn't make sense please let me know and I can update the descriptions.

torstenwalter:
Looks food to me, but I just helped to migrate this repository from stable/grafana. @zanhsieh @rtluckie @maorfr can one of you review?

zanhsieh (approved):

PR #31, opened 2020-09-25
Fix condition for secret checksum to match that in secret.yaml

isatokunaga:
Signed-off-by: Isa Tokunaga <isa.tokunaga@thoughtworks.com>

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.6.11
+version: 5.6.12
 appVersion: 7.2.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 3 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.6.8
+version: 5.6.9
```
zanhsieh, line 3: Need to advance to 5.6.10.

isatokunaga: Done. Thanks!

torstenwalter:
Looks good to me as condition in secret.yaml is the same:

`{{- if or (and (not .Values.admin.existingSecret) (not .Values.env.GF_SECURITY_ADMIN_PASSWORD__FILE) (not .Values.env.GF_SECURITY_ADMIN_PASSWORD)) (and .Values.ldap.enabled (not .Values.ldap.existingSecret)) }}
`

torstenwalter (approved):

PR #46, opened 2020-10-07
test: add test for enabled image-renderer

GabrielCastro:


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.2
+version: 5.7.3
```
zanhsieh, line 3: Need to bump 5.7.4

zanhsieh (approved):

PR #45, opened 2020-10-07
Added resources for imageRenderer

odavid:
Signed-off-by: odavid <ohad.david@gmail.com>

PR #47, opened 2020-10-08
Fix indentation of #45

odavid:
Thank you for merging #45, however, I found a bug in the resource indentation.
This PR fixes it.

Sorry for the bug

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.6
+version: 5.7.7
 appVersion: 7.2.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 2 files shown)


torstenwalter:
Linting error should be fixed with #48 This is unrelated to this change.

GabrielCastro:
I've fixed this along side a couple of other errors in `imageRenderer` in #46 where I also added tests for it so hopefully it's not missed in the future

odavid:
Thanks @GabrielCastro - I also fixed the tolerations and nodeSelector indentation in this PR

charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.3
+version: 5.7.4
```
zanhsieh, line 3: should now 5.7.5.

odavid: Hi @zanhsieh - rebased and changed the version

zanhsieh (approved):

PR #49, opened 2020-10-10
[grafana] Update kiwigrid/k8s-sidecar

mrueg:
This adds support for arm64/arm images as well.

https://hub.docker.com/layers/kiwigrid/k8s-sidecar/0.1.209/

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.4
+version: 5.7.6
 appVersion: 7.2.0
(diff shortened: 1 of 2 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.3
+version: 5.7.4
```
zanhsieh, line 3: I think this later shall be 5.7.6

mrueg: Thanks, updated.

zanhsieh (approved):

PR #50, opened 2020-10-11
AWS EKS support for cloudwatch with an example in values.yaml

phntom:
AWS EKS support for cloudwatch with an example in values.yaml

Background:

I'm using AWS EKS and CloudWatch for some of my metrics in kube-prometheus-stack (previously known as prometheus-operator)

This is my minimal patch and example configuration with a howto guide on using EKS iam roles for service accounts to gain access to cloudwatch metrics.

If this patch is merged it will save me some effort next time i need to upgrade

This is not a breaking change although it adds new functionality in form of the projected field.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.7
+version: 5.7.8
 appVersion: 7.2.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.6
+version: 5.7.7
 appVersion: 7.2.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.4
+version: 5.7.6
 appVersion: 7.2.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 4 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.7.6
+version: 5.7.7
```
zanhsieh, line 3: Should be 5.7.8

phntom: updated

charts/grafana/values.yaml
```diff
@@ -284,6 +285,27 @@ admin:
...
+## to provide grafana with access to CloudWatch on AWS EKS:
+## 1. create an iam role of type "Web identity" with provider oidc.eks.* (note the provider for later)
+## 2. edit the "Trust relationships" of the role, add a line inside the StringEquals clause using the
+## same oidc eks provider as noted before (same as the existing line)
+## also, replace NAMESPACE and prometheus-operator-grafana with the service account namespace and name
+##
+##  "oidc.eks.us-east-1.amazonaws.com/id/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX:sub": "system:serviceaccount:NAMESPACE:prometheus-operator-grafana",
+##
+## 3. attach a policy to the role, you can use a built in policy called CloudWatchReadOnlyAccess
+## 4. use the following env: (replace 123456789000 and iam-role-name-here with your aws account number and role name)
```
GabrielCastro, line 298: This is handled by the pod admission hook automatically if it's setup only requiring the annotation to be present on the service account. I think this comment should mention these steps aren't commonly required.

maybe even link to the AWS eks docs

phntom: I agree there are alternatives,
steps 1+2+3 can be replaced with an eksctl command
and steps 4+5 can be replaced with an admission hook if it's installed

I do feel these instructions should be present somewhere in their current form.

I'm okay with moving these notes to some other section/file if you feel values.yaml is not the place for this, possibly they'll be better in the service account section or a separate doc file.

I do believe these are important, mostly since a lot of EKS users don't have the permissions to create IAM roles themselves and admission hooks are not installed in vanilla EKS and sometimes aren't an option (for security/auditing reasons).

zanhsieh:
@phntom 
Please submit another PR for documentation update as @GabrielCastro stated.

phntom:
Will do, I intend to remove the instruction steps and replace it with a link to AWS docs on eksctl, irsa docs and admission hooks.

Issue #10, opened 2020-09-11
New pod initialised on helm upgrade even without any changes


Fuuzetsu:
Any time I run `helm upgrade` even without touching the config, the grafana pod is torn down and new one is started in its place. If I run `helm upgrade` 5 times back-to-back, it goes through 5 pods. This seems strange.

I'm not sure what kind of information would help, please let me know.

I have a very basic setup. In Chart.yaml `dependencies` I have

```yaml
  - name: grafana
    version: 5.6.2
    repository: https://grafana.github.io/helm-charts
  - name: prometheus
    version: 11.15.0
    repository: https://prometheus-community.github.io/helm-charts
```

and in values.yaml I have
```
grafana:
  datasources:
    datasources.yaml:
      apiVersion: 1
      datasources:
      - name: Prometheus
        type: prometheus
        url: …
        access: proxy
        isDefault: true
  dashboardProviders:
   dashboardproviders.yaml:
     apiVersion: 1
     providers:
     - name: default
       orgId: 1
       folder: ""
       type: file
       disableDeletion: false
       editable: true
       options:
         path: /var/lib/grafana/dashboards/default
  dashboards:
    default:
     kubernetes-cluster-monitoring-prometheus:
       gnetId: 315
       datasource: Prometheus
```

Fuuzetsu:
After some further searching, it seems the problem is https://github.com/helm/charts/issues/7891 and the linked https://github.com/helm/helm/issues/3053

Fuuzetsu:
I have looked at the chart itself and indeed it is this:
```
  {{- if .Values.adminPassword }}
  admin-password: {{ .Values.adminPassword | b64enc | quote }}
  {{- else }}
  admin-password: {{ randAlphaNum 40 | b64enc | quote }}
  {{- end }}
```

`adminPassword` is not documented in README as something one can set though I see it in `values.yaml`. Setting this to some dummy value no longer re-provisions the service.

I think to resolve issue, these things would be nice:

* Document `adminPassword` but point to `admin.existingSecret` as the proper way to set user/password.
* Secondly, on `adminPassword` document that if left unset, a random password will be generated (visible in k8s secrets post-deploy) and every subsequent `helm upgrade` will cause the password to be regenerated and fresh instance deployed onto a new pod.
* On `admin.existingSecret` mention that if it is unset, then `adminPassword` is used.

zanhsieh:
@Fuuzetsu 
Would you PR it?

[...]

PR #62, opened 2020-10-20
[grafana] Prevent admin password from changing every upgrade

0xMH:
Solving #10 
Using a lookup function to check if the password secret exists or not before generating new one, If It's indeed exit It'll be reused. 

Lookup functions are [first introduced in helm3](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function) so By introducing a new lookup function this chart is pretty much incompatible with Helm2 But at this point of time, I'm not really sure that this is a problem cause 
[On November 13, 2020, Helm v2 support will formally end](https://github.com/helm/charts#status-of-the-project)

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.9.2
+version: 6.9.3
 appVersion: 7.5.5
 kubeVersion: '^1.8.0-0'
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_helpers.tpl
+++ b/charts/grafana/templates/_helpers.tpl
@@ -114,3 +114,16 @@ Return the appropriate apiVersion for rbac.
 {{- print "rbac.authorization.k8s.io/v1beta1" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Looks if there's an existing secret and reuse its password. If not it generates
+new password and use it.
+*/}}
+{{- define "grafana.password" -}}
+{{- $secret := (lookup "v1" "Secret" (include "grafana.namespace" .) (include "grafana.fullname" .) ) -}}
+  {{- if $secret -}}
+    {{-  index $secret "data" "admin-password" -}}
+  {{- else -}}
+    {{- (randAlphaNum 40) | b64enc | quote -}}
+  {{- end -}}
+{{- end -}}
--- a/charts/grafana/templates/secret.yaml
+++ b/charts/grafana/templates/secret.yaml
@@ -17,7 +17,7 @@ data:
   {{- if .Values.adminPassword }}
   admin-password: {{ .Values.adminPassword | b64enc | quote }}
   {{- else }}
-  admin-password: {{ randAlphaNum 40 | b64enc | quote }}
+  admin-password: {{ template "grafana.password" . }}
   {{- end }}
   {{- end }}
   {{- if not .Values.ldap.existingSecret }}


zanhsieh:
@0xMH 
Could you bump the chart version to latest please? Thank you.

0xMH:
@zanhsieh Can you check now please? Any other changes required?

charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 5.8.10
+version: 5.8.11
```
zanhsieh, line 3: Actually the version shall be 5.8.13 or larger.

0xMH: Done

0xMH:
@zanhsieh I increased the version numb again. Any more changes?

zanhsieh (approved):

charts/grafana/templates/_helpers.tpl
```diff
@@ -100,3 +100,16 @@ Selector labels ImageRenderer
 app.kubernetes.io/name: {{ include "grafana.name" . }}-image-renderer
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end -}}
+
+{{/*
+Looks if there's an existing secret and reuse its password. If not it generates
+new password and use it.
+*/}}
+{{- define "grafana.password" -}}
+{{- $secret := (lookup "v1" "Secret" (include "grafana.namespace" .) (include "grafana.fullname" .) ) -}}
```
zanhsieh, line 109: ==> Linting charts/grafana
Error:  templates/: parse error at (grafana/templates/_helpers.tpl:109): function "lookup" not defined

torstenwalter:
@0xMH DCO check is failing (https://github.com/grafana/helm-charts/runs/1302545388) and also linting fails

0xMH:
@torstenwalter The Linting error is kinda expected because the lookup func is addded on Helm3. I think we should consider to drop support form helm 2 as its EOF is already in the really near future.

charts/grafana/templates/_helpers.tpl
0xMH: Lookup func is introduced on Helm v3

zanhsieh:
@0xMH 
Please rebase after https://github.com/grafana/helm-charts/pull/69 merged. Thank you.

@torstenwalter @maorfr 
Please help out with https://github.com/grafana/helm-charts/pull/69 .

zanhsieh:
@0xMH 
https://github.com/grafana/helm-charts/pull/69 has been merged. please rebase from there. Thank you and sorry for waiting.

0xMH:
Done. Can you check now please? @zanhsieh

torstenwalter:
@0xMH Build is failing. Can you try to specify an `image` for https://github.com/grafana/helm-charts/blob/main/.github/workflows/lint-test.yaml#L17

I think `quay.io/helmpack/chart-testing:v3.3.0` contains an updated helm version.

0xMH:
@torstenwalter Done.

.github/workflows/lint-test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
 
       - name: Run chart-testing (lint)
         id: lint
-        uses: helm/chart-testing-action@v1.0.0
+        uses: quay.io/helmpack/chart-testing:v3.3.0
```
torstenwalter, line 17: This does not work. We have to use the chart-testing-action. Internally it's using the chart-testing image. If we want to specify a specific version then we need to do that via parameters.

torstenwalter (dismissed):

.github/workflows/lint-test.yaml
0xMH: @torstenwalter If we're going to migrate to helm3 shouldn't we start using [chart-testing-action@v3.0.0](https://github.com/marketplace/actions/helm-chart-testing-helm-3#example-workflow)should I update it?

0xMH:
I'm not sure why exactly is fails now. The linter give me `Error:  templates/: parse error at (grafana/templates/_helpers.tpl:118): unexpected <define> in command`

sathieu:
Not stale

sathieu:
@0xMH @zanhsieh @torstenwalter Any news on this PR?

charts/grafana/templates/_helpers.tpl
```diff
@@ -113,4 +113,16 @@ Return the appropriate apiVersion for rbac.
 {{- else -}}
 {{- print "rbac.authorization.k8s.io/v1beta1" -}}
 {{- end -}}
```
steven-sheehy, line 115: This is the reason for the lint errors `unexpected <define> in command`. The previous define is missing its final end block.

```suggestion
{{- end -}}
{{- end -}}
```

steven-sheehy (changes requested):
Need to fix the lint and bump the chart version

0xMH:
I did the required changes, Hopefully we could merge now?

steven-sheehy (approved):
LGTM though I don't have commit rights.

zanhsieh (approved):

PR #63, opened 2020-10-22
Add runAsNonRoot to false for initChown

StrongMonkey:
Specifically set runAsNonRoot to false for initChown container so that
it can always run as root. This is to avoid any pod.securityContext overriding
runAsNonRoot.

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -27,6 +27,7 @@ initContainers:
     {{- end }}
     imagePullPolicy: {{ .Values.initChownData.image.pullPolicy }}
     securityContext:
+      runAsNonRoot: false
       runAsUser: 0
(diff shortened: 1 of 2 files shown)


StrongMonkey:
Related issue: https://github.com/rancher/rancher/issues/29639

charts/grafana/templates/_pod.tpl
```diff
@@ -27,6 +27,7 @@ initContainers:
     {{- end }}
     imagePullPolicy: {{ .Values.initChownData.image.pullPolicy }}
     securityContext:
+      runAsNonRoot: false
       runAsUser: 0
```
cbron, line 31: Weird, wouldn't `runAsUser` handle this for you ? Or do you need both to enable root user.

zanhsieh:
@StrongMonkey 
You need to bump the chart version to latest. Without it, this PR can't merge.

StrongMonkey:
@zanhsieh Chart version is updated.

zanhsieh (approved):

PR #69, opened 2020-10-28
Helm3 support

zanhsieh:
### Testing
```
$ kubectl get po
NAME                      READY   STATUS    RESTARTS   AGE
grafana-6c5596bb7-bg2sg   1/1     Running   0          16m

$ helm list
NAME   	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART        	APP VERSION
grafana	default  	1       	2020-10-28 12:03:10.848532 +0800 HKT	deployed	grafana-5.9.0	7.2.1

$ helm version
version.BuildInfo{Version:"v3.4.0", GitCommit:"7090a89efc8a18f3d8178bf47d2462450349a004", GitTreeState:"dirty", GoVersion:"go1.15.3"}
```

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 5.9.0
+version: 6.0.0
 appVersion: 7.2.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
-apiVersion: v1
+apiVersion: v2
 name: grafana
-version: 5.8.16
+version: 5.9.0
 appVersion: 7.2.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
@@ -18,4 +18,4 @@ maintainers:
   - name: Xtigyro
     email: miroslav.hadzhiev@gmail.com
 engine: gotpl
-tillerVersion: ">=2.12.0"
+type: application
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -1,3 +1,5 @@
+autoscaling:
+  enabled: false
 rbac:
   create: true
   ## Use an existing ClusterRole/Role (depending on rbac.namespaced false/true)
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -1,5 +1,3 @@
-autoscaling:
-  enabled: false
 rbac:
   create: true
   ## Use an existing ClusterRole/Role (depending on rbac.namespaced false/true)
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -42,6 +42,10 @@ This version requires Helm >= 2.12.0.
 
 You have to add --force to your helm upgrade command as the labels of the chart have changed.
 
+### To 6.0.0
+
+This version requires Helm >= 3.4.0.
+
 ## Configuration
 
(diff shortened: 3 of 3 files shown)


maorfr (approved):

charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
-apiVersion: v1
+apiVersion: v2
 name: grafana
-version: 5.8.16
+version: 5.9.0
```
torstenwalter, line 3:
```suggestion
version: 6.0.0
```
Let's increment the major version for this.

charts/grafana/values.yaml
```diff
@@ -1,3 +1,5 @@
+autoscaling:
+  enabled: false
```
torstenwalter, line 2:
```suggestion
```

I could not find this value used anywhere.

torstenwalter (commented):
I also suggest to add an entry in README.md similar to this one: https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#to-400-and-3121

### To 6.0.0

This version requires Helm >= 3.0.0.

torstenwalter (approved):
lgtm

torstenwalter:
@zanhsieh I leave merging up to you. For me this looks good.

Xtigyro:
@zanhsieh @maorfr @torstenwalter 
Hey guys - is there a particular reason why we just don't leave the compatibility with Helm `v2`?

torstenwalter:
@Xtigyro https://github.com/grafana/helm-charts/pull/62 is an example which needs helm 3.

According to https://helm.sh/blog/helm-v2-deprecation-timeline/ helm 2 will soon receive no security patches anymore so people are encouraged to migrate to helm 3 anyhow.

A migration guide is also available https://helm.sh/docs/topics/v2_v3_migration/

Xtigyro:
@torstenwalter Thanks for the reply! 

I understand. I wanna mention though that I'm yet to see a big company migrate to `v3`. I know even companies where some of the Helm maintainers work which still use `v2`.

I like having two Helm versions support. Some changes in Helm `v4` will cut off directly any attempts at supporting `v2` - which I agree. The Helm folks created `v3` so that people can use `v2` if they want. 

So - I'm just fearing that people will start working this PR around (that even still includes myself, regrettably).

torstenwalter:
It's sad to see that companies are still using helm 2. Especially if one considered presence of tiller and security risks attached.

Migration to helm 3 is not a big deal as it's mainly client side and does not require new CRDs on the server. So one could easily migrate individual charts or use it just in namespaces.

I know enterprises which migrated to helm 3 a while ago without issues ;-)

Xtigyro:
@torstenwalter Totally agree with ya!

Is putting `if` clauses in the code that check which Helm version is in use for certain features to be enabled an option - or it's too time-consuming to be maintained?

torstenwalter:
Not sure is the helm version available for templates at all? https://helm.sh/docs/topics/charts/ shows Kubernetes version.

Xtigyro:
Yeah - seems not: https://helm.sh/docs/chart_template_guide/builtin_objects/
Ah - case closed. 

Thanks for your time, @torstenwalter.

zanhsieh:
@Xtigyro 
Actually there's a PR https://github.com/grafana/helm-charts/pull/62 caught my attention. The problem was 'lookup' call in the template. We can't just put on some if statement to get both version compatible.

Xtigyro:
@zanhsieh Yes, agreed - thank you!

desaintmartin:
What changes make this compatible only with Helm >= v3.4.0 and not previous Helm 3 versions?

zanhsieh:
@desaintmartin 
Yes.

desaintmartin:
I mean, I have been able to install, lint and `template --validate` the chart using helm 3.3.4, is there a feature that requires helm >=3.4?

PR #73, opened 2020-10-30
bump appversion and image to 7.3.1

kubicgruenfeld:
This PR bumps the app- and image version to grafana 7.3.1

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 6.0.1
-appVersion: 7.2.1
+version: 6.1.0
+appVersion: 7.3.1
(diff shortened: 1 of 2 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 6.0.1
-appVersion: 7.2.1
+version: 6.0.2
```
Skaronator, line 3: Shouldn't we bump the minor as well?
```suggestion
version: 6.1.0
```

kubicgruenfeld: agree and done

Skaronator (approved):

torstenwalter (approved):

PR #92, opened 2020-11-20
Update tools versions in GitHub workflow

unguiculus:
This updates GitHub actions and tools used in GitHub workflows to the
latest versions.

See https://github.com/helm/charts-repo-actions-demo for reference.

This update adds support for GPG signing which will be required for the
planned addition of Loki charts. See grafana/loki/issues/2593

Signed-off-by: Reinhard Nägele <unguiculus@gmail.com>

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -8,23 +8,38 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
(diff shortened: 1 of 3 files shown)


unguiculus:
cc @torstenwalter

torstenwalter (approved):
LGTM - verified that it works with https://github.com/grafana/helm-charts/pull/93/checks?check_run_id=1432718817

PR #105, opened 2020-12-03
fix: Fix network communication with ImageRenderer so it works with the

AaronKalair:
Hello 👋 

I was using your HelmChart today to install the ImageRenderer plugin and experienced some confusion when my Grafana instance couldn't talk to the imageRenderer.

It seems the default values.yaml Values don't provide a working setup, I've tweaked them to fix that.

Thanks!

### DIFF
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -633,7 +633,8 @@ imageRenderer:
     # image-renderer ImagePullPolicy
     pullPolicy: Always
   # extra environment variables
-  env: {}
+  env:
+    HTTP_HOST: "0.0.0.0"
     # RENDERING_ARGS: --disable-gpu,--window-size=1280x758
     # RENDERING_MODE: clustered
   # image-renderer deployment securityContext
@@ -647,7 +648,7 @@ imageRenderer:
     portName: 'http'
     # image-renderer service port used by both service and deployment
     port: 8081
-    targetPort: 3001
+    targetPort: 8081
   # name of the image-renderer port on the pod
   podPortName: http
   # number of image-renderer replica sets to keep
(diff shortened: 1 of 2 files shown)


charts/grafana/values.yaml
```diff
@@ -633,7 +633,8 @@ imageRenderer:
     # image-renderer ImagePullPolicy
     pullPolicy: Always
   # extra environment variables
-  env: {}
+  env:
+    HTTP_HOST: "0.0.0.0"
```
AaronKalair, line 637: It listens on `localhost` by default so you can't reach it from a different pod

charts/grafana/values.yaml
```diff
@@ -647,7 +648,7 @@ imageRenderer:
     portName: 'http'
     # image-renderer service port used by both service and deployment
     port: 8081
-    targetPort: 3001
+    targetPort: 8081
```
AaronKalair, line 651: It listens on port `8081` not `3001` so previously this just gave `Connection refused` errors

zanhsieh:
@AaronKalair 
Just curious what you intend to achieve. You supposed to change your own PR branch without any restriction.

AaronKalair:
> @AaronKalair
> Just curious what you intend to achieve. You supposed to change your own PR branch without any restriction.

Hey,

Thanks for taking a look at this, I wanted to provide the chart with a set of default values that produced a working deployment of the imageRenderer.

Currently this isn't the case as 

1. The imageRenderer only listens on `localhost` so you can' talk to it from outside the container
2. The targetPort is pointing at a port that the imageRenderer isn't listening on, so even if you fix 1. you still can't communicate with the imageRenderer via the Kubernetes Service

zanhsieh (approved):

PR #115, opened 2020-12-09
Turn on GPG signing

unguiculus:
In order to be able to test it, the Grafana chart version is bumped.


unguiculus:
cc @torstenwalter

torstenwalter (approved):

PR #118, opened 2020-12-09
Add loki-distributed and loki-canary charts

unguiculus:
* Merges in the charts from https://github.com/unguiculus/loki-helm-chart 
  including the full Git history.
* Configures helm-docs for loki-charts


torstenwalter:
Can you split this up into separate PRs?

- changes to grafana chart
- adding new loki charts
- we should not check docs unless all charts are doing it or only check the charts you are adding

PR #123, opened 2020-12-09
Add helm-docs

unguiculus:
loki-canary and loki-distributed charts, which are added in separate
PRs, use helm-docs to generate README.md files. Charts that don't use
it yet must be added to .helmdocsignore.


torstenwalter (approved):

torstenwalter:
LGTM as charts which are not using it are ignored.

PR #135, opened 2020-12-15
[loki-distributed] Add clusterDomain and dnsService config options

piontec:
This allows configuring 2 options:

- `clusterDomain` nor all clusters use `cluster.local`
- `dnsService` - not all clusters have the default DNS service under this name, we have some clusters that go with `coredns`

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.1.0
-version: 0.22.2
+version: 0.23.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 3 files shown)


unguiculus (changes requested):
Thanks for the PR. There's a few things you'll have to do:
* Bump the minor chart version
* Fix DCO
* Update the readme running `docker run --rm --volume "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.4.0`

charts/loki-distributed/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.1.0
-version: 0.22.1
+version: 0.22.2
```
unguiculus, line 6: As mentioned, please bump the minor version. I consider this a new feature, not a patch.
```suggestion
version: 0.23.0
```

unguiculus (changes requested):

unguiculus (approved):

PR #148, opened 2020-12-23
[loki-distributed] Add extraVolumes and extraVolumeMounts in global

bangau1:
This PR proposes to add new 2 fields in `global`:
- `extraVolumes`
- `extraVolumeMounts`.

Adding those 2 fields so that I can inject new volume (and mount it) to the all containers. In my use case, I need them so I can deploy loki with GCS storage, where it needs GCP Service Account secret file.

The way I use is like this:
```
global:
  extraVolumes:
  - name: secret # to hold the Loki GCP service account to access gcs bucket
    secret:
      secretName: ${gcp_sa_secret_name}

  extraVolumeMounts:
  - name: secret
    mountPath: /secret/loki/

```

Then for each loki component, I add `extraEnvs`:
```
...
  extraEnv:
  - name: GOOGLE_APPLICATION_CREDENTIALS 
    value: /secret/loki/credentials.json
...
```

Please help review. 
Thanks in advance.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.0.0
-version: 0.20.1
+version: 0.20.2
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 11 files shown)


unguiculus (changes requested):
Thanks for the PR. I don't think this is something that should be done on a global level. Not all components need the credentials. However, I do see the need for supporting additional volumes.

unguiculus:
Closing this in favor of #160. Thanks.

PR #149, opened 2020-12-24
Update charts for Loki 2.1.0

slim-bean:
This should update everything to the latest Loki release


slim-bean:
going to open up new PR's, one for each chart

slim-bean:
on second thought...

slim-bean:
I'll have to put some thought into how to handle this better for a future release, for now I'm going to cheat a little and force the merge because I can override the requirement that a PR does not cover multiple charts

joe-elliott (approved):
Merry Christmas!  Ho Ho Ho!

PR #151, opened 2020-12-28
[loki-distributed] Change the `ServiceAccount`'s name to use the correct template

AmitBenAmi:
Using the correct template for the `ServiceAccount` name in `loki-distributed`


unguiculus (approved):
Good catch. Thank you.

unguiculus (dismissed):

unguiculus (approved):

Issue #131, opened 2020-12-11
Unable to install loki stack when grafana is disabled


adnan-kamili:
I have loki stack insatalled in my k8s cluster and it works fine. Here is the config and install command:

```
helm repo add loki https://grafana.github.io/loki/charts
helm repo update
helm upgrade --install loki loki/loki-stack --atomic --create-namespace --namespace cluster-management --values ./loki/loki.yaml
```

And loki.yaml

```
loki:
  persistence:
    enabled: true
    size: 20Gi
  config:
    table_manager:
      retention_deletes_enabled: true
      # retain logs for two weeks
      retention_period: 336h

promtail:
  enabled: false

fluent-bit:
  enabled: true

grafana:
  enabled: false

prometheus:
  enabled: false
```

This worked fine with 1.x but fails with 2.x for the fluentbit with following error:

```
2020-12-11T07:09:02.805142608Z Fluent Bit v1.4.6
2020-12-11T07:09:02.805165872Z * Copyright (C) 2019-2020 The Fluent Bit Authors
2020-12-11T07:09:02.805172280Z * Copyright (C) 2015-2018 Treasure Data
2020-12-11T07:09:02.805175930Z * Fluent Bit is a CNCF sub-project under the umbrella of Fluentd
2020-12-11T07:09:02.805179296Z * https://fluentbit.io
2020-12-11T07:09:02.805182583Z 
2020-12-11T07:09:02.805185697Z Output plugin 'grafana-loki' cannot be loaded
2020-12-11T07:09:02.805189030Z Error: You must specify an output target. Aborting
```

We already have grafana installed so we don't want to re-install in the loki-stack

utkudarilmaz:
Same issue occurred when installing loki/fluent-bit v2.0.1 chart.

knfoo:
I just tried to install it via the loki-stack

`helm upgrade --install loki grafana/loki-stack --set fluent-bit.enabled=true,promtail.enabled=false,grafana.enabled=true,prometheus.enabled=true,prometheus.alertmanager.persistentVolume.enabled=false,prometheus.server.persistentVolume.enabled=false --namespace loki`

I get the same error

utkudarilmaz:
If change the output plugin name to `loki` error is resolving.

ohdearaugustin:
Same issue here with loki-stack version 2.2.0.

unguiculus:
[...]

PR #154, opened 2020-12-30
fixes output plugin name

eiser82:
after deploying the loki-stack helm chart into k8s using fluent-bit as the log exporter, the fluent-bit pod fails into a crash-loop because it can't find the `grafana-loki` output. Ideally this should be turned into a variable but for now this change should fix the problem.


zanhsieh:
@eiser82 
Could you bump the chart version to latest and fix DCO first please?

unguiculus:
With Loki 2.1.0, the issue should be fixed. No need to revert it back. See https://github.com/grafana/loki/releases/tag/v2.1.0

ohdearaugustin:
Please take a look at #131

eiser82:
Thanks for the quick response, I will try 2.1.0 and test

unguiculus:
Closing this, especially since #155 was merged.

PR #160, opened 2021-01-06
[loki-distributed] Add capability to add Volumes and Volume mounts

Whyeasy:
**Description of changes**

Add the capability to add Volumes and Volume mounts to the needed services. This is needed to be able to mount the GCP service account for example.

**Benefits**

Able to mount the GCP service account json file to the needed services.

**Additional Information**

It's the same capability as PR https://github.com/grafana/helm-charts/pull/148 but not a global level.


unguiculus (changes requested):
Thanks. Please bump the minor version for a new feature.

unguiculus (approved):

PR #161, opened 2021-01-06
[loki-distributed] Add `ServiceMonitor` to `gateway` component

AmitBenAmi:
- Adding `ServiceMonitor` to `loki-gateway`
- Bumping the chart version

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.1.0
-version: 0.21.1
+version: 0.21.2
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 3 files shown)


unguiculus (commented):
That's interesting. Currently, the gateway does not expose any metrics. Why would you need that?

However, that brings me to the idea that it could be a good idea to also use the gateway to proxy metrics endpoints. That would be beneficial for network policies.

AmitBenAmi:
Do you mean to provide a single endpoint for all of the metrics of the components?
What would then be in the NetworkPolicy? Only for Prometheus/other clients?

unguiculus:
No, I thought about separate endpoints for each components. That was just a thought. I guess it's not a good idea because it would actually have to be per pod.

Anyways, why do you need a service monitor for the gateway? We'd have to add an exporter for Nginx in the first place, I guess?

AmitBenAmi:
I actually have metrics in loki `v1.3.0` from `gateway`, but also there I can't find the `/metrics` endoint of nginx.
I guess it is not that important. I'll close the PR

PR #170, opened 2021-01-08
[grafana] add sidecar configmap label value select

ryanprobus:
* Allows specification of the LABEL_VALUE environment variable on the
  kiwigrid/k8s-sidecar as labelValue helm value for the datasource and
  dashboard sidecars.

Signed-off-by: Ryan Probus <ryan.probus@appian.com>

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.1.17
+version: 6.2.0
 appVersion: 7.3.5
 kubeVersion: '^1.8.0-0'
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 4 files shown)


Xtigyro (changes requested):
Hey @ryanprobus - thanks for the contribution!
What added functionality does `labelValue` bring? Why `label` is not enough?

ryanprobus:
@Xtigyro While the intended functionality could be done with only `label`, that requires adding a label key that is non-standardized between different deployments. It would be nice if we could use a common label key and differentiate by the label value on which deployments should have which dashboards. Thought it would be nice to have the option of doing it this way.

Xtigyro (approved):

Xtigyro:
@ryanprobus OK - sounds useful. Thanks!

@zanhsieh @torstenwalter @maorfr Hmm, there's an issue with the code owners - I'm not picked as a reviewer.

torstenwalter (approved):

PR #174, opened 2021-01-15
[promtail] Refactor chart

unguiculus:
This refactors the Promtail chart for better configurability and aligns
it to current best practices.

Notable changes:

* Helm 3 is required
* Labels are updated to follow the official Kubernetes
  [label recommendations](kubernetes.io/docs/concepts/overview/working-with-objects/common-labels)
* The config file must be specified as string which can be templated
* The default scrape configs have been updated to take new and old labels
  into consideration
* The config file is now stored in a Secret and no longer in a ConfigMap
  because it may contain sensitive data, such as basic auth credentials
* The readme is now generated using helm-docs

This PR contains breaking changes and, thus, bumps the major version of
the chart. Directions on how to upgrade are in the readme.


owen-d (approved):
Wow, this looks great. I have no qualms - glad to see the new version.

scottrigby:
@unguiculus very nice example of updating a chart to current best practices  👏 

One thing I might suggest is removing the "## Values" section from the README, since they often (eventually) get out of sync.

At the top of the "Configuration" section, there could be something like what we added to the prometheus charts:

        See [Customizing the Chart Before Installing](https://helm.sh/docs/intro/using_helm/#customizing-the-chart-before-installing).
        To see all configurable options with detailed comments, visit the chart's [values.yaml](./values.yaml), or run these configuration commands:
        
        ```console
        # Helm 2
        $ helm inspect values prometheus-community/prometheus
        
        # Helm 3
        $ helm show values prometheus-community/prometheus
        ```
        
        You may similarly use the above configuration commands on each chart [dependency](#dependencies) to see it's configurations.

unguiculus:
@scottrigby The values section is generated using helm-docs.

PR #180, opened 2021-01-19
[loki-distributed] Add extraVolumes and extraVolumeMounts to gateway

PierreNowak:
Mount custom volumes in gateway such as ssl certs


unguiculus (approved):
Makes sense. Thanks.

PR #182, opened 2021-01-20
[promtail] Fix scrape configs

unguiculus:
The default config was missing a few drop actions. This had caused logs
to be added twice, once the correct way and once with an incorrect job
label.

In order to help debug the config, the flag
`config.snippets.addScrapeJobLabel` is introduced. If enabled, an
additional label `scrape_job` is added which identifies the scrape job
that picked up the log message.

### DIFF
--- a/charts/promtail/values.yaml
+++ b/charts/promtail/values.yaml
@@ -264,6 +264,11 @@ config:
           - __meta_kubernetes_pod_uid
           - __meta_kubernetes_pod_container_name
         target_label: __path__
+
+    # If set to true, adds an additional label for the scrape job.
+    # This helps debug the Promtail config.
+    addScrapeJobLabel: false
+
     scrapeConfigs: |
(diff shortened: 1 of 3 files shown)


charts/promtail/values.yaml
```diff
@@ -278,10 +283,19 @@ config:
             source_labels:
               - __meta_kubernetes_pod_label_app_kubernetes_io_name
             target_label: app
+          - action: drop
+            regex: ''
+            source_labels:
+              - app
```
torstenwalter, line 289: are these changes intended? I don't see how this relates to the `addScrapeJobLabel` option.

unguiculus: Yes, that's what this PR is mostly about. We need to drop out here when there is no app label in order to avoid duplicates with an incomplete job label. These pods will be covered by one of the other scrape jobs.

That's not related to `addScrapeJobLabel`.

torstenwalter (approved):

PR #189, opened 2021-01-22
[promtail] Add more customizations to values.yaml

piontec:
This PR adds the following options, which were available previously:
- add custom keys to the `client` section of the config file
- add extra scrape configs directly (to allow for journal or static scrapping)

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.1.0
-version: 3.0.4
+version: 3.0.5
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.1.0
-version: 3.0.5
+version: 3.1.0
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
--- a/charts/promtail/values.yaml
+++ b/charts/promtail/values.yaml
@@ -214,14 +214,6 @@ podSecurityPolicy:
   requiredDropCapabilities:
     - ALL
 
-# -- You can put here any keys that will be directly added to the config file's 'client' block.
-# @default -- empty
-extraClientConfigs: {}
-
-# -- You can put here any additional scrape configs you want to add to the config file.
-# @default -- empty
-extraScrapeConfigs: {}
-
 # -- Section for crafting Promtails config file. The only directly relevant value is `config.file`
 # which is a templated string that references the other values and snippets below this key.
 # @default -- See `values.yaml`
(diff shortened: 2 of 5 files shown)


charts/promtail/values.yaml
```diff
@@ -214,6 +214,14 @@ podSecurityPolicy:
   requiredDropCapabilities:
     - ALL
 
+# -- You can put here any keys that will be directly added to the config file's 'client' block.
+# @default -- empty
+extra_client_configs: {}
+
+# -- You can put here any additional scrape configs you want to add to the config file.
+# @default -- empty
+extra_scrape_configs: {}
```
unguiculus, line 223: If we add something like this, then it must go under `config.snippets`. Also: camel case please.

unguiculus (changes requested):
Something like this is already possible. Have a look at the values file for tests:
https://github.com/grafana/helm-charts/blob/main/charts/promtail/ci/service-values.yaml

charts/promtail/values.yaml
piontec: RIght, but this requires overriding the whole `config.file`. I really liked the previous approach, where you could override only some sections and keep your actual config shorter. As for camel case - sure, I'm not sure how I missed that :P

unguiculus: OK, I guess it makes sense.

piontec:
Ping @unguiculus , can you please have a look again?

unguiculus (changes requested):
It doesn't look like you considered my comments. Please use camel case and bump the minor version. Please don't rebase or force-push. Just add additional commits so GitHub can properly display your changes. Thanks.

charts/promtail/values.yaml
```diff
@@ -431,9 +439,11 @@ config:
...
     client:
       url: {{ .Values.config.lokiAddress }}
+      {{- toYaml .Values.extraClientConfigs | nindent 2 }}
 
     positions:
       filename: /run/promtail/positions.yaml
 
     scrape_configs:
       {{- tpl .Values.config.snippets.scrapeConfigs $ | nindent 2 }}
+      {{- toYaml .Values.extraScrapeConfigs | nindent 2 }}
```
unguiculus, line 449: As suggested and to be consistent, these should go under `config.snippets`. We may also want to make them templated. As such, they should be strings.

```suggestion
      {{- tpl .Values.config.snippets.extraClientConfigs . | nindent 2 }}

    positions:
      filename: /run/promtail/positions.yaml

    scrape_configs:
      {{- tpl .Values.config.snippets.scrapeConfigs . | nindent 2 }}
      {{- tpl .Values.config.snippets.extraScrapeConfigs . | nindent 2 }}
```

unguiculus (changes requested):
Would you mind updating the examples in the readme to use the new configs?

charts/promtail/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.1.0
-version: 3.0.4
+version: 3.0.5
```
unguiculus, line 6: Minor bump, please. It's a new feature.

piontec: Good idea, thanks!

charts/promtail/values.yaml
```diff
@@ -214,6 +214,14 @@ podSecurityPolicy:
   requiredDropCapabilities:
     - ALL
 
+# -- You can put here any keys that will be directly added to the config file's 'client' block.
+# @default -- empty
+extraClientConfigs: {}
+
+# -- You can put here any additional scrape configs you want to add to the config file.
+# @default -- empty
+extraScrapeConfigs: {}
```
unguiculus, line 223: You will have to move these to the right place as well.

piontec:
> Would you mind updating the examples in the readme to use the new configs?

I simplified the existing syslog and journald examples with `extraScrapeConfigs` and added basic auth + tenant ID for `extraClientConfig`.

unguiculus:
One more thing: Please update `charts/promtail/ci/service-values.yaml` as well.

unguiculus (approved):

PR #194, opened 2021-01-26
[promtail] Disable initContainer by default

siavashs:
The Promtail initContainer is enabled by default which sets the `fs.inotify.max_user_instances` sysctl to `128`.
This sysctl is not namespaced and it's value is applied to the node.
Also the default value of `128` is too low and can cause issues for other pods running on bigger nodes.
Therefore the sysctl must be disabled by default.


unguiculus (approved):
Good catch. Thanks. It was accidentally enabled by default during the recent refactoring.

PR #202, opened 2021-02-03
[promtail] Support pod logs that use config.hash

siavashs:
Kubernetes uses `kubernetes.io/config.hash` annotation value for pod log paths if present, otherwise the pod UID is used.
This change adds support for pod logs which have the above annotation set.


unguiculus (approved):
Thanks for catching this.

james-callahan:
Does this really work with having two rules that both set the `__path__`: why wouldn't this break logs from pod without `kubernetes.io/config.hash`?

PR #207, opened 2021-02-08
[grafana] Support in-memory for non-persistent local storage

lorenzo-biava:
This PR adds support for mounting the Grafana non-persistent local storage in-memory (tmpfs).
That allows to get better performance when you don't need persistence and when the load on Grafana is high enough to create a lot of "locked db" issues, but not enough to warrant a dedicated external RDBMS.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.3.0
+version: 6.4.0
 appVersion: 7.4.0
 kubeVersion: '^1.8.0-0'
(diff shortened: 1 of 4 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.2.2
+version: 6.3.0
```
zanhsieh, line 3: @lorenzo-biava 
Could you bump the chart again please?

zanhsieh (changes requested):

zanhsieh (approved):

PR #236, opened 2021-02-19
[grafana] Added support for auth tokens for local dashboards

dkusidlo:
Signed-off-by: dkusidlo <dennis.kusidlo@gmail.com>

When using GitHub Enterprise for local dashboards the raw token expires after 14 days and tokenless curl is not supported.
Personal access tokens do not expire and would fix this problem, but do not work when provided in the URL with `?token=`. There we would need support for an optional `Authorization` header in the curl command of the download_dashboards.sh.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.4.3
+version: 6.4.4
 appVersion: 7.4.2
 kubeVersion: '^1.8.0-0'
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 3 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.4.2
+version: 6.4.3
```
zanhsieh, line 3: Please bump the chart to latest version.

zanhsieh:
@dkusidlo 
I think you have to rebase in order to get pass lint-test.

zanhsieh (approved):

PR #241, opened 2021-02-23
[tempo-distributed] Missing values, Extra volumes, synced with loki-distributed.

Whyeasy:
**Description of changes**

Some values were missing within the templates which were available in `values.yaml`. Added the capability to add extra volumes and volume mounts to the deployments. This way GCP SA files can be added the Tempo. Also synced the chart a bit more with the way how `loki-distributed` is. Also did some formatting regarding indents. Extracted the Memcached exporter out of the memcached values.

**Benefits**

Readability
Added missing values in templates
Able to mount the GCP service account json file to the needed services.
Memcached configuration build for expanding.

**Additional Information**

Sorry for the big PR.


charts/tempo-distributed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.6.0
+version: 0.7.0
```
joe-elliott, line 5: i was intending to keep the major and minor version of the chart in sync with the tempo version.  so this would be 0.6.1.  is that a bad idea?

charts/tempo-distributed/templates/distributor/persistentvolumeclaim-compactor.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.compactor.persistence.enabled }}
```
joe-elliott, line 1: compactors don't use the local disk.  this should not be necessary.

joe-elliott (commented):
two comments.  thanks for the cleanup!

charts/tempo-distributed/Chart.yaml
Whyeasy: As far as I know, minor updates are used for the patches / smaller changes. Major versions normally indicate there is a breaking change in the chart when you upgrade. The memcached update is a breaking change in this case.

charts/tempo-distributed/templates/distributor/persistentvolumeclaim-compactor.yaml
Whyeasy: Okay. Just for double checking, I know Loki uses it to reduce start up time. Is this something Tempo can do? or will do in the future?

Whyeasy: I also see this is in the wrong folder 😅

joe-elliott: at the moment we have no plans to use a disk for the compactor.  i wonder if they're using it to persist tokens?  

even if we did decide to add disks for token persistence we'd have to change the compactor deployment to a statefulset.  so it would require additional work.

charts/tempo-distributed/Chart.yaml
joe-elliott: yeah, good point. keeping these in sync while playing nice with semver will probably be impossible in the long run.  may as well rip off the band aid now.  0.7.0 is good :+1:

joe-elliott (approved):
i am mostly unfamiliar with the helm ecosystem.  thank you for bringing your expertise.  :pray:

charts/tempo-distributed/templates/distributor/persistentvolumeclaim-compactor.yaml
Whyeasy: I didn't check what they store, but you can keep the deployment as a deployment 😄 At least Loki-compactor is running as a deployment and run fines.

PR #245, opened 2021-02-24
[tempo-distributed] Add options for LoadBalancer Distributor.

Whyeasy:
**Description of changes**

With this change it would be possible to set additional settings in the Distributor service if you will use a `LoadBalancer` to receive traffic. Also fixed to add annotations in the QueryFrontend.

**Benefits**

- Specify which IP the LoadBalancer should get.
- Specify which IP ranges are allowed.
- Fix annotations in the queryFrontend service.

### DIFF
--- a/charts/tempo-distributed/templates/distributor/service-distributor.yaml
+++ b/charts/tempo-distributed/templates/distributor/service-distributor.yaml
@@ -1,4 +1,4 @@
-  apiVersion: v1
+apiVersion: v1
 kind: Service
 metadata:
   name: {{ template "tempo.fullname" . }}
--- a/charts/tempo-distributed/templates/distributor/service-distributor.yaml
+++ b/charts/tempo-distributed/templates/distributor/service-distributor.yaml
@@ -15,6 +15,10 @@ spec:
     - name: http
       port: 3100
       targetPort: http
+    - name: grpc
+      port: 9095
+      protocol: TCP
+      targetPort: 9095
     {{- if .Values.traces.jaeger.thriftCompact }}
     - name: distributor-jaeger-thrift-compact
       port: 6831
@@ -45,7 +49,7 @@ spec:
       protocol: TCP
       targetPort: zipkin
     {{- end }}
-    {{- if .Values.traces.otlp }}
+    {{- if or (.Values.traces.otlp.http) (.Values.traces.otlp.grpc) }}
     - name: distributor-otlp
       port: 55680
       protocol: TCP
--- a/charts/tempo-distributed/templates/distributor/service-distributor.yaml
+++ b/charts/tempo-distributed/templates/distributor/service-distributor.yaml
@@ -23,19 +23,19 @@ spec:
     - name: distributor-jaeger-thrift-compact
       port: 6831
       protocol: UDP
-      targetPort: jaeger-thrift-compact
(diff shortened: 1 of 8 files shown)


Whyeasy:
Please hold with this PR. Just noticed the current distributor-service can't be of type `LoadBalancer` because it has different protocols.

There are 2 ways to solve this. 

Either make the ports configurable within the distributor service, so users need to specify which type of traces/protocols they would like to receive. But if this would be UDP and TCP we hit the same issue.

Or we split the services in 2 if type is `LoadBalancer`, one for UDP and one for TCP.

Any thoughts?

Whyeasy:
@joe-elliott as discussed on Slack. I changed the way we specify ports in the service, pods and configuration. All protocols are disabled by default and you should select which protocols you would like to enable and ingest.

If there is a case where a person needs an UDP and TCP protocol with a `LoadBalancer`. We will consider this as an edge case and can be solved by deploying a separate `LoadBalancer` outside of this chart.

charts/tempo-distributed/values.yaml
```diff
@@ -316,16 +342,42 @@ config: |
...
+      {{- end }}
+      {{- if .Values.traces.zipkin}}
+      zipkin:
+        endpoint: 0.0.0.0:9411
+      {{- end }}
+      {{- if .Values.traces.otlp}}
+      otlp:
+        protocols:
+          http:
+            endpoint: 0.0.0.0:55680
```
joe-elliott, line 373: does this work?  opening both to 55680?  i feel like this would cause a port conflict.

charts/tempo-distributed/values.yaml
```diff
@@ -316,16 +342,42 @@ config: |
         store: memberlist
   distributor:
     receivers:
+      {{ if .Values.traces.jaeger.enabled }}
```
joe-elliott, line 345: can this just be an OR'ed combination of `.Values.traces.jaeger.grpc`, `.Values.traces.jaeger.thriftBinary`, etc.?

joe-elliott (commented):
Added a couple of questions.  Thanks for the work.

charts/tempo-distributed/templates/distributor/service-distributor.yaml
```diff
@@ -1,48 +1,68 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "tempo.fullname" . }}
+  name: {{ template "tempo.fullname" . }}-distributor
```
joe-elliott, line 4: we may want to mention this as a breaking change since it would change how people reference ingestion

Whyeasy: Added to the README.

charts/tempo-distributed/values.yaml
Whyeasy: Done.

charts/tempo-distributed/values.yaml
Whyeasy: Split the options for otlp. so only 1 endpoint will be enabled.

joe-elliott (approved):

joe-elliott:
Looks good to me.  Thanks for the fixes!

PR #249, opened 2021-02-25
[loki-stack] Fixed wrong org name for chart.

kbudde:
Signed-off-by: Kris Budde <Kris@budd.ee>


kbudde:
Was already fixed somewhere else.

PR #252, opened 2021-02-26
[tempo-distributed] Fix replicas querier & deployment selectorlabels

Whyeasy:
**Description**

Querier didn't had the correct replicas value pointed to it. Upgrading of a chart was not possible due to not using selector labels in the querier and queryFrontend deployment. Removed the hardcoded Jaeger client setting in `tempo-query`.

**Additional Information.**

Don't know if we need to make the chart `0.9.0` or `0.8.1` as upgrading was not working before and the actual deployments need to be deleted to make the upgrade work.


joe-elliott (approved):

joe-elliott:
8.1 works for me.  thanks!

PR #201, opened 2021-02-02
[grafana] making grafana ini templetable upstream

sushmareddem:
This PR makes grafana ini templetable

PR #253, opened 2021-03-01
[grafana] support templating grafana ini options

HeWhoWas:
Addresses the same issue as https://github.com/grafana/helm-charts/pull/201 but works. 

Only difference in output is whitespace output when nil values are supplied.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.4.4
+version: 6.4.5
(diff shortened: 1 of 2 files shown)


zanhsieh (approved):

BadgerOps:
Hi folks - commenting here but we could probably chat over in #335 - I'm curious why this change was merged with no updates to documentation? This broke the documented method of how to insert the grafana.ini into the values.yaml, which is causing issues for me. I don't mind changing my values if the documentation in this repo is updated to align with the purpose of this change. 

Thanks!

PR #262, opened 2021-03-03
[grafana] Modify ingress template to be compatible with new api version

EtienneBarbier:
Add support to new ingress API.


zanhsieh (approved):

ymatsiuk:
This doesn't seem to be a patch release as it introduced breaking change: in case people have their ingress configured in `values.yaml` they must've been manually updated to comply with new structure.

PR #263, opened 2021-03-03
[promtail] Run config.lokiAddress through the tpl function

diranged:
Run the lokiAddress through the `tpl` function.. this allows us to do things like this:

```yaml
promtail:
  config:
    lokiAddress: https://{{ .Release.Name }}-loki-gateway/loki/api/v1/push
```

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.1.0
-version: 3.1.0
+version: 3.2.0
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
--- a/charts/promtail/README.md
+++ b/charts/promtail/README.md
@@ -1,6 +1,6 @@
 # promtail
 
-![Version: 3.1.0](https://img.shields.io/badge/Version-3.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.1.0](https://img.shields.io/badge/AppVersion-2.1.0-informational?style=flat-square)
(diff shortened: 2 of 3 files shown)


unguiculus:
That's already possible and the chart itself does exactly do that. With your PR you template it twice.

diranged:
> That's already possible and the chart itself does exactly do that. With your PR you template it twice.

Can you show me how? I definitely tried it before and after that change and it did not work before. Here I first generate a chart with the 3.1.0 template:

diranged:
@unguiculus,
  Actually I take it back again. The 3.1.0 release will not handle a templated lokiAddress key. If we generate the template with the 3.1.0 release:
```
$ tail -3 Chart.yaml 
  - name: promtail
    version: 3.1.0
$ make template > orig
```

then we update the chart and point it to my test branch
```
$ tail -3 Chart.yaml 
  - name: promtail
    version: 3.2.0
    repository: https://diranged.github.io/grafana-helm-charts
$ make template > new && diff -b -u orig new
...
     app.kubernetes.io/instance: observability
     app.kubernetes.io/version: "2.1.0"
@@ -723,7 +723,7 @@
       http_listen_port: 3101
     
     client:
-      url: https://{{ .Release.Name }}-loki-gateway/loki/api/v1/push
+      url: https://observability-loki-gateway/loki/api/v1/push
@@ -44582,7 +44582,7 @@
 metadata:
   name: observability-promtail
   labels:
-    helm.sh/chart: promtail-3.1.0
```

You can see in the original template the value came through raw, but in my version it gets templatized properly.

diranged:
Here's a simpler test of the 3.1.0 stack:
```
MacBook-Pro:observability diranged$  (⎈ |cluster/dev1-test:observability)$ helm template grafana/promtail --set config.lokiAddress='{{ .Release.Name }}:3100' | grep 'url: '
Error: failed parsing --set data: key "}:3100" has no value
MacBook-Pro:observability diranged$  (⎈ |cluster/dev1-test:observability)$ 
```

unguiculus (approved):
I see. I misunderstood your request. Approved.

PR #267, opened 2021-03-04
[promtail] Add usage instructions

caleb15:
I was surprised to find out that promtail does not work with the `loki` chart out of the box. These directions should help other people going forward.


caleb15:
Closing this - need to redo for DCO :(

Issue #271, opened 2021-03-04
Helm chart reporting erros since grafana-6.4.6


andrew-pickin-epi:
"upgrade failed: unable to recognize \"\": no matches for kind \"Ingress\" in version \"networking.k8s.io/v1\""

Value file is available however this worked file until an hour ago or so, believe the problem is will upstream inconsistency.

ymatsiuk:
There is a breaking change introduced in `6.4.6` and if your `values.yaml` contains ingress configuration, please make sure those comply with `networking.k8s.io/v1` ingress spec.

andrew-pickin-epi:
Breaking changes should not be introduced as a minor version update.

ymatsiuk:
This is completely another story, but I agree!

andrew-pickin-epi:
How? 

This is my current ingress. It looks like it should match the current values.yaml file.

```
    ingress:
      enabled: true
      annotations:
        kubernetes.io/ingress.class: "nginx"
        nginx.ingress.kubernetes.io/auth-url: "https://$host/oauth2/auth"
        nginx.ingress.kubernetes.io/auth-signin: "https://$host/oauth2/start?rd=$escaped_request_uri"
        nginx.ingress.kubernetes.io/configuration-snippet: |
          auth_request_set $user   $upstream_http_x_auth_request_user;
          auth_request_set $email  $upstream_http_x_auth_request_email;
          proxy_set_header X-User  $user;
          proxy_set_header X-Email $email;
      labels: {}
      hosts:
        - grafana__NAME__.epimorphics.net
      paths:
        - /
```

ymatsiuk:
Check comments under [ingress section](https://github.com/grafana/helm-charts/blob/5ccc112675fba1c1d51cb4c59f30a108cda5095e/charts/grafana/values.yaml#L161)

andrew-pickin-epi:
Are you saying that I now need to specify and Ingres Class?
In which case the values.yaml fine needs to show that this is now this mandatory.

My ingress file matches the intent of the current values.yaml file. If this is no longer correct the values.yaml file requires updating to reflect this. Some commented link to documentation is not sufficient.

ymatsiuk:
Just to clarify: I'm in the same situation as you are currently - my helm is also broken due to validation errors. My issue is a bit different though and I've found my answers in the commented out section. 

[...]

PR #275, opened 2021-03-05
[grafana] validate existence of ingress object in new networking api version

bkbwese:
otherwise the if statement will incorrectly use the networking api on
versions of k8s that don't have networking.k8s.io/v1/Ingress <v1.19

earlier PR from issue 271 fixed another issue related to the same recent change,
this should fix the specific error regarding ingress missing from networking.k8s.io/v1
in k8s clusters older than v1.19
https://github.com/grafana/helm-charts/issues/271

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.4.7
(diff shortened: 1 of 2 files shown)


bkbwese:
jk this is a dupe of #273

PR #277, opened 2021-03-08
[loki-distributed] Ease storage config

ubergesundheit:
This PR tries to ease the configuration of storage for the loki-distributed chart.


unguiculus:
While the chart provides a reasonable config out of the box, you will most likely have to tweak and adapt it in various places. With your change, you're exposing a limited set of options to be configured directly. One could ask: Why only Azure and AWS? And why only a subset of S3 options, etc? That being said, this PR would open a can of worms. One could create similar PRs for any kind of config option. This wouldn't make sense and pollute the chart. If you seriously want to use Loki, you have to really roll your own config and know what you do. It's a design decision that the chart is not going to expose individual config options.

PR #279, opened 2021-03-08
[promtail] Disable __path__ override when config.hash annotation is not present

thlacroix:
Since https://github.com/grafana/helm-charts/pull/202, with the default config, promtail sends duplicate logs for pods with container of the same name running on the same host. See this fix and the discussion above for the technical details: https://github.com/grafana/helm-charts/pull/202#commitcomment-47148139

This PR uses @shkrid fix (thanks by the way), which adds a check that the `kubernetes.io/config.hash` annotation is present before replacing.

This might also fix https://github.com/grafana/loki/issues/3443


shkrid (approved):

unguiculus (approved):
Thanks. LGTM

PR #283, opened 2021-03-09
[promtail] fix extraPorts NetworkPolicy

tomasz-kuzemko:
The *-egress-extra-ports NetworkPolicy was always declared
regardless of the value of .Values.networkPolicy.enabled.
This commit fixes this behavior to declare the policy
only if network policies are enabled for the chart.

Signed-off-by: Tomasz Kuzemko <tomasz@kuzemko.net>


czunker (approved):
Good catch. Thank you.

unguiculus (approved):
Thanks.

PR #286, opened 2021-03-11
[tempo] Adding missing STS ports

jtyr:
This PR is adding missing ports for the Tempo `StatefulSet`. Without this, the Service cannot receive messages from any application.


joe-elliott (approved):
let's go!

thx for the fix!

joe-elliott:
Looks like you have to run the helm-docs command.  I think this just uses docker but if it's any trouble I can run it.

joe-elliott (approved):

PR #290, opened 2021-03-11
[loki-stack] updating for loki release 2.2.0

slim-bean:

### DIFF
--- a/charts/loki-stack/Chart.yaml
+++ b/charts/loki-stack/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: "v1"
 name: loki-stack
-version: 2.3.1
-appVersion: v2.1.0
+version: 2.4.0
+appVersion: v2.2.0
 kubeVersion: "^1.10.0-0"
 description: "Loki: like Prometheus, but for logs."
 home: https://grafana.com/loki
--- a/charts/loki-stack/values.yaml
+++ b/charts/loki-stack/values.yaml
@@ -13,7 +13,7 @@ grafana:
     datasources:
       enabled: true
   image:
-    tag: 6.7.0
+    tag: 7.4.3
 
 prometheus:
   enabled: false
--- a/charts/loki-stack/requirements.yaml
+++ b/charts/loki-stack/requirements.yaml
@@ -2,22 +2,22 @@ dependencies:
 - name: "loki"
   condition: loki.enabled
   repository: "https://grafana.github.io/helm-charts"
-  version: "^2.3.0"
+  version: "^2.4.0"
 - name: "promtail"
   condition: promtail.enabled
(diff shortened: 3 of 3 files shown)


owen-d (approved):

charts/loki-stack/values.yaml
```diff
@@ -13,7 +13,7 @@ grafana:
     datasources:
       enabled: true
   image:
-    tag: 6.7.0
+    tag: 7.4.3
```
mkilchhofer, line 16: Is there a reason that we explicitly override the imageTag here? Why we don't us the subchart defaults?

mkilchhofer (approved):

charts/loki-stack/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "v1"
 name: loki-stack
-version: 2.3.1
-appVersion: v2.1.0
+version: 2.4.0
```
mkilchhofer, line 3: I'd bump the major version since at least the promtail chart had a breaking change.

mkilchhofer (changes requested):

christopher-wong:
Hey there :wave: any update on when this might get fixed up and merged in?

christopher-wong:
@slim-bean any chance you could lend a hand here?

patrickhuy:
I hope you had enjoyable easter holidays! Any chance this could be moved forward? @slim-bean  @owen-d

slim-bean:
We just released 2.2.1 so I'm hoping to get back to this today and try to
figure out why it was failing tests.

coultenholt:
We are also waiting on this change to go through as well. Any update @slim-bean ?

coultenholt:
@slim-bean 
What is blocking this PR from being merged? Anything I can help with?

mcpappas:
I just tried to use the loki-stack chart and was confused why promtail wasn't parsing the logs correctly. Several hours of wading through old commits and kubernetes/docker/containerd documentation I've found that this has bitten me too. The simple fact that loki-stack is using a promtail so far behind main is very confusing. If I can find the time, I may submit my own PR to see if we can't get this pushed through to prevent anyone else like me burning time on this.

onedr0p:
It's been more than a couple months @slim-bean Please provide an update ❤️

lucaim:
Hi @slim-bean, it looks to me that the tests are not passing because the promtail chart assumes that it is `loki-gateway` is present (https://github.com/grafana/helm-charts/blob/0a2deec6784b58e06a6794593652ba62e1c706bf/charts/promtail/values.yaml#L235) instead of being just `loki` as a service. 
I suppose that this is missing in the helm chart compared to the Jsonnet version (https://github.com/grafana/loki/blob/main/production/ksonnet/loki/gateway.libsonnet)

PR #295, opened 2021-03-11
[grafana] add ability to specify subPath for the callback url for image renderer

### LINKED ISSUE #294: Image Renderer plugin don't work when serve_from_sub_path is true
I tried to install the image renderer plugin using the remote container option in the Grafana helm chart, we are also hosting our grafana instance using a sub path so we have `serve_from_sub_path: true`.

There is no option in the helm chart to add the sub path to the GF_RENDERING_CALLBACK_URL env variable that is added then imageRenderer.enabled is true.

If the subpath is not used the rendered image will just show the grafana welcome page instead of the actual graph.

rsteneteg:
fixes #294

Signed-off-by: Roger Steneteg <rsteneteg@ea.com>

### DIFF
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -217,6 +217,7 @@ This version requires Helm >= 3.1.0.
 | `imageRenderer.priorityClassName`          | image-renderer deployment priority class                                           | `''`                             |
 | `imageRenderer.service.portName`           | image-renderer service port name                                                   | `'http'`                         |
 | `imageRenderer.service.port`               | image-renderer service port used by both service and deployment                    | `8081`                           |
+| `iamgeRenderer.grafanaSubPath`             | Grafana sub path to use for image renderer callback url                            | `''`                             |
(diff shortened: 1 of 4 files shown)


charts/grafana/README.md
```diff
@@ -217,6 +217,7 @@ This version requires Helm >= 3.1.0.
 | `imageRenderer.priorityClassName`          | image-renderer deployment priority class                                           | `''`                             |
 | `imageRenderer.service.portName`           | image-renderer service port name                                                   | `'http'`                         |
 | `imageRenderer.service.port`               | image-renderer service port used by both service and deployment                    | `8081`                           |
+| `iamgeRenderer.grafanaSubPath`             | Grafana sub path to use for image renderer callback url                            | `''`                             |
```
zanhsieh, line 220: Should it be `imageRender.grafanaSubPath` ?

zanhsieh (changes requested):

charts/grafana/README.md
rsteneteg: Ah, thank you for noticing

zanhsieh (approved):

PR #297, opened 2021-03-11
[loki-stack] Allow overrides for test pod image

### LINKED ISSUE #296: Loki-stack doesn't allow to override image for test pod
There are use-cases when we may want to override hardcoded image for test pod:
1) Loki stack chart uses bats/bats:v1.1.0 for test pod which has High CVE-2019-14697. Users may want to replace it with a different version where the issue was resolved.
2) If the environment, where the chart is being deployed, has restricted access to the external resource we may want to use a different image for test pod from the internal registry.

All that leads to the need to make this image overridable on the chart's level.

sshturm:
fixes #296


zanhsieh:
@sshturm 
Could you fix CI problem please?

zanhsieh (approved):

PR #298, opened 2021-03-12
[tempo] Use specific tag for tempo-query image

jtyr:
This PR is setting a specific version for the `temp-query` image. Without this change the query fails because of https://github.com/grafana/tempo/pull/574.


joe-elliott (approved):
thx for the fix!

PR #300, opened 2021-03-12
[loki-distributed] Memcached can be enabled/disabled in loki

sergeyshaykhullin:


sergeyshaykhullin:
Hmm. i got how memcached works in this chart. Don't need to have enabled for "common" memcached object

PR #304, opened 2021-03-15
[tempo] Fix tempo template for extraVolumeMounts

takuhiro:
This PR fixes broken indents, which causes the error below when extraVolumeMounts is enabled.
```
Error: YAML parse error on tempo/templates/statefulset.yaml: error converting YAML to JSON: yaml: line 61: did not find expected key
```
I also added trim markers for removing empty lines.


joe-elliott (approved):
nice!  thanks for the fix!

Issue #301, opened 2021-03-12
[fluent-bit,loki-stack,loki]: Move all Charts to apiVersion v2


oliverbaehler:
Hi there

I was wondering if it would be possible to upgrade all charts in the repository to the Helm apiVersion v2. Since v1 is already considered deprecated a while. The following Charts are currently still apiVersion 1:

  - [fluent-bit](https://github.com/grafana/helm-charts/blob/2e5fa0f1079948d638230c69e8688fbf147fc177/charts/fluent-bit/Chart.yaml#L1)
  - [loki-stack](https://github.com/grafana/helm-charts/blob/2e5fa0f1079948d638230c69e8688fbf147fc177/charts/loki-stack/Chart.yaml#L1)
    * Requires Dependency Model Change
  - [loki](https://github.com/grafana/helm-charts/blob/2e5fa0f1079948d638230c69e8688fbf147fc177/charts/loki/Chart.yaml#L1)

Changes would include to change all the mentioned charts, Change the dependency model for the loki chart, and bump all the charts to a new major and add as upgrade documentation, that at least helm v3.0.0 is required. I would be more than happy to do that contribution, but I would first like to check with you guys what your opinion is on this matter. :)

torstenwalter:
I am in favor of doing this. @slim-bean what's your opinion?

Sheikh-Abubaker:
The loki-stack chart is deprecated and no longer maintained. Please migrate to the actively maintained [Loki](https://github.com/grafana/loki/tree/main/production/helm/loki) chart referenced in https://github.com/grafana/helm-charts/pull/3974, Closing this issue as not planned

PR #306, opened 2021-03-15
[fluent-bit]  Change default helm apiVersion

oliverbaehler:
Increases to default apiVersion from Helm v2.x to Helm v3.x. Since this change is not backwards compatible (due to certain Helm versions not working anymore), this change requires a new major bump of the chart.

This change is being implemented as part of following issue(s):
  * #301


Whyeasy:
This PR will be closed, due to deprecating the Fluent-bit chart in this repo. Please use the official helm chart from Fleunt-bit: https://github.com/fluent/helm-charts

PR #333, opened 2021-03-22
[enterprise-metrics] Use non deprecated memcached repo

jdbaldry:
Resolves: #309

Uses the bitnami memcached repository but with the same images that were set in the deprecated helm stable chart. Diff has been reduced to a minimum for the change between charts but there are some extra changes to note.

- Memcached limits have been set to match requests. 
Limits don't need to be higher since memcached manages its memory buffer anyway
- Memcached client max item sizes for the store-gateway/querier/ruler have been corrected.
Previously they were based on the wrong memcached flag.

This PR also adds YAML exports for each of the provided values files which can be used to diff the affect of values changes between commits. For instance, diffing the last two commits shows the memcached changes. Unfortunately, line based diffing of YAML files isn't the most effective but it's better than nothing.


jdbaldry:
Resolves: #323

jdbaldry:
Note, for a more readable diff you can exclude certain paths. The exports directory is just an export of the manifests. The vendor directory is just vendored Jsonnet libraries that are used to produce the export.
```console
$ git diff origin/main -- :/: ':!/:charts/enterprise-metrics/exports' ':!/:charts/enterprise-metrics/vendor'
```

charts/enterprise-metrics/CHANGELOG.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Unreleased
+
+## 1.1.0
+
+* [CHANGE] The memcached chart from the deprecated Helm stable repository has been removed and replaced with a Bitnami chart. #333
+  > **Warning:** This change will result in then cycling of your memcached Pods and will invalidate the existing cache.
+* [CHANGE] Memcached Pod resource limits have been lowered to match requests. #333
+* [FEATURE] YAML exports have been created for all chart values files. #333
+* [BUGFIX] The values for the querier/ruler/store-gateway `-<prefix>.memcached.max-item-size` have been corrected to match the limit configuredon the memcached server. #333
```
simonswine, line 21:
```suggestion
* [BUGFIX] The values for the querier/ruler/store-gateway `-<prefix>.memcached.max-item-size` have been corrected to match the limit configured on the memcached server. #333
```

simonswine (approved):
I have tested the upgrade between the last released helm chart and this. All worked fine. :+1: 

Good work & LGTM

simonswine (dismissed):
LGTM

simonswine (approved):
LGTM

Issue #240, opened 2021-02-22
Promtail Helm Chart only Creates Service if ServiceMonitor is Enabled


Alxander64:
Relevant file can be found here: https://github.com/grafana/helm-charts/blob/main/charts/promtail/templates/service-metrics.yaml

The `Service` is only created if `.Values.serviceMonitor.enabled` is `true`.

I create my own `ServiceMonitor` elsewhere for promtail, and I need just the `Service` created by the chart. I shouldn't have to enable the chart's `ServiceMonitor` just to have the relevant `Service` created with it.

unguiculus:
PRs are welcome.

PR #334, opened 2021-03-22
[promtail] create service without needing to enable servicemonitor

Alxander64:
Fixes https://github.com/grafana/helm-charts/issues/240

Added new value to default Helm values `service.enabled` which defaults to false to match `serviceMonitor.enabled`.

Updated the yaml template for the Promtail metrics Service to be created either when `service.enabled` or `serviceMonitor.enabled` is `true`. Decided to use an or condition here so it can still be provisioned by just setting `serviceMonitor.enabled`, so `service.enabled` can be omitted, but also want the new `service.enabled` to create the Service, without needing to enable `serviceMonitor.enabled` and have it provision whatever else that comes with it.

First PR here, so apologies for anything I've missed. Also unsure if the naming is fine, let me know if the new value should be more along the lines of `serviceMetrics.enabled` or something like that.

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.2.0
-version: 3.5.0
+version: 3.5.1
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
--- a/charts/promtail/README.md
+++ b/charts/promtail/README.md
@@ -113,6 +113,7 @@ The new release which will pick up again from the existing `positions.yaml`.
 | rbac.pspEnabled | bool | `false` | Specifies whether a PodSecurityPolicy is to be created |
 | readinessProbe | object | See `values.yaml` | Readiness probe |
 | resources | object | `{}` | Resource requests and limits |
+| service.enabled | bool | `false` | If enabled, Service resource for Promtail is created |
 | serviceAccount.annotations | object | `{}` | Annotations for the service account |
 | serviceAccount.create | bool | `true` | Specifies whether a ServiceAccount should be created |
 | serviceAccount.imagePullSecrets | list | `[]` | Image pull secrets for the service account |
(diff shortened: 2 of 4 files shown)


charts/promtail/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.2.0
-version: 3.5.0
+version: 3.5.1
```
unguiculus, line 6:
```suggestion
version: 3.6.0
```

unguiculus (changes requested):
I don't think it makes much sense to use two different flags for the condition. Since this service is specifically for metrics only anyways, I think it's ok to really only create it when `serviceMonitor.enabled` is `true`.

Alxander64:
My original complaint with just requiring `serviceMonitor.enabled` equal to `true` was that I did not need the actual ServiceMonitor object created by the Helm chart, just the service. By default the `kube_prometheus_stack` chart I have configured only looks at ServiceMonitors within the same namespace it's deployed in, which are usually defined within its values file, and I have that chart and `loki_stack` in separate namespaces. Not to mention it's also configured to only pickup on ServiceMonitors with the same set of labels/annotations applied by the helm chart on all of its resources.

So all this to say that the actual ServiceMonitor object created by the promtail chart is unused sometimes, and may just be clutter and create confusion, so I'd prefer if there were a way to create the necessary Service, but not bother making the ServiceMonitor along with it.

Might it make more sense to instead have a key `serviceMonitor.serviceOnly` with a value of `true`/`false`, with a default of `false`? This would avoid having to create `service.enabled` and it should still keep the same original behaviour by default.

unguiculus:
The service is for metrics only. It is only necessary when a ServiceMonitor is created. I don't think we should really change anything then. It's probably only a configuration issue on your side. Either configure kube-prometheus-stack such that it picks up ServiceMonitors from other namespaces or configure Promtail's ServiceMonitor to be created in kube-prometheus-stack's namespace.

PR #351, opened 2021-03-30
[tempo-distributed] Update component labels

Whyeasy:
**Description**

Updated the component label within the chart. So it is shorter and is not impacted by the helm chart full name.  Now it is more aligned with the loki-distributor chart and with the mixin dashboards.

**Added Benefits**

- Shorter and fixed labels for components, easier to search.
- Smoother integration with mixin dashboards.

**Additional information**

Changing the labels and selector labels, all deployments and statefulsets need to be recreated. Hopefully a once of chart lifetime thing to do. Will be beneficial for the longer run I think.


charts/tempo-distributed/README.md.gotmpl
```diff
@@ -20,6 +20,12 @@ helm repo add grafana https://grafana.github.io/helm-charts
 
 A major chart version change indicates that there is an incompatible breaking change needing manual actions.
 
+### From Chart versions < 0.9.0
+
+This release the component label was shortened to be more aligned with the Loki-distributed chart and the [mixin](https://github.com/grafana/tempo/tree/master/operations/tempo-mixin) dashboards.
+
+Please delete the deployments and statefulsets before executing this upgrade.
```
joe-elliott, line 27: i don't disagree with these changes. the shorter names would definitely be clearer. is it common to require completely removing deployments/statefulsets before upgrading? that's my only concern here.

Whyeasy: Don't know, but I don't see another way to fix this 😞 @unguiculus Do you have any insight regarding these kind of upgrades?

unguiculus: I changed labels for Promtail a while ago. Upgrade options are documented here:
https://github.com/grafana/helm-charts/tree/main/charts/promtail#from-chart-versions--300

Whyeasy: Thanks! will update the README either today or tomorrow.

joe-elliott (approved):
Thanks for the update!

PR #357, opened 2021-03-31
[enterprise-metrics] Use helm deployment name instead of cluster name

simonswine:
I think that makes it clearer. For a moment I got myself confused, until I realized I need to set the cluster name like that:

```
--set "config.cluster_name=<cluster name>"
```

### DIFF
--- a/charts/enterprise-metrics/Chart.yaml
+++ b/charts/enterprise-metrics/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 1.0.0
+version: 1.0.1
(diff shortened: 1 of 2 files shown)


jdbaldry (approved):
Yeah this is definitely more clear

jdbaldry:
@unguiculus Is there an existing way we can opt out of this check for documentation or administrative changes where it makes little sense to bump the chart version?

jdbaldry:
Think this has been superseded by https://github.com/grafana/helm-charts/pull/500

PR #364, opened 2021-04-03
[loki-distributed] Make container name unique

Skaronator:
We use Prometheus for our alerting and since all container name are just "loki" it's quite difficult differentiate which loki microservice is affected. This PR simply change the container names to their appropriate name.


unguiculus:
I'm not opposed to this, but I wonder why you use the container name for alerting. All pods have component labels which you could use.

PR #366, opened 2021-04-05
[grafana] Enhance Grafana PSP to enforce security by policy(Second line of defense)

iordachelm:
Grafana works with less privileges than are allowed by existing PSP. 
In order to make sure that by mistake Grafana is not run with more privileges than it needs, PSP was enhanced.

Motivation for this is to assure that Grafana deployment is secure enough to satisfy ISO 27001 certification, SOC type 1& 2


zanhsieh:
@iordachelm 
Could you resolve the conflict of charts/grafana/Chart.yaml ?

iordachelm:
cannot fix DCO

iordachelm:
trying to fix DCO

PR #371, opened 2021-04-08
[grafana] Enhance Grafana PSP to enforce security by policy(Second line of defense)

iordachelm:
Grafana works with less privileges than are allowed by existing PSP.
In order to make sure that by mistake Grafana is not run with more privileges than it needs, PSP was enhanced.

Motivation for this is to assure that Grafana deployment is secure enough to satisfy ISO 27001 certification, SOC type 1& 2


zanhsieh (approved):

Tolsto:
This PR breaks our installation: `spec.initContainers[0].securityContext.runAsNonRoot: Invalid value: false: must be true`

mattrapa:
Hi, @zanhsieh 

```
 requiredDropCapabilities:
    # Default set from Docker, with DAC_OVERRIDE and CHOWN
      - ALL
```

breaks the Grafana init container from running chown, and the init container is also configured to run as root as per https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl#L31

PR #374, opened 2021-04-09
[loki-distributed] default number of ingester should be 2

nvanheuverzwijn:
If we take the default value of 1 ingester, the distributor output there log:

```
level=warn ts=2021-04-09T20:16:09.463337401Z caller=logging.go:71 traceID=76858732ade9f1b2 msg="POST /loki/api/v1/push (500) 151.072µs Response: \"at least 2 live replicas required, could only find 1\\n\" ws: false; Connection: close; Content-Length: 937; Content-Type: application/x-protobuf; User-Agent: promtail/2.2.0; "
```

Signed-off-by: Nicolas Vanheuverzwijn <nicolas.vanheu@gmail.com>

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.2.0
-version: 0.28.0
+version: 0.28.1
 home: https://grafana.github.io/helm-charts
 sources:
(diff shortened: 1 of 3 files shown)


unguiculus (changes requested):
By default, all services are configured with one replica. This is, so it can be easily run on small systems for testing. In order to fix the issue you mentioned, we should set the replication factor to 1 instead.

nvanheuverzwijn:
@unguiculus I am not sure what you mean by replication factor. Where should I look ?

PR #377, opened 2021-04-12
[loki-rule-operator] add new chart to install loki rules crds

nvanheuverzwijn:
This PR introduces a new helm-chart that add CRDs to manage loki rules.

I beleive it resolves these two issues :
https://github.com/grafana/loki/issues/2990
https://github.com/grafana/loki/issues/3456

See https://github.com/opsgy/loki-rule-operator for source code

### DIFF
--- a/charts/loki-rule-operator/Chart.yaml
+++ b/charts/loki-rule-operator/Chart.yaml
@@ -1,0 +1,8 @@
+apiVersion: v2
+name: loki-rule-operator
+description: Install CRDs to handle Loki rules
+type: application
+sources:
+  - https://github.com/opsgy/loki-rule-operator
+version: 1.0.0
+appVersion: v0.5.0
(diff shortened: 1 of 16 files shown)


unguiculus (changes requested):
I don't think this repo should host charts for non-Grafana apps. Please host that chart in your own repo.

nvanheuverzwijn:
@unguiculus Well, if grafana people wants to take over the source code, I'm pretty sure the author wouldn't mind. However, if Grafana does not care, well, I don't care either. This is a requested feature for a part of the community so it makes sense to have it here or at least, have a reference somewhere if it's only in this PR.

It wouldn't be the first time Grafana would use thrid party code. They do use kiwigrid side-car containers which is not from Grafana team itself.

sharovmerk:
@trevorwhitney can it be integrated to Grafana then?

PR #379, opened 2021-04-16
[fluent-bit] make the container parser value configurable

qzio:
The format of the logs in `/var/log/containers/*.log` on my nodes (k3s) is not in the default docker json-file format, which means I need to specify a custom parser.

`parsers` is already configurable, but if I `kubectl edit` the config map to change the `Parser` value of the [INPUT] , and then run `helm upgrade ..` my changes will be overritten.

This PR makes the value of `Parser` in the default `[INPUT] ` (/var/log/containers/*.log) to be configurable.

There might be a better variable name than `containerParser` though, but it was the best I could come up with.

### DIFF
--- a/charts/fluent-bit/Chart.yaml
+++ b/charts/fluent-bit/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: fluent-bit
-version: 2.2.0
+version: 2.2.1
 appVersion: v2.1.0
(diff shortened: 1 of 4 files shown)


Whyeasy:
This PR will be closed, due to deprecating the Fluent-bit chart in this repo. Please use the official helm chart from Fluent-bit: https://github.com/fluent/helm-charts

PR #390, opened 2021-04-22
[tempo-distributed] Set explicit kvstore in default config.

Whyeasy:

**Description**

Update the default config to set `kvstore` explicit for the distributor and ingester.


joe-elliott:
Any particular reason for this? Tempo defaults to memberlist.

Whyeasy:
Apparently there was an issue with this as mentioned by Christopher on Slack. When I asked him to add the `kvstore` value his issue was fixed. You are tagged in the message as well, maybe give it a read to be sure this is needed 😄.

joe-elliott:
Yeah, I just reviewed that thread. Seems odd. I don't think this should matter, but I don't mind merging. It shouldn't have a negative impact either.

Thanks for following up with this @Whyeasy !

joe-elliott (approved):

PR #394, opened 2021-04-22
[loki-distributed] Add the restartPolicy to all of the Deployments in the chart

diranged:
We have run into a situation where we want to make sure the Loki gateway pod is replaced rather than restarted if it errors out or stops, but we noticed that the `restartPolicy` option was not on any of the `Deployment` templates.


diranged:
https://github.com/kubernetes/kubernetes/issues/24725

Of course, this PR is actually useless and can be closed. I completely forgot about this issue - _which I have commented on already in the past_ ... so oops!

Issue #399, opened 2021-04-24
[Tempo] Not possible to change service type in QueryFrontend


renanqts:
Currently is not possible to change the service type in QueryFronted in Grafana tempo.
In my case, I want to expose it using AWS ALB and there It needs to be NodePort.

PR #400, opened 2021-04-24
[tempo-distributed] Make possible to change service type for queryFrontend

renanqts:
I want to get rid of issue #399.
It should be pretty straightforward just creating the variable `.Values.queryFrontend.service.type`.


joe-elliott:
Thank you for the contribution!

You will need to increment the helm chart version and rerun the command to regenerate the docs:

`docker run --rm --volume "$(pwd):/helm-docs" -u $(id -u) jnorwood/helm-docs:v1.4.0`

renanqts:
@joe-elliott done

PR #415, opened 2021-04-30
[enterprise-metrics] feat: Update to use enterprise metrics v1.3.0

jdbaldry:


simonswine (changes requested):
Nice one, looks all good to me

simonswine (approved):
LGTM

PR #420, opened 2021-05-03
[grafana] Add horizontal pod autoscaling

mleclerc00:
Looking to add horizontal pod autoscaling for grafana when using a deployment resource.

Have tested this locally and seems to work as expected.  This is my first time attempting to contribute here so if there is anything I missed just let me know.

Thanks!


jbritt1:
This is a great addition and would be awesome if HPA was a configurable option while leveraging the prescribed helm chart here in this repository!

zanhsieh (approved):

PR #431, opened 2021-05-10
[promtail] add hostPort support for promtail

lmercl:
There is one approach, when you need to support to promtail listen on local host port. It when you need to collect logs from kubernetes nodes and you want to push these logs to local host port, not to k8s service and to be addicted  to kubernetes DNS.
Working rsyslog config is:
`action(type="omfwd" protocol="tcp" port="31514" target="0.0.0.0" Template="RSYSLOG_SyslogProtocol23Format" TCP_Framing="octet-counted")`


charts/promtail/values.yaml
```diff
@@ -178,6 +178,8 @@ extraPorts: {}
 #    name: tcp-syslog
 #    containerPort: 1514
 #    protocol: TCP
+#    # -- hostPort definition will ensure that you can send logs from host to promtail pod
+#    #hostPort: 31514
```
unguiculus, line 182: Uncomment the port and make it `null`.
```suggestion
    hostPort: null
```

unguiculus (changes requested):
You will have to regenerate the readme using this command:

`docker run --rm --volume "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.4.0`

zanhsieh:
@lmercl 
Closed this PR since has been a long time. Feel free to reopen and rebase the PR.

PR #445, opened 2021-05-20
[fluent-bit] Deprecate the Fluent-bit chart.

Whyeasy:
**Description**

As discussed on Slack; This chart should be deprecated as there as an official helm chart from Fluent-bit and Loki is an official output. 

**Changes**

- Set the `deprecrated` flag. So users will get a warning when installing this chart.
- Updated README to point to the official chart and guide.

**Additional Information**

This [page](https://grafana.com/docs/loki/latest/clients/fluentbit/) should probably be updated.

### DIFF
--- a/charts/fluent-bit/templates/NOTES.txt
+++ b/charts/fluent-bit/templates/NOTES.txt
@@ -4,10 +4,6 @@ Please use the official fluent-bit chart
 
 https://github.com/fluent/helm-charts
 
-and the official loki output
-
-https://docs.fluentbit.io/manual/pipeline/outputs/loki
-
 !WARNING! !WARNING! !WARNING! !WARNING! !WARNING!
 
 Verify the application is working by running these commands:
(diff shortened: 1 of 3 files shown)


unguiculus (changes requested):
You should also print a deprecation message in NOTES.txt.

charts/fluent-bit/templates/NOTES.txt
```diff
@@ -1,3 +1,15 @@
...
+
+Please use the official fluent-bit chart 
+
+https://github.com/fluent/helm-charts
+
+and the official loki output
+
+https://docs.fluentbit.io/manual/pipeline/outputs/loki
+
+!WARNING! !WARNING! !WARNING! !WARNING! !WARNING!
```
unguiculus, line 11: Not sure if this is the official recommendation. The plugin from Fluent may not have reached feature parity. I'd just remove that part.

unguiculus (changes requested):

charts/fluent-bit/templates/NOTES.txt
Whyeasy: Updated the message in README and NOTES.

unguiculus (approved):

Issue #450, opened 2021-05-25
[grafana] Grafana.ini paths not respected in container


jacksgt:
I recently wanted to re-configure the directories Grafana uses to store data.
This Helm chart exposes the "paths" under the "grafana.ini" section, so I thought changing those should have the desired effect.

https://github.com/grafana/helm-charts/blob/76ef792c3d2910f2a4c4996a66c77cc65daa0074/charts/grafana/values.yaml#L529-L538

So I changed those:
```yaml
grafana.ini:
  paths:
    data: /test/data
    plugins: /test/plugins
```

After re-deploying the Helm chart, I was surprised to find out that Grafana was still writing the data to the old directories.
After tracking this issue down, it seems that the Helm chart only modifies the *grafana.ini config file*, but not the launch options of the Grafana container.
The Grafana container uses the environment variables `GF_PATHS_DATA` etc. for launching the Grafana server:
https://github.com/grafana/grafana/blob/0f4806db8a6fba20974a7caf7755f83ac5a7944f/packaging/docker/run.sh#L80-L89

Thus, I was able to achieve the desired effect by setting:
```yaml
env:
  GF_PATHS_DATA: /test/data
  GF_PATHS_PLUGINS: /test/plugins

grafana.ini:
  paths:
    data: /test/data
    plugins: /test/plugins
```

But I think this (setting appropriate environment variables) should be the default behavior when the user explicitly sets `grafana.ini.paths.data` and friends.

I'm happy to make a PR for this if there is consensus.

wilfriedroset:
I've stumble upon the same issue.
I would expect the same behavior as the rest of the configuration file.
The configMap information and what Grafana show either via http://<grafana>/admin/settings or in the logs are different.

Legion2:
https://github.com/grafana/helm-charts/blob/f0e236156f935ccc46c70ca02ec0cb41151b0e7b/charts/grafana/values.yaml#L535
https://github.com/grafana/grafana/blob/0f4806db8a6fba20974a7caf7755f83ac5a7944f/packaging/docker/Dockerfile#L21
the chart valus.yaml already set a `grafana.ini.paths.data` path which is different from the default set in the Dockerfile. As a result this PR introduced a breaking change.

jacksgt:
@Legion2 Wow, I'm sorry to hear that. Thanks for reporting back.

[...]

PR #453, opened 2021-05-26
[grafana] Start Grafana container with GF_PATHS_* env vars

jacksgt:
When the user sets the grafana.ini.paths.* values for the Helm chart,
it is expected that Grafana is also started with the appropriate
configuration (instead of "just" setting the values in the config file).

This patch implements the behavior.

Fixes https://github.com/grafana/helm-charts/issues/450

Signed-off-by: Jack Henschel <jackdev@mailbox.org>

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.9.1
+version: 6.10.0
 appVersion: 7.5.5
 kubeVersion: '^1.8.0-0'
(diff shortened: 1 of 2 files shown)


zanhsieh:
@jacksgt 
Could you bounce the chart version to latest in charts/grafana/Chart.yaml ?

jacksgt:
@zanhsieh I bumped the chart version from 6.9.1 to 6.10.0. Hope that's correct.

zanhsieh (approved):

PR #454, opened 2021-05-26
[agent] create grafana agent chart

AndreZiviani:
A very simple chart to install grafana-agent

To keep it simple I configured it to have all services (prometheus, loki and tempo) on the same pod (daemonset) and a separate pod (deployment) for another prometheus to scrape cluster metrics as recommended.

All services configuration are based from the [upstream](https://github.com/grafana/agent/blob/main/production/kubernetes/) with some improvements.

### DIFF
--- a/charts/agent/Chart.yaml
+++ b/charts/agent/Chart.yaml
@@ -1,0 +1,13 @@
+apiVersion: v2
+name: agent
+description: Grafana Agent is a telemetry collector for sending metrics, logs, and trace data to the opinionated Grafana observability stack.
+type: application
+appVersion: 0.14.0
+version: 1.0.0
+home: https://github.com/grafana/agent/
+sources:
+  - https://grafana.com/docs/grafana-cloud/agent/
+icon: https://raw.githubusercontent.com/grafana/agent/main/docs/assets/logo_and_name.png
+maintainers:
+  - name: ???
+    email: ???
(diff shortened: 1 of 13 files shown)


charts/agent/Chart.yaml
```diff
@@ -0,0 +1,13 @@
...
+type: application
+appVersion: 0.14.0
+version: 1.0.0
+home: https://github.com/grafana/agent/
+sources:
+  - https://grafana.com/docs/grafana-cloud/agent/
+icon: https://raw.githubusercontent.com/grafana/agent/main/docs/assets/logo_and_name.png
+maintainers:
+  - name: ???
+    email: ???
```
AndreZiviani, line 13: Should I add the grafana-agent maintainers or someone else?

rfratto:
Hi, thanks for your contribution! Unfortunately, we're not looking to accept Helm charts for the Grafana Agent into this repository for the time being:

1. The Grafana Agent team have plans to provide official Helm charts for an upcoming Grafana Agent Operator. We'd like to avoid having Grafana-hosted Helm charts for both the non-Operator Agent and the Operator.
2. We don't currently have the bandwidth to maintain a chart for the Agent and become familiar with how it works. While many of the charts in this repository are driven by the community, we still get support questions, and would prefer to have an engineer able to answer questions to help out our support team.

That being said, we still want to help you get your chart recognition. If you're willing to host it yourself, we'd more more than happy to link to it from the Grafana Agent docs as a community-maintained chart.

AndreZiviani:
Oh I see, no problem. Is [this](https://github.com/grafana/agent/pull/549) PR the best place to check for progress (btw nice to see some progress there!) on the Operator development?

I will set this chart up on my account then, will post the link here in case anyone wants to use it, thanks!

rfratto:
> Oh I see, no problem. Is this PR the best place to check for progress (btw nice to see some progress there!) on the Operator development?

Yep, that's the place!

AndreZiviani:
I've uploaded this chart to my personal account in case someone wants to use it while the operator is not finished
https://github.com/AndreZiviani/helm-charts/tree/main/charts/grafana-agent

Any feedback is greatly appreciated!

PR #455, opened 2021-05-27
[promtail] Add SecurityContextConstraints to make it compatible with OpenShift

ahmedwaleedmalik:
This overcomes the limitations set for the default security context in the Openshift distro of Kubernetes. 

@unguiculus I am of the opinion that maybe we can set the key as `openshift.securityContextConstraints.enabled` instead of just `securityContextConstraints.enabled`. Since in general people might end up adding more `openshift` specific overrides. 

I am working on deploying `loki-stack` on openshift so I'll try extend this support to wherever it's required.


ahmedwaleedmalik:
Closing it since this PR didn't get any response.

PR #463, opened 2021-06-02
[loki] Kubernetes recommanded labels and custom labels

nlamirault:
See: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/


charts/loki/templates/_helpers.tpl
```diff
@@ -59,3 +59,26 @@ Create the app name of loki clients. Defaults to the same logic as "loki.fullnam
...
+
+{{/*
+Common labels
+*/}}
+{{- define "loki.labels" -}}
+helm.sh/chart: {{ include "loki.chart" . }}
+{{ include "loki.selectorLabels" . }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+app.kubernetes.io/component: logging
+app.kubernetes.io/part-of: {{ template "loki.name" . }}
```
alrzbv, line 79: why not "include" function as in other similar situations?

charts/loki/templates/configmap-alert.yaml
```diff
@@ -7,9 +7,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: {{ template "loki.name" . }}
```
alrzbv, line 9: why leave the label "app"? if there is "app.kubernetes.io/name"

zanhsieh:
@nlamirault 
Close due to https://github.com/grafana/helm-charts/pull/1772

PR #464, opened 2021-06-02
[promtail] Custom labels and Kubernetes recommanded labels

nlamirault:


unguiculus (changes requested):
Thanks for the PR. The component label doesn't make sense because the chart does not have multiple components. The same applied to the part-of label. We could think about adding custom labels but not on a global level. They could, e.g. make sense on the pod.

PR #465, opened 2021-06-02
[loki-distributed] Table manager replicas from values

sergeyshaykhullin:
Does table manager is designed to have 1+ replicas?

If it is - i guess replicas should be set from values

If not - need to delete replicas from tableManager values and fix pdb


Whyeasy:
As far as I know table manager should only run on it's own. @slim-bean @owen-d Could you please confirm?

unguiculus:
I think so, too. That's why replicas cannot be set.

Whyeasy:
I would say, try it and see if it's a breaking change locally 😄

Whyeasy (approved):
LGTM, thanks!

PR #468, opened 2021-06-02
[tempo-distributed] Allow to configure storage backend through chart values

mfamador:
Hi!

I added the storage value parameter to allow us to configure different storage backends without needing to overwrite the whole config string that is used to create the config map.

The default behaviour is still using the local storage:

```
storage:
  trace:
    backend: local
```

but we can now overwrite it with each specific backend config:

```
storage:
  trace:
    backend: azure
    azure:
      container-name: <REDACTED>
      storage-account-name: <REDACTED>
      storage-account-key: <REDACTED>
```

Thanks!


joe-elliott:
Before we can merge you will need to update the helm chart version and run helm-docs to update it with the parameters you've added.

I appreciate the help on improving the chart!

mfamador:
Hi @joe-elliott 
I've bumped the chart and ran `helm-docs`.
Thank you for reviewing this one.

joe-elliott (approved):
Excellent :).

Thank you for the improvement!

PR #473, opened 2021-06-04
[loki-distributed] Allow to configure storage through chart values

mfamador:
Hi!

I've added `schemaConfig ` and `storageConfig ` to allow us to configure different storages without needing to overwrite the whole config string that is used to create the config map.

The default behaviour is still using the same values:

```
loki:
  schemaConfig:
    configs:
      - from: 2020-09-07
        store: boltdb-shipper
        object_store: filesystem
        schema: v11
        index:
          prefix: loki_index_
          period: 24h

  storageConfig:
    boltdb_shipper:
      shared_store: filesystem
      active_index_directory: /var/loki/index
      cache_location: /var/loki/cache
      cache_ttl: 168h
    filesystem:
      directory: /var/loki/chunks
```

but we can now overwrite it with each specific storage, e.g., azure:

```
loki:
  schemaConfig:
    configs:
    - from: "2018-04-15"
      index:
        period: 168h
        prefix: index_
      object_store: filesystem
      schema: v9
      store: boltdb
    - from: "2020-09-28"
      store: boltdb-shipper
      object_store: azure
      schema: v11
      index:
        prefix: loki_index_
        period: 24h

  storageConfig:
    boltdb:
      directory: /var/loki/index
    filesystem:
      directory: /var/loki/chunks
    azure:
      container_name: <REDACTED>
      account_name: <REDACTED>
      account_key: <REDACTED>
      request_timeout: 0
    boltdb_shipper:
      active_index_directory: /var/loki/index
      shared_store: azure
      cache_location: /var/loki/cache
```

NOTE: I've tried to run `helm-docs` but I'm getting an error.

Thanks!

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.2.1
-version: 0.31.3
+version: 0.31.4
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/charts/loki-distributed/values.yaml
+++ b/charts/loki-distributed/values.yaml
@@ -89,24 +89,14 @@ loki:
       reject_old_samples_max_age: 168h
       max_cache_freshness_per_query: 10m
 
+    {{- if .Values.loki.schemaConfig}}
     schema_config:
-      configs:
-        - from: 2020-09-07
-          store: boltdb-shipper
-          object_store: filesystem
-          schema: v11
-          index:
-            prefix: loki_index_
-            period: 24h
-
+      {{- toYaml .Values.loki.schemaConfig | nindent 2}}
+    {{- end}}
+    {{- if .Values.loki.storageConfig}}
     storage_config:
-      boltdb_shipper:
-        shared_store: filesystem
-        active_index_directory: /var/loki/index
-        cache_location: /var/loki/cache
-        cache_ttl: 168h
-      filesystem:
-        directory: /var/loki/chunks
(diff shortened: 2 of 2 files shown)


unguiculus:
The chart has a reasonable default config to get started. If you need something else, you will have to create your own config file. 

See also: https://github.com/grafana/helm-charts/tree/main/charts/loki-distributed#example-configuration-using-memberlist-boltdb-shipper-and-s3-for-storage

mfamador:
Hi! Yes, that is exactly the point, I didn't want to write the whole config file, that's one of the reasons to use helm in the first place.

mfamador:
Btw, @unguiculus, did you ever try to use this helm chart using remote storage?
Are you aware that if you want to do it, you'd need to rewrite the whole config? And with these new parameters, you wouldn't? And then yes, you could use the default parameters? The way it is it's impossible.
Thank you!

mfamador:
I don't believe that the posture 'if you don't like it the way it is, write the whole thing. closing it.` is a good way to improve the project.

Whyeasy:
Hi @mfamador,

Thanks for you PR, but there are to many variables you can configure in the config file. When you chose to run Loki in a distributed way. You are already picking the "advanced" way.

If we want to be able to configure the whole config file via the `values.yaml` it will be to complex. So the best/easiest practice for the Loki-distributed chart is to copy paste the default config from the yaml and configure it to your liking.

mfamador:
Thanks for the explanation @Whyeasy, I'll rewrite the whole config string despite the fact of wanting all the default values unless my specific storage parameters. But I don't agree, rewriting the whole file should be always the last option. These parameters would a step closer to not needing to. The ultimate improvement would be not having the config string at all and the config map being built dynamically using the chart values.
Thanks again and sorry for my frustrated comments :)

PR #475, opened 2021-06-07
[promtail] allow for potentially more/different initContainers

pitwegner:
In order for the promtail process to run as non-root, an init container must change the permissions of the `/run/promtail` directory. This is not accounted for in the current version of the chart.


charts/promtail/values.yaml
```diff
@@ -4,20 +4,19 @@ nameOverride: null
...
-  # -- Specifies whether the init container for setting inotify max user instances is to be enabled
-  enabled: false
-  image:
-    # -- The Docker registry for the init container
-    registry: docker.io
-    # -- Docker image repository for the init container
-    repository: busybox
-    # -- Docker tag for the init container
-    tag: 1.33
+initContainers:
```
Whyeasy, line 7: I think it would be better to provide an empty list and provide a commented example. By default we don't enable init containers.

Whyeasy (changes requested):

Darnesey:
Are there any updates on this Pull Request? My company really wants to ensure none of our active containers are running as root.

zanhsieh:
Close in favour of https://github.com/grafana/helm-charts/pull/1680 .

PR #481, opened 2021-06-08
Disable loki-stack loki tests if loki disabled

dpkirchner:
This fixes a bug that occurs when trying to install the loki-stack chart with loki disabled:

```
│ Error: template: loki-stack/templates/tests/loki-test-pod.yaml:22:24: executing "loki-stack/templates/tests/loki-test-pod.yaml" at <.Values.loki.service.port>: nil pointer evaluating interface {}.port
```

This error is from terraform helm_release, but the error makes sense and should be the same with regular ol' helm.

### DIFF
--- a/charts/loki-stack/templates/tests/loki-test-configmap.yaml
+++ b/charts/loki-stack/templates/tests/loki-test-configmap.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.loki.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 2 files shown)


dpkirchner:
Nevermind, I don't have time to track down a lawyer to verify and sign a CLA/DCO, sorry.

yasincebeci:
Hello,

I tried this solution but it didn't work for me. thank you
I guess {{- if .Values.loki.enabled }] should be {{- if .Values.loki.enabled }} in loki-test-pod.yaml file.

PR #489, opened 2021-06-09
[loki-stack] Update values.yaml

santk119git:


zanhsieh:
@santk119git 
Could you sign DCO and bump Chart.yaml version to latest please?

PR #490, opened 2021-06-09
[loki-distributed] Configurable gateway strategy

sboschman:
The current default number of replicas for the nginx gateway is one, combined with the rollingupdate with maxSurge=0 and maxUnavailable=1 basically gives a 'recreate' strategy experience. In other words, the single running gateway pod is shutdown before a new pod is getting started during a update/upgrade of the gateway deployment, which is less than ideal imho.

As the update strategy is hardcoded into the gateway deployment template, the only option is to run at least 2 replicas to keep at least one gateway pod up during an update/upgrade. This PR introduces the same mechanism used in the Grafana chart to allow users to overwrite the update strategy for the gateway. At the same time the default behavior is changed in case you use only a single gateway pod (default replicas = 1), a new pod is started before the running pod is terminated. I believe this is the better default behavior.


Whyeasy:
Hi @sboschman,

Please solve the conflict so we can merge your PR.

Thanks!

Whyeasy (approved):
LGTM, Thanks!

PR #500, opened 2021-06-16
[enterprise-metrics] feat: Use {{ .Release.Name }} as the config cluster_name

### LINKED ISSUE #472: [enterprise-metrics] Chart should try to use install name as cluster_name if possible
If not docs should be updated to set the cluster_name:
```console
$ CLUSTER_NAME=test helm install "${CLUSTER_NAME}" grafana/enterprise-metrics --set-file 'license.contents=./license.jwt' --set "config.cluster_name=${CLUSTER_NAME}"
```

jdbaldry:
See 3cd6729 for the change and 84f46a0 for the effects on the exports.

Before the value was `CHANGEME` and needed to be manually changed by the installer in order to match their licensed cluster name. Now, by specifying the licensed cluster name as the release name, the user can install the chart in a single step without needing to override the `cluster_name` in the configuration.

To check the new configuration in one of the exported Kubernetes Secret configuration files:
```console
$ git fetch origin
$ git checkout jdb/re-use-cluster_name
$ grep enterprise-metrics.yaml charts/enterprise-metrics/exports/default/v1.Secret-enterprise-metrics.yaml | awk '{ print $2 }' | base64 -d | grep cluster_name
cluster_name: 'enterprise-metrics'
```

Fixes #472


eamonryan (approved):
Seems solid to me! Nice little fix for the out-of-the-box copy-pasters :)

09jvilla:
This is great @jdbaldry . I'll leave it to @eamonryan to give the technical stamp as my Helm skills are limited. My only question is whether we need some update to the docs on where to specify this value / how to override it. If its just a bug fix, maybe its not actually necessary.

jdbaldry:
@09jvilla Good question. I believe this change is backward compatible as either a user had not changed the value and had the default `CHANGEME` which would not be a valid license, or they had changed the value in their own `values.yaml` file and this change won't affect them.

Issue #516, opened 2021-06-22
Grafana chart


guidoffm:
The Grafana chart uses the properties

`extraContainers` and `extraInitContainers`. Both values use a different syntax.

`extraContainers` expects a string.
`extraInitContainers` expects a list.
`extraContainerVolumes` expects a list.

Why not using a list for `extraContainers` too?

All these properties also differ from the proberties used by 

https://github.com/prometheus-community/helm-charts/

So if you use 

https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack

you have to use different property names.

For Grafana `extraContainers` instead of `containers`
For Grafana `extraInitContainers` instead of `initContainers`
For Grafana `extraContainerVolumes` instead of `volumes`

PR #533, opened 2021-06-29
[grafana] #516/Grafana chart proposal for properties refactoring

guidoffm:


zanhsieh:
@guidoffm 
Thanks for your contribution. Would you mind to make it backward compatible, e.g. the people who already use this chart could be upgraded without rewrite value.yaml?

Other things:

- Sign DCO
- Bump the chart version for major version update, if you insist to push this PR since I am not sure this PR will be backward compatible.

guidoffm:
> @guidoffm
> Thanks for your contribution. Would you mind to make it backward compatible, e.g. the people who already use this chart could be upgraded without rewrite value.yaml?
> 
> Other things:
> 
> * Sign DCO
> * Bump the chart version for major version update, if you insist to push this PR since I am not sure this PR will be backward compatible.

Yes. You are absolutely right. This is a breaking change and might be postponed to the next major relase.

PR #545, opened 2021-07-03
[grafana] chore: grafana: remove namespace from podsecuritypolicy

takirala:
> A Pod Security Policy is a cluster-level resource that controls security sensitive aspects of the pod specification.

https://kubernetes.io/docs/concepts/policy/pod-security-policy/#what-is-a-pod-security-policy

PR #549, opened 2021-07-06
[grafana] Update GitHub action

torstenwalter:
@zanhsieh This is just to try out if updating helm/kind-action solves the issue identified in #545

closes #545 as it adds those changes and also updates github actions.

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -24,7 +24,7 @@ jobs:
           python-version: 3.7
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.0.1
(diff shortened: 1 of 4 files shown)


torstenwalter:
Looks like this works. I see three options to proceed:

- merge this in favor of #545 assuming that @takirala is ok with it as his changes are untouched
- create a PR just with the GitHub actions update, merge it into main and then merge from main into #545
- I file this PR against the source of #545 @takirala merges it into his branch and we get it from their into main

The end result will be the same. The easiest is probably the first option.

takirala:
> @takirala is ok with it as his changes are untouched

i am absolutely fine with any option y'all chose :)

zanhsieh (approved):

PR #553, opened 2021-07-07
[loki-distributed] Give Loki Gateway a /health endpoint

stribb:
Per the API docs, "In microservices mode, the /ready endpoint is exposed by all components" -- yet this is not true for the gateway.

Let's fix this :)

This is specifically a problem for `vector` in its (recommended) health checking mode: it aborts if it can't contact the gateway on start-up.


charts/loki-distributed/values.yaml
```diff
@@ -680,6 +680,11 @@ gateway:
             auth_basic off;
           }
 
+          location = /ready {
+            return 200 'OK';
+            auth_basic off;
+          }
```
torstenwalter, line 686: This will always be ready as soon as nginx is up and running, but it says nothing about the services behind it. Is that by intention? I am not sure if that makes lot of sense here as "/" endpoint serves the same purpose.

readiness probe is currently defined as:

```
  readinessProbe:
    httpGet:
      path: /
      port: http
    initialDelaySeconds: 15
    timeoutSeconds: 1
```

So if one really wants to have a dedicated /ready endpoint then the probe should also be adjusted.
I think the documentation you are referring to is https://grafana.com/docs/loki/latest/api/#microservices-mode

> When deploying Loki in microservices mode, the set of endpoints exposed by each component is different.
> 
> These endpoints are exposed by all components:
> 
> GET /ready
> GET /metrics
> GET /config

stribb: @torstenwalter Thanks, yes, that's what I was alluding to. I know at least `vector` checks health of its backend by going to `<loki-service>/ready`, and that works fine for single-instance Loki but -- at present -- fails for distributed Loki.

I agree, simply replying "OK" feels wrong but with a distributed Loki, what to do? Does (e.g.) the distributor check the ingester service's health before replying OK? If not, simply replying "OK" is probably correct from Gateway's POV. 

I'd be happy to change `readinessProbe` if that's acceptable.

torstenwalter: > I'd be happy to change readinessProbe if that's acceptable.

I think that's ok.

torstenwalter:
You also need to increase the chart version.

Issue #551, opened 2021-07-07
Helm chart suggestions surrounding Tempo and Agent


cabrinha:
First of all, I'd like to thank the team at Grafana Labs for providing such awesome tools like Tempo!

I've heard that there is a dilemma between making helm charts more maintainable vs. more extensible. Example:

# Tempo
The tempo chart uses multiline strings to add yaml to the configmap object [here](https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/values.yaml#L338). But, that makes it hard to change just one value because you'd have to copy the entire (multi-line) string to just add or change one line.

As a solution, I'd suggest to use the [with statement](https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with), because this would allow us to merge defaults from the values.yaml file with user defined overrides. Example:

## configmap-tempo.yaml
```yaml
# configmap-tempo.yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ template "tempo.fullname" . }}
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "tempo.distributorLabels" . | nindent 4 }}
data:
  {{- with .Values.queryFrontend.query.config }}
  tempo-query.yaml: |
    {{- toYaml . | nindent 4 }}
  {{- end }}
  {{- with .Values.overrides }}
  overrides.yaml: |
    {{- toYaml . | nindent 4 }}
  {{- end }}
  {{- with .Values.config }}
  tempo.yaml: | # does this need to be a multiline string?
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

## values.yaml
```yaml
# values.yaml
---
config: # no multiline string syntax here
  multitenancy_enabled: false
  compactor:
    compaction:
      block_retention: 48h
    ring:
      kvstore:
        store: memberlist
  distributor:
    ring:
      kvstore:
        store: memberlist
```

Using this pattern, users can either extend or override the the config, while still keeping all defaults in place.

To change just one option, a user can just pass in:
`helm install -f values.yaml --set config.multitenancy_enabled=true`

All defaults would be preserved and only one option would be changed.

This seems to be a fair way to strike a balance between maintainability and extensibility.

I'd also like to suggest the ability to create an ingress. With a large deployment like Cortex, writing directly into the distributor service could put a lot of load on the kube-proxy. By using an ingress, that load could be shifted to nginx-ingress controller or whatever ingress solution the user employs on their cluster.

[...]

PR #554, opened 2021-07-08
[tempo-distributed] introduce new config structure

cabrinha:
https://github.com/grafana/helm-charts/issues/551

Currently, the config block is a large multiline string, which makes changing just one option cumbersome.

In the multiline string, there is helm templating present, which could easily be moved into the ConfigMap. I saw there were some defaults in place like:

```yaml
  querier:
    frontend_worker:
      frontend_address: {{ include "tempo.queryFrontendFullname" . }}-discovery:9095
```

So, I did my best to keep those sane defaults while still allowing users to override them via the values file.

One of the main themes here is to consolidate the config to a single section or single block.

### DIFF
--- a/charts/tempo-distributed/templates/configmap-tempo.yaml
+++ b/charts/tempo-distributed/templates/configmap-tempo.yaml
@@ -6,9 +6,47 @@ metadata:
   labels:
     {{- include "tempo.distributorLabels" . | nindent 4 }}
 data:
+  {{- with .Values.queryFrontend.query.config }}
   tempo-query.yaml: |
-    {{- tpl .Values.queryFrontend.query.config . | nindent 4 }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  {{- with .Values.overrides }}
   overrides.yaml: |
-    {{- tpl .Values.overrides . | nindent 4 }}
+    overrides:
+      {{- toYaml . | nindent 6 }}
+  {{- end }}
   tempo.yaml: |
-    {{- tpl .Values.config . | nindent 4 }}
+    multitenancy_enabled: {{ .Values.config.multitenancy_enabled }}
+    {{- with .Values.config.compactor }}
+    compactor:
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+    {{- with .Values.config.distributor }}
+    distributor:
+      {{- toYaml . | nindent 4 }}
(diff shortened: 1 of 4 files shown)


joe-elliott:
Lint is catching two things:
- Chart version needs to be bumped
- helm-docs needs to be ran

Technically since our major version is 0.x breaking changes are allowed, but I'd still like to communicate it by bumping the major version. Thoughts?

We should also include an "Upgrading" section in the main readme that outlines the changes for people upgrading from a previous version.

cabrinha:
Sorry for opening this up and letting it get a little stale. We're currently evaluating Tempo vs. ElasticSearch APM. I'm not sure if I'll be able to move these ideas forward. Forgive me! 😅

cabrinha:
> @cabrinha any chance I can continue your work on this? The current config structure is a pain point for me as well and I'd like to improve it. For isntance, the only way of configuring retention right now is to rewrite the entire config value.
> 
> I can either use write access to your fork and work there, or I can re-create this PR using my own fork, whichever works for everyone.

Invited you to collab on my fork. Do with it what you will! More than welcome to take over from here.

trevorwhitney:
There is continued discussion about structured vs. string config values [here](https://github.com/grafana/helm-charts/pull/943). If there is desire to keep our charts consistent, then that may impact this PR.

I personally prefer the string config value because then you know the config at runtime will be exactly as you provided it. If we switch to a structured config as is suggested in this PR, it means a user would have to explicitly `null` out fields from the defaults they don't want. I also think it creates more of a maintenance burden as more thought has to be put into how certain user provided values may interact with defaults in negative ways, since the config is not replace wholesale with what the user provides.

trevorwhitney:
Some more arguments for string value can be found in #588

cabrinha:
We are back on the tempo train! It looks like loki-distributed allows users to use either/or, string or object: https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/values.yaml#L185

trevorwhitney:
@cabrinha circling back on my comments above, I agree we should bring this in-line with the loki config. That philosophy is to use a string config with templating, and template out the sections that most commonly need to be changed from the defaults, like the common config or schema configs.

`loki-distributed` does support a structured config, which will override a string config when present, but I don't think any templates should assume that's present. In other words, templates should not reach into the structured config's properties.

PR #559, opened 2021-07-09
[tempo-distributed] Add option to disable tempo-query sidecar

rverma-nsl:
This also add to fully arm64 compatibility, tempo-query is not needed  for grafana > 7.5

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -241,7 +241,7 @@ querier:
 # Configuration for the query-frontend
 queryFrontend:
   query:
-    enabled: false # required for grafana version <7.5 for compatibility with jaeger-ui. Doesn't work on ARM arch
+    enabled: true # required for grafana version <7.5 for compatibility with jaeger-ui. Doesn't work on ARM arch
     image:
       # -- The Docker registry for the query-frontend image. Overrides `tempo.image.registry`
       registry: null
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -241,7 +241,8 @@ querier:
 # Configuration for the query-frontend
 queryFrontend:
   query:
-    enabled: true # required for grafana version <7.5 for compatibility with jaeger-ui. Doesn't work on ARM arch
+    # -- Required for grafana version <7.5 for compatibility with jaeger-ui. Doesn't work on ARM arch
+    enabled: true
     image:
       # -- The Docker registry for the query-frontend image. Overrides `tempo.image.registry`
       registry: null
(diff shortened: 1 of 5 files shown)


kvrhdn:
Hi, thanks for this! CI is catching some linting issues, can you address those?

- I'd prefer to keep tempo-query enabled by default (for now at least). If we disable it we should probably make this a breaking change and document this as such.

- Can you also remove port 16686 and 16687 from the service?
  https://github.com/grafana/helm-charts/blob/f348d3ea20227bded75d903fd5bab57c9a02bccf/charts/tempo-distributed/templates/query-frontend/service-query-frontend.yaml#L22-L27

charts/tempo-distributed/values.yaml
```diff
@@ -241,6 +241,7 @@ querier:
 # Configuration for the query-frontend
 queryFrontend:
   query:
+    enabled: true # required for grafana version <7.5 for compatibility with jaeger-ui. Doesn't work on ARM arch
```
kvrhdn, line 244: Linting is complaining about this, I think you have to add at least two spaces before a comment.

kvrhdn (changes requested):
CI is still complaining about the PR title but I'm not sure why 🤷🏻 

Can you address this as well:

- Can you also remove port 16686 and 16687 from the service?
  https://github.com/grafana/helm-charts/blob/f348d3ea20227bded75d903fd5bab57c9a02bccf/charts/tempo-distributed/templates/query-frontend/service-query-frontend.yaml#L22-L27

charts/tempo-distributed/values.yaml
rverma-nsl: done

kvrhdn:
Sorry, I formulated that wrong 😓

> Can you also remove port 16686 and 16687 from the service?

We shouldn't remove these ports since they are used when you run tempo-query, but make them conditional instead. Can you also wrap them with:
```
{{- if .Values.queryFrontend.query.enabled }}
...
{{- end }}
```
Remove was the wrong word, should have used disable instead.

rverma-nsl:
> Sorry, I formulated that wrong 😓
> 
> > Can you also remove port 16686 and 16687 from the service?
> 
> We shouldn't remove these ports since they are used when you run tempo-query, but make them conditional instead. Can you also wrap them with:
> 
> ```
> {{- if .Values.queryFrontend.query.enabled }}
> ...
> {{- end }}
> ```
> 
> Remove was the wrong word, should have used disable instead.

No worries, I assumed that even the jaeger ui queries are now happening through port 3100. But I learnt thats not the case :)

kvrhdn (approved):
Thanks for all the changes, this is looking good to me :+1:

PR #562, opened 2021-07-14
[enterprise-metrics] Use correct data directory for compactor persistence

jdbaldry:


charts/enterprise-metrics/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## 1.4.1
 
+* [CHANGE] The compactor data directory configuration has been corrected to `/data`.
+  > **Note:** The compactor is stateless and no data stored in the existing data directory needs to be moved in order to facilitate this upgrade.
+  > For more information, refer to the [Cortex Compactor documentation](https://cortexmetrics.io/docs/blocks-storage/compactor/).
```
chaudum, line 19: You updated the Chart version to 1.4.2 but the changelog is under 1.4.1

chaudum (changes requested):

MichelHollands (dismissed):
LGTM (after addressing the comments)

charts/enterprise-metrics/CHANGELOG.md
jdbaldry: Good catch, thanks

charts/enterprise-metrics/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 1.4.1
+version: 1.4.2
 appVersion: v1.4.1
```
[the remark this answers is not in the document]
jdbaldry: I think at the time the download for v1.4.2 didn't exist but it won't hurt to update the PR to include this. Note, that despite the versioning appearing to match up, the Helm chart versioning is completely separate to the GEM versioning. It's a bit too close at the moment that it causes confusion but I expect it will diverge at some point.

chaudum (approved):
lgtm

PR #571, opened 2021-07-19
[grafana] Optionally automount SA token

jonkerj:
**Description of the change**
It disables the automounting of the service account token in the pod. Grafana does not need this by itself. This commit disables it by default, but leaves it configurable if anyone needs to use it.

**Benefits**
By disabling the automount, potential attackers cannot access the Kubernetes API on behalf/through the pod.

**Possible drawbacks**
In the unlikely event that anyone is using some sidecar or plugin to access the Kubernetes API, they will have to explicitly enable the mounting of the SA token in the values.

**Applicable issues**
If raising an issue is required or helpful, please tell me and I'd be glad to do so.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.13.10
+version: 6.14.0
 appVersion: 8.0.5
 kubeVersion: '^1.8.0-0'
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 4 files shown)


chrisjohnson00:
> **Possible drawbacks**
> In the unlikely event that anyone is using some sidecar or plugin to access the Kubernetes API, they will have to explicitly enable the mounting of the SA token in the values.

This indeed breaks existing installs for those of us who load dashboards and datasources via the config map side car.

Adding:

```
    serviceAccount:
      autoMount: 'true'
```
To my `values.yaml` override gets it to work again.

jonkerj:
> This indeed breaks existing installs for those of us who load dashboards and datasources via the config map side car.
> 
> Adding:
> 
> ```
>     serviceAccount:
>       autoMount: 'true'
> ```
> 
> To my `values.yaml` override gets it to work again.

Whoops, how could I have missed that?! I even use this myself 😳 Thanks for noticing, @chrisjohnson00 

Looking back, I think the default should be `true`. Should I send a new PR?

zanhsieh:
@jonkerj 
Yes. Just make a new PR and link to @chrisjohnson00 comment.

Issue #581, opened 2021-07-26
[enterprise-metrics] Consul configuration should be purged


jdbaldry:
Consul has not been used in recent releases of the chart and all values files can have the Consul specific configuration removed.

09jvilla:
PR: https://github.com/grafana/helm-charts/pull/655

PR #583, opened 2021-07-27
[enterprise-metrics] Add values files for CPU capped installations

chaudum:
These values files contain Kubernetes resource limits and should only be used
when running GEM through a marketplace subscription where you need to have
control over costs which are based on provisioned CPUs.

In these files, resource requests and resource limits are set to the same values
in order to ensure ["Guaranteed" QoS](https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/).

### Sizing

The resource requests and limits are equal to the already existing files `small.yaml` and `large.yaml` respectively.

```console
$ capped-small.yaml 
service               replicas    total cpu request    total mem requests    total cpu limits  total mem limits
------------------  ----------  -------------------  --------------------  ------------------  ------------------
admin_api                    3                    3                   768                   3  768.0 MiB
alertmanager                 3                    3                    24                   3  24.0 GiB
compactor                    1                    2                     2                   2  2.0 GiB
consul                       1                    1                     2                   0  0.0 Byte
distributor                  3                    6                    12                   6  12.0 GiB
gateway                      3                    3                  1152                   3  1152.0 MiB
ingester                     4                   16                    60                  16  60.0 GiB
memcached                    1                    0                     0                   0  0.0 Byte
memcached-queries            1                    0                     0                   0  0.0 Byte
memcached-metadata           1                    0                     0                   0  0.0 Byte
minio                        1                    0                     0                   0  0.0 Byte
querier                      2                    2                    24                   2  24.0 GiB
query_frontend               1                    2                     6                   2  6.0 GiB
ruler                        2                    2                    12                   2  12.0 GiB
store_gateway                1                    1                     6                   1  6.0 GiB
TOTAL                       28                                                             40  147.875 GiB
```

```console
$ capped-large.yaml 
service               replicas    total cpu request    total mem requests    total cpu limits  total mem limits
------------------  ----------  -------------------  --------------------  ------------------  ------------------
admin_api                    3                    3                   768                   3  768.0 MiB
alertmanager                 3                    3                    24                   3  24.0 GiB
compactor                    1                    2                     2                   2  2.0 GiB
consul                       1                    4                     4                   0  0.0 Byte
distributor                 15                   30                    60                  30  60.0 GiB
gateway                      8                    8                  3072                   8  3072.0 MiB
ingester                    25                  100                   375                 100  375.0 GiB
memcached                    1                    0                     0                   0  0.0 Byte
memcached-queries            1                    0                     0                   0  0.0 Byte
memcached-metadata           1                    0                     0                   0  0.0 Byte
minio                        1                    0                     0                   0  0.0 Byte
querier                      6                    6                    72                   6  72.0 GiB
query_frontend               2                    4                    12                   4  12.0 GiB
ruler                        2                    2                    12                   2  12.0 GiB
store_gateway                1                    1                     6                   1  6.0 GiB
TOTAL                       71                                                            159  566.75 GiB
```

### DIFF
--- a/charts/enterprise-metrics/capped-large.yaml
+++ b/charts/enterprise-metrics/capped-large.yaml
@@ -1,0 +1,144 @@
+# These values configure the Grafana Enterprise Metrics cluster to
+# handle production ingestion of ~10M active series using the blocks
+# storage engine.
+# Query requirements can vary dramatically depending on query rate and query
+# ranges. The values here satisfy a "usual" query load as seen from our
+# production clusters at this scale.
+# It is important to ensure that you run no more than one ingester replica
+# per node so that a single node failure does not cause data loss. Zone aware
+# replication can be configured to ensure data replication spans availability
+# zones. Refer to [Zone Aware Replication](https://cortexmetrics.io/docs/guides/zone-aware-replication/)
+# for more information.
+# Minio is no longer enabled and you are encouraged to use your cloud providers
+# object storage service for production deployments.
+admin_api:
+  replicas: 3
+  resources:
+    requests:
+      cpu: 1
+      memory: 256Mi
+    limits:
+      cpu: 1
+      memory: 256Mi
+
+alertmanager:
+  persistentVolume:
+    enabled: true
+  replicas: 3
+  resources:
+    requests:
+      cpu: 1
+      memory: 8Gi
+    limits:
+      cpu: 1
+      memory: 8Gi
+  statefulset:
+    enabled: true
+
+compactor:
+  persistentVolume:
+    size: 50Gi
+  resources:
+    requests:
+      cpu: 2
+      memory: 2Gi
+    limits:
+      cpu: 2
+      memory: 2Gi
+
+consul:
+  resources:
+    requests:
+      cpu: 4
+      memory: 4Gi
+  server:
+    replicas: 3
+    bootstrapExpect: 3
+
+distributor:
+  replicas: 15
+  resources:
+    requests:
+      cpu: 2
+      memory: 4Gi
+    limits:
+      cpu: 2
+      memory: 4Gi
+
+gateway:
+  replicas: 8
+  resources:
+    requests:
+      cpu: 1
+      memory: 384Mi
+    limits:
+      cpu: 1
+      memory: 384Mi
+
+ingester:
+  persistentVolume:
+    size: 50Gi
+  replicas: 25
+  resources:
+    requests:
+      cpu: 4
+      memory: 15Gi
+    limits:
+      cpu: 4
+      memory: 15Gi
+
+memcached:
+  enabled: true
+  replicaCount: 32
+
+memcached-queries:
+  enabled: true
+  replicaCount: 10
+
+memcached-metadata:
+  enabled: true
+
+minio:
+  enabled: false
+
+querier:
+  replicas: 6
+  resources:
+    requests:
+      cpu: 1
+      memory: 12Gi
+    limits:
+      cpu: 1
+      memory: 12Gi
+
+query_frontend:
+  replicas: 2
+  resources:
+    requests:
+      cpu: 2
+      memory: 6Gi
+    limits:
+      cpu: 2
+      memory: 6Gi
+
+ruler:
+  replicas: 2
+  resources:
+    requests:
+      cpu: 1
+      memory: 6Gi
+    limits:
+      cpu: 1
+      memory: 6Gi
+
+store_gateway:
+  persistentVolume:
+    size: 50Gi
+  replicas: 1
+  resources:
+    requests:
+      cpu: 1
+      memory: 6Gi
+    limits:
+      cpu: 1
+      memory: 6Gi
--- a/charts/enterprise-metrics/CHANGELOG.md
+++ b/charts/enterprise-metrics/CHANGELOG.md
@@ -12,6 +12,10 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## Unreleased
 
+## 1.4.3
+
+* [ENHANCEMENT] Added values files for installations that require setting resource limits. #583
+
 ## 1.4.2
 
 * [CHANGE] The compactor data directory configuration has been corrected to `/data`. #562
--- a/charts/enterprise-metrics/Chart.yaml
+++ b/charts/enterprise-metrics/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 1.4.2
+version: 1.4.3
 appVersion: v1.4.2
 description: 'Grafana Enterprise Metrics'
 engine: gotpl
--- a/charts/enterprise-metrics/capped-small.yaml
+++ b/charts/enterprise-metrics/capped-small.yaml
@@ -1,0 +1,146 @@
+# These values configure the Grafana Enterprise Metrics cluster to
+# handle production ingestion of ~1M active series using the blocks
+# storage engine.
+# Query requirements can vary dramatically depending on query rate and query
+# ranges. The values here satisfy a "usual" query load as seen from our
+# production clusters at this scale.
+# It is important to ensure that you run no more than one ingester replica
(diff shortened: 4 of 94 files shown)


charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+# replication can be configured to ensure data replication spans availability
+# zones. Refer to [Zone Aware Replication](https://cortexmetrics.io/docs/guides/zone-aware-replication/)
+# for more information.
+# Minio is no longer enabled and you are encouraged to use your cloud providers
+# object storage service for production deployments.
+admin_api:
+  replicas: 3
+  resources:
+    requests:
+      cpu: 1
```
jdbaldry, line 18: I noticed that the requests have been bumped from the `200m` in `large.yaml` to `1` in this configuration, was this intended and do we want to update `large.yaml` too?

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+      memory: 256Mi
+
+alertmanager:
+  persistentVolume:
+    enabled: true
+  replicas: 3
+  resources:
+    requests:
+      cpu: 1
```
jdbaldry, line 30: Similar question as above for the admin_api requests. `300m` -> `1`

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+      memory: 8Gi
+  statefulset:
+    enabled: true
+
+compactor:
+  persistentVolume:
+    size: 50Gi
+  resources:
+    requests:
+      cpu: 2
```
jdbaldry, line 43: Similar question as above for the admin_api requests. `1` -> `2`. Additionally, I'm not sure if it is worth setting this to `2` CPUs if we don't increase compactor concurrency too.

It's also drawn my attention to the compactor PV size which should probably be close to 150Gi given the goal to handle 10M active series (https://cortexmetrics.io/docs/blocks-storage/compactor/). This is dependent on the size of the largest tenant but 150Gi should cover us even if it is only a single tenant.

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+      memory: 2Gi
+    limits:
+      cpu: 2
+      memory: 2Gi
+
+consul:
```
jdbaldry, line 49: As we discussed previously, we should clean this up but in a separate PR (#581)

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+  persistentVolume:
+    size: 50Gi
+  replicas: 25
+  resources:
+    requests:
+      cpu: 4
+      memory: 25Gi
+    limits:
+      cpu: 4
+      memory: 25Gi
```
jdbaldry, line 88: This has bumped the ingester memory requests to match the limits from the `large.yaml` file. I believe that we should already have adequate head room with the requests alone and I probably should not have included an arbitrary limit in the `large.yaml` configuration in the first place.

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+
+memcached:
+  enabled: true
+  replicaCount: 2
```
jdbaldry, line 92: This drastically changes the replica count from what is in the `large.yaml` file. Possibly the original value is overkill (resulting in around 32GiB) of cache but was this change intended?

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+
+memcached-queries:
+  enabled: true
+  replicaCount: 3
```
jdbaldry, line 96: Similar question to the one above about memcached replica changes. `10` -> `3`.

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+  enabled: true
+
+minio:
+  enabled: false
+
+querier:
+  replicas: 6
+  resources:
+    requests:
+      cpu: 2
```
jdbaldry, line 108: Similar question to the above about request changes from those in `large.yaml`. `1` -> `2`.

Also, as with the ingesters, this is taking the memory limit rather than the request from the `large.yaml` file. The query path is much harder to anticipate resource-wise so perhaps this makes more sense than matching with the `large.yaml`.

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+      memory: 6Gi
+    limits:
+      cpu: 2
+      memory: 6Gi
+
+ruler:
+  replicas: 2
+  resources:
+    requests:
+      cpu: 4
```
jdbaldry, line 128: Similar question regarding changes from `large.yaml`. `1` ->  `4`.

charts/enterprise-metrics/capped-large.yaml
```diff
@@ -0,0 +1,144 @@
...
+      memory: 16Gi
+
+store_gateway:
+  persistentVolume:
+    size: 50Gi
+  replicas: 1
+  resources:
+    requests:
+      cpu: 2
```
jdbaldry, line 140: Similar question regarding changes from `large.yaml`. `1` ->  `2`.

jdbaldry (commented):
Only taken a look at `capped-large.yaml` so far. Have a few questions to full understand the intention of this configuration.

charts/enterprise-metrics/capped-large.yaml
chaudum: GEM is billed by CPU cores, which means you'd pay for a full core, even if you'd limit the container to only a fraction of it. Therefore the maximization.

charts/enterprise-metrics/capped-large.yaml
chaudum: Good catch, that was not intended!

charts/enterprise-metrics/capped-large.yaml
chaudum: I'll go through the replicas values again!

charts/enterprise-metrics/capped-large.yaml
chaudum: limit was increased from `1200m` to `2` for the same reason as stated above.
therefore also the requests changed from `1` to `2`

09jvilla:
Thanks for putting this together @chaudum . Appreciate the thoughtful feedback from @jdbaldry . I'll leave it to him or another IC to give the final thumbs up, but just giving my approval that this makes sense as something we want for marketplace installs and that rounding up to the next whole number CPU makes sense giving the billing model.

chaudum:
CI linting fail due to 

` ✖︎ enterprise-metrics => (version: "1.4.2", path: "charts/enterprise-metrics") > Chart version not ok. Needs a version bump!`

Not sure why we need to bump the version when auxiliary files change 😞

jdbaldry:
> CI linting fail due to
> 
> ` ✖︎ enterprise-metrics => (version: "1.4.2", path: "charts/enterprise-metrics") > Chart version not ok. Needs a version bump!`
> 
> Not sure why we need to bump the version when auxiliary files change

Yup, unfortunately that is the case. I haven't invested any time in making this more sympathetic to charts with auxiliary files

jdbaldry (approved):
LGTM, thanks for your patience. One last thing, would you be able to sign off all the commits? There are instructions in the failing DCO GitHub action.

chaudum:
> LGTM, thanks for your patience. One last thing, would you be able to sign off all the commits? There are instructions in the failing DCO GitHub action.

Sure, I wanted to wait for your ACK before squashing and signing.

jdbaldry (approved):
Thanks! All good assuming CI passes

PR #584, opened 2021-07-27
[promtail] Introduce .Release.Namespace in objects meta

marucha:
Description of the change

This PR adds .Release.Namespace to objects meta for promtail chart. Related to discussion in this [issue](https://github.com/bitnami/charts/issues/2006).

Benefits

Ability to use the helm chart when having declarative definition of all cluster objects rendered using helm template.

Possible drawbacks

I don't see any, it will work as previously for helm install

Signed-off-by: Mariusz Waligóra <mariusz@piwik.pro>

### DIFF
--- a/charts/promtail/templates/service-extra.yaml
+++ b/charts/promtail/templates/service-extra.yaml
@@ -4,6 +4,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "promtail.fullname" $ }}-{{ $key | lower }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "promtail.labels" $ | nindent 4 }}
     {{- with .labels }}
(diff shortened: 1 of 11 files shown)


torstenwalter (dismissed):
For me this looks good. @unguiculus what's your opinion?

torstenwalter:
@marucha please look at the failed ci jobs:
> Documentation not up to date. Please run helm-docs and commit changes!
> templates/: template: promtail/templates/service-extra.yaml:7:24: executing "promtail/templates/service-extra.yaml" at <.Release.Namespace>: nil pointer evaluating interface {}.Namespace

charts/promtail/templates/service-extra.yaml
```diff
@@ -4,6 +4,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "promtail.fullname" $ }}-{{ $key | lower }}
+  namespace: {{ .Release.Namespace }}
```
torstenwalter, line 7: I think it's best to define a variable for the `.Release.Namespace` before the `range` loop.
https://helm.sh/docs/chart_template_guide/variables/

marucha: Fixed it.

torstenwalter (approved):

PR #505, opened 2021-06-17
[grafana] Improve ingress support for K8s v1.18

stevehipwell:
The current Grafana chart doesn't support configuring `pathType` for ingress on K8s v1.18 despite it being supported, this PR fixes this with a pattern similar to that used by the _kube-prometheus-stack_ chart.

Fixes #503.

PR #587, opened 2021-07-29
[grafana] Improve ingress API logic

stevehipwell:
This PR improves the logic in #505.

### DIFF
--- a/charts/grafana/templates/_helpers.tpl
+++ b/charts/grafana/templates/_helpers.tpl
@@ -118,13 +118,6 @@ new password and use it.
 {{- end -}}
 
 {{/*
-Get KubeVersion removing pre-release information.
-*/}}
-{{- define "grafana.kubeVersion" -}}
-  {{- default .Capabilities.KubeVersion.Version (regexFind "v[0-9]+\\.[0-9]+\\.[0-9]+" .Capabilities.KubeVersion.Version) -}}
-{{- end -}}
-
-{{/*
 Return the appropriate apiVersion for rbac.
 */}}
 {{- define "grafana.rbac.apiVersion" -}}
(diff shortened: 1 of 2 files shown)


charts/grafana/templates/_helpers.tpl
```diff
@@ -139,7 +132,7 @@ Return the appropriate apiVersion for rbac.
 Return the appropriate apiVersion for ingress.
 */}}
 {{- define "grafana.ingress.apiVersion" -}}
-  {{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= 1.19.x" (include "grafana.kubeVersion" .)) -}}
+  {{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= 1.19-0" .Capabilities.KubeVersion.Version) -}}
```
torstenwalter, line 135: Where do I find an explanation for the `-0`?

stevehipwell: It's documented in the [Helm docs](https://helm.sh/docs/chart_template_guide/function_list/#semvercompare), but the short explanation is that `-0` is the lowest possible pre-release version. After I made the original changes in #505 I spent a bit of time looking around and this pattern seems to be idiomatic and once I figured out what it was doing I saw that it was the correct solution for where we want to constrain on a K8s release line.

zanhsieh (approved):

naimo84:
Hey ho, 

I have similar problems like here: https://github.com/argoproj/argo-helm/issues/753

If you use argocd to deploy your Helm Chart the .Capabilities.KubeVersion.Version is v1.20.0 but the actual k8s cluster version is v1.18.0

So the ingress apiVersion is wrong :(

Is it possible to introduce a "kubeVersionOverride"-like property?

```yaml
argo-workflows:
  kubeVersionOverride: "1.18-3"
```

Currently I can't use the grafana helm chart.

Greets,
Benjamin

stevehipwell:
@naimo84 that sounds like an Argo defect to me. Do they not have the equivalent of the Helm flag to set the cluster version?

naimo84:
@stevehipwell unfortunatelly not 😐 as argumented here: 
https://github.com/argoproj/argo-cd/issues/5202#issuecomment-756406140

only a "workaround" is possible

stevehipwell:
@naimo84 I'm referring to the `--kube-version' flag which doesn't need any cluster credentials. When I updated the ingress templates I made sure they'd work like this.

naimo84:
@stevehipwell many many thanks for pointing me the right way 👍 it's Part of the next version https://github.com/argoproj/argo-cd/pull/6359
and thanks for your great work regarding the templates 😀

Issue #433, opened 2021-05-11
[loki-distributed] Provide loki configuration not as string


dbluxo:
The README says:

> Note that `loki.config` must be configured as string.
That's required because it is passed through the `tpl` function in order to support templating.
This means that a complete configuration needs to be supplied to the charts which is a good thing anyways.
Also, this allows using a separate YAML file which can be passed in using `--set-file loki.config=/path/to/config.yaml`.

I don't understand the reasoning or the great benefit of it.

In my opinion it would be much more advantageous to use a "normal" YAML structure here, because:

1. via a YAML linter you would be able to detect "errors" in the config
2. we would have a syntax highlighting within the config part
3. you could overwrite individual values of the chart without always having to specify the huge config

Templating would also work with the `tpl` function in the configmap template by using:

```
data:
  config.yaml: |
    {{- tpl (toYaml .Values.loki.config) . | nindent 4 }}
```

Am I missing something? What is your opinion on this? I would be happy to provide a PR for this.

stevehipwell:
It's not just the config where this is an issue, affinity also incorrectly uses a string for the same quoted reason. @dbluxo it might be worth creating a PR for this even if just as a draft?

dbluxo:
In the PR, this proposal was rejected.

PR #588, opened 2021-07-30
[loki-distributed] Provide loki configuration as object

dbluxo:
Closes https://github.com/grafana/helm-charts/issues/433


stevehipwell:
@dbluxo could you also convert the `affinity` values into objects instead of strings?

unguiculus (commented):
I don't think we should change this. It has been designed like this on purpose. Converting the config object YAML for templating has its limitations. You will only be able to template values in the YAML that are strings because you need to quote template expressions. You wouldn't be able to template ports anymore, for example. Also, it would be a breaking change that breaks existing installation that rely on this.
Furthermore, with the existing solution you can externalize the config as YAML file using `--set-file`, which would no longer work.

stevehipwell:
I've just come to the same conclusion today @unguiculus.

Whyeasy:
Can we close this PR if we are not going to change it?

PR #598, opened 2021-08-06
[loki-distributed] Ability overwrite default image command for all loki pods

andy-nix:
Added ability to overwrite default image command for all loki pods.
It is helpful for such situations like:
* integration with vault-agent-injector that mount credentials as files and you want to bring credentials from there to Loki config
* Mount and execute wrapper script to run some actions before loki process start


charts/loki-distributed/templates/compactor/deployment-compactor.yaml
```diff
@@ -39,6 +39,9 @@ spec:
         - name: loki
           image: {{ include "loki.compactorImage" . }}
           imagePullPolicy: {{ .Values.loki.image.pullPolicy }}
+          {{- if or .Values.loki.command .Values.compactor.command }}
+          command: [{{ coalesce .Values.compactor.command .Values.loki.command | quote }}]
+          {{- end }}
```
unguiculus, line 44: Please use block style for YAML lists. Apply that to all your changes.

```suggestion
          {{- if or .Values.loki.command .Values.compactor.command }}
          command: 
            - {{ coalesce .Values.compactor.command .Values.loki.command | quote }}
          {{- end }}
```

unguiculus (dismissed):

zanhsieh (approved):

PR #599, opened 2021-08-06
[agent-operator] Adding agent-operator chart

jtyr:
This PR is adding Helm chart for Grafana Agent Operator.

Signed-off-by: Jiri Tyr <jiri.tyr@gmail.com>

PR #604, opened 2021-08-06
[agent] Adding agent chart

jtyr:
This PR is adding Helm chart for Grafana Agent that gets installed via Grafana Agent Operator.

Signed-off-by: Jiri Tyr <jiri.tyr@gmail.com>

### DIFF
--- a/ct.yaml
+++ b/ct.yaml
@@ -12,6 +12,3 @@ chart-repos:
   - minio=https://helm.min.io
 helm-extra-args: --timeout 600s
 validate-maintainers: false
-# TODO: Remove this after the Helm chart for agent-operator is merged (PR #599)
-excluded-charts:
-  - agent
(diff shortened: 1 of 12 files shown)


ct.yaml
```diff
@@ -12,3 +12,6 @@ chart-repos:
   - minio=https://helm.min.io
 helm-extra-args: --timeout 600s
 validate-maintainers: false
+# TODO: Remove this after the Helm chart for agent-operator is merged (PR #599)
+excluded-charts:
+  - agent
```
torstenwalter, line 17: Why do we need it?

jtyr: This is here to skip running tests for this chart because it contains `GrafanaAgent` Custom Resource whose CRD is deployed by the [`agent-operator` chart](https://github.com/grafana/helm-charts/pull/599) that is not yet merged. Once the `agent-operator` chart is merged, we should remove this exclusion and perhaps define dependency on the `agent-operator` chart.

zanhsieh:
Close because agent-operator chart already merged to the main branch.

PR #613, opened 2021-08-09
[enterprise-logs] helm chart redo

techgnosis:
Here's the bones of a GEL helm chart. It uses loki-distributed as a dependency and borrows Admin API, Gateway, and Compactor from the GEM chart. 

Both the new templates and the loki-distributed chart use the same config file so the loki-distributed chart became the lowest common denominator. I'll refer to this limitation as 'LCD' in the list of issues below.

I kept the service account stuff since it had to do with PSPs. I don't know if anyone cares about PSPs anymore. Perhaps they can be removed.

I was not working in a branch and one of Ed's changes made it into the history between my commits so I can't rebase the history. It's probably best to use this PR to discuss and then just pull the files out and make a new branch

Current issues I'm aware of:
1. There is a config file issue I can't figure out. It only affects distributor, querier, and frontend. I suspect it's something small and not related to the structure of the chart.
2. The loki-distributed Pods are still named loki-distributed. I had an alias in Chart.yaml that fixed that but it was making the work harder to reason about so I removed it. It can probably be added back in.
3. There is no large.yaml for sizing the cluster. I used the values given to me by JenV for the small.yaml file.
4. I'm not really sure what the `clients` section of the helpers file is for. It's still in the file.
5. LCD caused me to have to remove the runtime config section of the config file. This could be fixed at the loki-distributed chart.
6. LCD caused me to have to duplicate a very small section of the config file in a config: block like the GEM chart. It only has the server ports in it. They are used by some of the templates so I couldn't remove them.


charts/enterprise-logs/values.yaml
```diff
@@ -0,0 +1,404 @@
...
+license:
+  contents: "NOTAVALIDLICENSE"
+  external: false
+  secretName: 'enterprise-logs-license'
+
+tokengenJob:
+  enable: true
+  extraArgs: {}
+  env: []
+  annotations: {}
```
chaudum, line 56: The `tokengen-job.yaml` template is missing.

chaudum (commented):
I think this looks pretty good already 👏🏼

charts/enterprise-logs/small.yaml
```diff
@@ -0,0 +1,121 @@
...
+consul:
+  client:
+    enabled: false
+  resources:
+    requests:
+      cpu: 1
+      memory: 2Gi
+  server:
+    replicas: 3
+    bootstrapExpect: 3
```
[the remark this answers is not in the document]
jdbaldry: Yeah, consul is unused and can be removed

charts/enterprise-logs/templates/admin-api-dep.yaml
```diff
@@ -0,0 +1,164 @@
...
+    matchLabels:
+      app: {{ template "enterprise-logs.name" . }}-admin-api
+      release: {{ .Release.Name }}
+  strategy:
+    {{- toYaml .Values.admin_api.strategy | nindent 4 }}
+  template:
+    metadata:
+      labels:
+        app: {{ template "enterprise-logs.name" . }}-admin-api
+        # The name label is important for cortex-mixin compatibility which expects certain names for services.
```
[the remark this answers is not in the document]
jdbaldry: The name label is still valuable in our environment for our specific scrape configuration though I imagine we can instead use our internal tooling to add it in. I'm not sure that our specific Prometheus scrape configs are ubiquitous enough to justify this labels inclusion. 

It was originally added as the GEM helm chart wanted to be compatible with the Cortex monitoring mixin but I think we self monitoring this is much less important.

charts/enterprise-logs/templates/compactor-statefulset.yaml
```diff
@@ -0,0 +1,157 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
jdbaldry: Yeah I do think running the compactor as a StatefulSet is preferable semantically. Since the loki-distributed chart uses a Deployment, I would say we should just inherit that for now and consider introducing a change upstream.

charts/enterprise-logs/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: "v1"
+name: enterprise-logs
+version: 0.1.0
+appVersion: v2.2.1
+kubeVersion: "^1.10.0-0"
+description: "Grafana Enterprise Logs"
+home: https://grafana.com/products/enterprise/logs/
+icon: tbd
```
jdbaldry, line 8: We don't yet have a GEL logo so we can use the Loki one

```suggestion
icon: https://grafana.com/docs/loki/latest/logo_and_name.png
```

charts/enterprise-logs/Chart.yaml
```diff
@@ -0,0 +1,23 @@
...
+sources:
+- https://github.com/grafana/loki
+maintainers:
```
jdbaldry, line 11: I'd be tempted to remove this completely for now.

charts/enterprise-logs/small.yaml
```diff
@@ -0,0 +1,121 @@
...
+  replicas: 4
+  resources:
+    limits:
+      cpu: 2
+      memory: 14Gi
+    requests:
+      cpu: 1
+      memory: 7Gi
+
+memcached:
```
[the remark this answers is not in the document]
jdbaldry: I've no strong opinion on which ones we use, as far as I can tell at a glance both the loki-distributed memcached templates and the bitnami ones offer similar control to the user. It looks like tempo-distributed has its own memcached templates too. Perhaps avoiding the extra external dependency is a win

charts/enterprise-logs/templates/admin-api-dep.yaml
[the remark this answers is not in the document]
jdbaldry: That being said, the comment is certainly unclear in the GEL context. I'd consider removing it.

charts/enterprise-logs/values.yaml
```diff
@@ -0,0 +1,404 @@
...
+  #  - secretName: chart-example-tls
+  #    hosts:
+  #      - chart-example.local
+
+serviceAccount:
+  create: true
+  name:
+  annotations: {}
+
+useExternalConfig: false
```
jdbaldry, line 38: Do these configuration settings make sense with this Chart?

charts/enterprise-logs/values.yaml
```diff
@@ -0,0 +1,404 @@
...
+  external: false
+  secretName: 'enterprise-logs-license'
+
+tokengenJob:
+  enable: true
+  extraArgs: {}
+  env: []
+  annotations: {}
+
+config:
```
jdbaldry, line 58: Aren't we looking to orient around a single configuration string?

charts/enterprise-logs/values.yaml
```diff
@@ -0,0 +1,404 @@
...
+    http_listen_port: 3100
+    grpc_listen_port: 9095
+
+runtimeConfig: {}
+
+rbac:
+  create: true
+  pspEnabled: true
+
+admin_api:
```
jdbaldry, line 69: nit: I believe we will want to camelCase these fields. Don't worry about consistency with the GEM chart which is the one that has historic inconsistencies.

charts/enterprise-logs/values.yaml
```diff
@@ -0,0 +1,404 @@
...
+  extraContainers: []
+  extraVolumes: []
+  nodeSelector: {}
+  affinity: {}
+  tolerations: []
+  terminationGracePeriodSeconds: 60
+
+
+compactor:
+  enabled: true
```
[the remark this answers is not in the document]
jdbaldry: Yeah this can be removed in favor of the distributed chart configuration for the compactor.

charts/enterprise-logs/values.yaml
```diff
@@ -0,0 +1,404 @@
...
+  gateway:
+    enabled: false
+  compactor:
+    enabled: false
+  loki:
+    image:
+      repository: grafana/enterprise-logs
+      tag: v1.1.0
+    # this config is also used for Admin API and Gateway. I can't tell the subchart to look in gel so I'm telling gel to look here
+    config: |
```
jdbaldry, line 305: can we plumb the runtimeConfig stuff in here too?

jdbaldry (commented):
Really great progress on this! Thanks for all your effort so far

charts/enterprise-logs/values.yaml
jdbaldry: I see that you noted this in your PR description, let's discuss this in our meeting today and follow up here.

charts/enterprise-logs/values.yaml
jdbaldry: Also see that this is discussed in your PR description. We can probably remove all runtime configuration for now.

charts/enterprise-logs/small.yaml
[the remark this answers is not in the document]
techgnosis: commit pushed

PR #614, opened 2021-08-10
[tempo] Enabled the opencensus receive in the default Helm template for Tempo

rsanting:
In the Tempo values file it is mentioned that all ports and protocols that Tempo supports are enabled. However not all of them actually are enabled. I added opencensus, there are more like Zipkin that I could add but I haven't tested them personally. I did test Opencensus and that works.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.7.4
+version: 0.7.5
 appVersion: v1.0.1
 engine: gotpl
 home: https://grafana.net
--- a/charts/tempo/README.md
+++ b/charts/tempo/README.md
@@ -1,6 +1,6 @@
 # tempo
 
-![Version: 0.7.4](https://img.shields.io/badge/Version-0.7.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v1.0.1](https://img.shields.io/badge/AppVersion-v1.0.1-informational?style=flat-square)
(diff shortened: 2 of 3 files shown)


joe-elliott:
Thanks for the addition! Will merge once PR passes.

joe-elliott:
You may need to push an empty commit to get CI to wake up. I can't seem to force it to retry through the github UI.

Also, you will need to increment the chart version and re-run helm-docs so you may as well do that in the commit :).

rsanting:
Thanks, I have to admit I forgot about the chart version and I didn't know about the helm-docs. Next time I will come better prepared. All should be well in the latest commit. Let's hope CI runs.

rsanting:
Ci ran and it came up with two errors. I fixed them in a force push but if that stops CI from running again I can push an empty commit.

rsanting:
Sorry for the mess, I'm not used to running helm-docs manually and I messed up the tag one time. Now all should be fine. Looking at the helm-docs though, would a pull request that adds pre-commit configs be useful or is that not wanted?

joe-elliott:
> Sorry for the mess, I'm not used to running helm-docs manually and I messed up the tag one time.
Not at all! Thank you for the contribution.

> Looking at the helm-docs though, would a pull request that adds pre-commit configs be useful or is that not wanted?
I dunno. Unsure what you mean by pre-commit configs. I'm not super familiar with helm-docs or the way this CI is setup. I just know it makes us run it when we submit a PR.  Generally I'm open to any improvement.

Thanks again!

charts/tempo/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.7.4
-appVersion: v1.0.1
+version: 0.7.5
+appVersion: v1.0.2
```
joe-elliott, line 6: let's keep this at 1.0.1. this should match the actual tempo version which will remain 1.0.1.

joe-elliott (changes requested):

rsanting:
Pre-commit is a framework for managing and maintaining multi-language pre-commit hooks. If we add a config for it anyone that has pre-commit installed and enables it wouldn't need to run the helm-docs by hand. I will test it out with this repo and just make a pull-request, it could always be denied.

joe-elliott (approved):
Thanks for the contribution!

PR #616, opened 2021-08-11
[enterprise-logs] Add GEL helm chart

techgnosis:
Here's the bones of a GEL helm chart. It uses loki-distributed as a dependency and borrows Admin API, Gateway, and Compactor from the GEM chart.

Both the new templates and the loki-distributed chart use the same config file so the loki-distributed chart became the lowest common denominator.

I kept the service account stuff since it had to do with PSPs. I don't know if anyone cares about PSPs anymore. Perhaps they can be removed.

Current issues I'm aware of:

1. There is a config file issue I can't figure out. It only affects distributor, querier, and frontend. I suspect it's something small and not related to the structure of the chart.
2. The loki-distributed Pods are still named loki-distributed. I had an alias in Chart.yaml that fixed that but it was making the work harder to reason about so I removed it. It can probably be added back in.
3. There is no large.yaml for sizing the cluster. I used the values given to me by JenV for the small.yaml file.
4. I. 'm not really sure what the clients section of the helpers file is for. It's still in the file.
5. I had to have to remove the runtime config section of the config file since it was incompatible with loki-distributed. This could be fixed at the loki-distributed chart.
6. I had to duplicate a very small section of the config file in a config: block like the GEM chart. It only has the server ports in it. They are used by some of the templates so I couldn't remove them.

### DIFF
--- a/charts/enterprise-logs/Chart.yaml
+++ b/charts/enterprise-logs/Chart.yaml
@@ -1,0 +1,17 @@
+apiVersion: "v2"
+name: enterprise-logs
+version: 0.1.0
+appVersion: v2.2.1
+kubeVersion: "^1.10.0-0"
+description: "Grafana Enterprise Logs"
+home: https://grafana.com/products/enterprise/logs/
+dependencies:
+- name: "loki-distributed"
+  repository: "https://grafana.github.io/helm-charts"
+  version: "^0.35.2"
+  alias: lokidistributed
+- name: minio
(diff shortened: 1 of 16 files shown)


mkemeric (approved):

jdbaldry:
Closing this as it was superseded by #629. Thanks for all your work on this @techgnosis

jdbaldry:
j

PR #622, opened 2021-08-13
[enterprise-metrics] Enable enterprise auth by default

chaudum:
Ref https://github.com/grafana/backend-enterprise/issues/2026


charts/enterprise-metrics/CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## Unreleased
 
+## 1.4.5
+
+* [CHANGE] Enabled enterprise authentication by default.
+  > **Breaking:** This change can cause loosing accesss to the GEM cluster in case auth.type has not
```
jdbaldry, line 18:
```suggestion
  > **Breaking:** This change can cause losing access to the GEM cluster in case `auth.type` has not
```

charts/enterprise-metrics/CHANGELOG.md
```diff
@@ -12,6 +12,13 @@ Entries should include a reference to the Pull Request that introduced the chang
...
+  > been set explicitly.
+  > This is a security related change and therefore released as bugfix.
```
jdbaldry, line 20: I understand this to mean that we are incrementing the patch version rather than a minor version despite it being a breaking change although I was initially confused by the word "bugfix" which is one of our CHANGELOG categories.

Perhaps:

```suggestion
  > This is a security related change and therefore is in a patch release.
```

^ I'm struggling with the right wording too.

charts/enterprise-metrics/values.yaml
```diff
@@ -56,7 +56,9 @@ tokengenJob:
   annotations: {}
 
 config:
-  auth_enabled: false
```
jdbaldry, line 59: You probably already knew the default was `true` and so it didn't need to be explicitly included but I thought I'd just mention it anyway as a point of note -- both `auth_enabled: true` and `auth.type: enterprise` are necessary to enable enterprise authentication.

jdbaldry (dismissed):
LGTM, I'm sorry that I caused conflicts by merging my branch before noticing your PR. 

Would you mind rebasing?

charts/enterprise-metrics/values.yaml
```diff
@@ -56,7 +56,10 @@ tokengenJob:
...
+  auth_enabled: true
```
jdbaldry, line 59: I like the explicit nature of this change, at least in the context of this patch thanks for including it even though it is the GEM default.

jdbaldry (approved):
LGTM! Thanks for rebasing and addressing the comments.

PR #624, opened 2021-08-13
[loki-distributed] nginx istio patch

josiahg:
The loki-distributed helm chart does not work in a cluster using Istio. 

By default nginx proxies using HTTP 1.0, which Istio rejects.

This PR sets the nginx proxy HTTP version to 1.1, which Istio accepts.


unguiculus (changes requested):
Please fix DCO, update readme, bump chart version.

zanhsieh (approved):

PR #627, opened 2021-08-17
[loki-distributed] Adding namespace to all resources to support those who use helm templates

chrism417:
This allows those who don't use Helm to deploy, but use helm template to create the manifests.


torstenwalter:
@chrism417 can you fix the DCO check, please?

chrism417:
> @chrism417 can you fix the DCO check and increase the chart version, please?

Done, sorry for the delay

torstenwalter:
@chrism417 please have a look at errors reported by CI.

chrism417:
Thanks I removed the namespace addition from secret-gateway.yaml

chrism417:
@torstenwalter I don't understand what I'm doing wrong here. It's telling me to run helm-docs and recommit. I ran "helm docs" and committed, but it's still failing.

torstenwalter (approved):

stevehipwell:
Could I ask why this was needed? The `helm template` command is meant to be namespace agnostic allowing you to target any namespace with the payload. If you need to target a namespace this is the responsibility of the apply stage. This PR fundamentally breaks idiomatic Helm behaviour.

Whyeasy:
Can we revert this change please? As this shouldn't be needed and I agree with @stevehipwell

PR #633, opened 2021-08-20
add ingressClass for loki

ams0:
Similarly to #80 , this adds `ingressClassName` support for Loki chart.


muffl0n:
Fixed in #956

PR #635, opened 2021-08-23
docs: Fix jsonpath of loadBalancer service hostname

prashant-shahi:


zanhsieh:
@prashant-shahi 
Could you bump the chart version to latest and sign DCO please?

PR #647, opened 2021-08-30
[loki-distributed] Adds missing affinity in compactor

karlmuscat-dpg:
All microservices can have their `affinity` set, apart from the compactor.


unguiculus (dismissed):
Please update values.yaml, readme, and bump the chart version.

zanhsieh (approved):

PR #648, opened 2021-08-31
[grafana] Added required parameter to example documentation

TheDruidsKeeper:
The image-render requires the [default](https://github.com/grafana/grafana-image-renderer/blob/v3.0.1/src/config.ts#L65) `--no-sandbox,` parameter to run in headless mode, or it [fails](https://bugs.chromium.org/p/chromium/issues/detail?id=638180).


zanhsieh:
@TheDruidsKeeper 
Could you sign DCO and update the chart version to latest please?

TheDruidsKeeper:
@zanhsieh Done.

zanhsieh (approved):

charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.16.2
+version: 6.16.3
```
zanhsieh, line 3: Now should be 6.16.4.

zanhsieh (changes requested):

charts/grafana/Chart.yaml
TheDruidsKeeper: Updated to 6.16.6 (currently the next version)

zanhsieh (approved):

Issue #611, opened 2021-08-09
[loki-distributed] Support index-gateway


stevehipwell:
The new _index-gateway_ component looks like something that should be supported in the _loki-distributed_ chart. Based on the documentation this would allow the _querier_ component to be made into a stateless deployment and remove the need for the _ruler_ deployment to have persistence, which would simplify the chart.

luistilingue:
It'll be great.

stevehipwell:
Is anyone currently planning on doing this work? I was half expecting it in the last release as a headline feature of the Loki release.

luistilingue:
Any news?

sherifkayad:
@luistilingue @stevehipwell I could try to give it a spin with a PR .. I would need some support testing that though. Would you be able to help me with testing and validating?

stevehipwell:
@sherifkayad I'd be happy to test your changes. This would be easiest if it was to be sooner than later as we're currently working on integrating _loki-distributed_ into a cluster template and as such are actively testing it.

sherifkayad:
@stevehipwell I linked my PR for that. Would be great if you can have a look and give it some rounds of testing. Feedback is much welcome.

stevehipwell:
Closed by https://github.com/grafana/helm-charts/pull/649.

PR #649, opened 2021-08-31
[loki-distributed] Enabling Loki Index Gateway

sherifkayad:
As per #611 here's a PR to enable the Index Gateway component (introduced in Loki v2.3)


sherifkayad:
@stevehipwell will be great if you can give that setup a try and test it ;)

sherifkayad:
@Whyeasy @unguiculus could you please have a look at this PR? .. The `helm-docs` is causing some Linting failures .. probably has to do with the fact that I'm using a non-Linux machine?! I ran the helm-docs docker container and still the same issue .. That was the case in previous PRs with me .. I would appreciate it if you can run the helm-docs commands for me ..

stevehipwell:
@sherifkayad you have some line ending issues that might be causing lint failures.

stevehipwell:
@sherifkayad the changes look good to me, let me give it a try.

sherifkayad:
@stevehipwell feel free to run the helm-docs from your machine .. not sure if that might help. Should I give you access to my personal repo?

stevehipwell:
@sherifkayad the line endings are on some of the yaml files, the PR diff view shows them up with an icon and your editor should be able to fix them.

sherifkayad:
@stevehipwell did you give it a try? how are the results?

stevehipwell:
@sherifkayad I've given it a spin and everything looks correct, but I'd need a released chart to give it a more in depth and longer test.

I'd suggest disabling the index gateway by default for a release before setting it to default in the release after that, assuming there are no issues. I also assume that with the index gateway off the output of `helm template` is identical to the previous version?

sherifkayad:
@stevehipwell yes .. the Output should be identical given that the gateway is off. That's a good idea with releasing a minor version now with the index-gateway off and later release a major one with it being off.

I would also like to get the opinion of the reviewers on this point @Whyeasy @unguiculus what do you think?

sherifkayad:
@Whyeasy any feedback from your side?

Whyeasy:
So I think it's better to have the index gateway off by default. This way it won't bring a long big changes for the people that are not interested in the change. The ones that are interested can simply turn it on.

sherifkayad:
@Whyeasy Thanks for the feedback. Will adjust that accordingly. Do you think it should be a minor release then?

Whyeasy:
It doesn't bring any breaking changes if people upgrade, so minor would be fine.

sherifkayad:
@Whyeasy changes applied. Can you please have a look?

Whyeasy:
@sherifkayad Update the chart version and we are good to go!

sherifkayad:
@Whyeasy just rebased to the latest main. Looking forward to the merge once the build succeeds.

sherifkayad:
@Whyeasy all green ;)

Whyeasy (approved):
LGTM! Thanks for all the work.

sherifkayad:
@Whyeasy Sure thing! thank you for the quick feedback.

PR #675, opened 2021-09-08
[enterprise-metrics] Enable leader election for the admin-api by default

jdbaldry:
This change does not affect single replica deployments of the
admin-api but does fix the potential for an inconsistent state when
running with multiple replicas of the admin-api and experiencing
parallel writes for the same objects.

Signed-off-by: Jack Baldry <jack.baldry@grafana.com>


charts/enterprise-metrics/CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 1.5.3
+
+* [BUGFIX] This change does not affect single replica deployments of the
+admin-api but does fix the potential for an inconsistent state when
+running with multiple replicas of the admin-api and experiencing
+parallel writes for the same objects. #675
```
chaudum, line 18: Seems like indentation is missing:

```suggestion
* [BUGFIX] This change does not affect single replica deployments of the
  admin-api but does fix the potential for an inconsistent state when
  running with multiple replicas of the admin-api and experiencing
  parallel writes for the same objects. #675
```

jdbaldry: Thanks! I hope you don't mind, I manually committed the indentation as part of fixing up some commits. Committing suggestions from GitHub has made the DCO fail for me in the past.

jdbaldry: Just realized that may be because my public email was wrong in my GitHub settings...

chaudum (approved):

PR #684, opened 2021-09-17
[enterprise-logs] Remove table-manager and deploy index-gateway

chaudum:
Fixes #682
Fixes #683 
Fixes #685

PR #691, opened 2021-09-20
[enterprise-logs] Run GEL as non-root user `enterprise-logs`

### LINKED ISSUE #687: [enterprise-logs] Run GEL containers as non root user
The Pods that are created by the parent chart do not have a security context configured. However, since the GEL Docker container runs as user `enterprise-logs` (`uid=10001,gid=10001`), their security context needs to be configured accordingly.

```yaml
securityContext:
  runAsNonRoot: true
  runAsGroup: 10001
  runAsUser: 10001
  fsGroup: 10001
```

chaudum:
This PR contains PR #684 and therefore needs to be rebased after that is merged.

Fixes #687

### DIFF
--- a/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
+++ b/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
@@ -42,8 +42,10 @@ spec:
       priorityClassName: {{ .Values.adminApi.priorityClassName }}
       {{- end }}
       securityContext:
-        {{- toYaml .Values.adminApi.securityContext | nindent 8 }}
+        fsGroup: 10001
       initContainers:
+      # Taken from
+      # https://github.com/minio/charts/blob/a5c84bcbad884728bff5c9c23541f936d57a13b3/minio/templates/post-install-create-bucket-job.yaml
       {{- if .Values.minio.enabled }}
         - name: minio-mc
           image: "{{ .Values.minio.mcImage.repository }}:{{ .Values.minio.mcImage.tag }}"
@@ -115,6 +117,7 @@ spec:
             {{- toYaml .Values.adminApi.resources | nindent 12 }}
           securityContext:
             readOnlyRootFilesystem: true
+            {{- toYaml .Values.adminApi.securityContext | nindent 12 }}
           env:
             {{- if .Values.adminApi.env }}
             {{ toYaml .Values.adminApi.env | nindent 12 }}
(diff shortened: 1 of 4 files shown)


charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
```diff
@@ -42,8 +42,10 @@ spec:
       priorityClassName: {{ .Values.adminApi.priorityClassName }}
       {{- end }}
       securityContext:
-        {{- toYaml .Values.adminApi.securityContext | nindent 8 }}
+        fsGroup: 10001
```
jdbaldry, line 45: If I understand this correctly, this affects all containers in the Pod but I guess minio is fine with this as its user can still read the configuration that is mounted in. We need the Pod level SecurityContext to support the volumeMounts in the AdminAPI container because the Container SecurityContext doesn't affect the Pod Volumes.

jdbaldry (commented):
Looks pretty good to me, will have another quick look after the rebase just to be sure

charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
chaudum: Right. `fsGroup` is only supported on the Pod spec, but not on container level. Since minio runs as root, this is ok.

chaudum:
@jdbaldry rebased with `main`

charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
vlad-diachenko, line 45: why do we use hardcoded value here but not variable that can be changed in values file?

vlad-diachenko (approved):

charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
jdbaldry: I think if we did want to make it a variable, the loki-distributed chart uses `podSecurityContext` to differentiate it from the container SecurityContexts IIRC.

jdbaldry (approved):

charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
chaudum: Good question, and one I asked myself as well. I decided to hard-code the value here, because it must not be changed or otherwise it breaks.
The other `securityContext`s are configurable to be as coherent with the `loki-distributed` chart as possible.

vlad-diachenko: ok, thanks.

PR #697, opened 2021-09-23
[grafana] fix grafana image tag

nalbam:
A version different from the grafana image tag set in Chart.yaml is distributed.
Change it so that the same version is deployed.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.16.8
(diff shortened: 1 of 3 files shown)


nalbam:
tested with

```
      source:
        repoURL: https://opspresso.github.io/helm-charts
        targetRevision: "6.16.9"
        chart: grafana
```

BitProcessor:
Double: https://github.com/grafana/helm-charts/pull/698

zanhsieh:
Favor https://github.com/grafana/helm-charts/pull/698

PR #698, opened 2021-09-23
[grafana] Fix the image tag in the grafana chart

BitProcessor:
Signed-off-by: Maarten De Wispelaere <maarten@bitprocessor.be>

The grafana chart isn't using the `appVersion` from the Chart.yaml as image tag. 
The previous 2 changes only changed the appVersion, not the `image.tag` in the values.yaml.
This commit fixes that.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.16.8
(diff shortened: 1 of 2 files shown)


BitProcessor:
Alternative: https://github.com/grafana/helm-charts/pull/697

zanhsieh (approved):

PR #585, opened 2021-07-28
[loki-distributed] Fix ingress behaviour

stevehipwell:
Fixes https://github.com/grafana/helm-charts/issues/574.

PR #700, opened 2021-09-23
[loki] Fix ingress behavior (closes #646)

### LINKED ISSUE #646: [loki] ingress should support "networking.k8s.io/v1"
The ingress in the loki graph should also support "networking.k8s.io/v1" as the version "networking.k8s.io/v1beta1" will be more and more deprecated (nginx-ingress-1.0.0 already removed support).

Probably all ingresses in this repo could be checked if they already support it or should be updated.

mgcrea:
Backported #585 to the standalone loki chart, aligned the `values.yaml` implementation to make updates easier in the future so it's a breaking change to update (adds support for `pathType`).

From:

```yaml
ingress:
  hosts:
    - host: loki.barberine.fr
      paths:
        - /
```

To:

```yaml
ingress:
  hosts:
    - host: loki.barberine.fr
      paths:
        - path: /
```

### DIFF
--- a/charts/loki/templates/_helpers.tpl
+++ b/charts/loki/templates/_helpers.tpl
@@ -59,3 +59,37 @@ Create the app name of loki clients. Defaults to the same logic as "loki.fullnam
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "loki.ingress.apiVersion" -}}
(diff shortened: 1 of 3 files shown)


onedr0p:
Any maintainer of this repository able to verify?

anthr76:
@owen-d any chance this can get looked at?

tuxpeople:
Would close #859 which holds me back from upgrading my cluster to Kubernetes 1.22. If there is an issue with this PR, I'm happy to provide one which "only" contains and fixes the API issue in 1.22+ if this would help to make progress here.

NeckBeardPrince:
Can we please get a review on this?

onedr0p:
Maybe this is fixed in https://github.com/grafana/helm-charts/pull/879

PR #714, opened 2021-09-29
[enterprise-metrics] Add compactor to gateway

techgnosis:
Before this PR the compactor showed as broken in the "Ring Health" tab of the GEM plugin. This was because the gateway did not have a compactor URL configured, and even if it did there was no Service in front of the compactor pods. 

This PR adds a Service for the compactor Pods and configures the gateway to use that Service to reach compactor.

### DIFF
--- a/charts/enterprise-metrics/CHANGELOG.md
+++ b/charts/enterprise-metrics/CHANGELOG.md
@@ -10,6 +10,11 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 1.5.4
+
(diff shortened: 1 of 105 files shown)


chaudum (commented):
LGTM 👏🏼 

To make the DCO check pass, you will need to sign your commit with `git commit --amend --no-edit --signoff`
Also, please run `make exports` in the `charts/enteprise-metrics` folder and add the changes in a 2nd commit

jdbaldry:
I've pushed the exports for convenience. @techgnosis, you only need to sign your commit now and this should be good to go.

jdbaldry (approved):
LGTM! Thanks

Issue #609, opened 2021-08-09
[loki] /tmp read-only file system


nick-oconnor:
Similar issue to #577 with `/tmp`. I'm getting bombarded with `level=warn ts=2021-08-09T00:03:54.191957114Z caller=marker.go:214 msg="failed to process marks" path=/data/loki/boltdb-shipper-compactor/retention/markers/1628441994354518113 err="open /tmp/marker-view-656536243: read-only file system"`

I've worked around it by adding:
```
extraVolumes:
  - name: temp
    emptyDir: {}
extraVolumeMounts:
  - name: temp
    mountPath: /tmp
```
A+ for allowing extra volume mounts!

pr0PM:
I'm also facing the exact same issue, but is this supposed to happen, needing us to mount extra, in that case we can add it directly in the loki-chart, or am I missing something?

Here is my loki-conf:
```
  auth_enabled: false
  chunk_store_config:
    max_look_back_period: 72h
  compactor:
    compaction_interval: 10m
    retention_delete_delay: 2h
    retention_delete_worker_count: 150
    retention_enabled: true
    shared_store: filesystem
    working_directory: /data/retention
  ingester:
    chunk_block_size: 262144
    chunk_idle_period: 60m
    chunk_retain_period: 1m
    chunk_target_size: 2097152
    lifecycler:
      ring:
        kvstore:
          store: inmemory
        replication_factor: 1
    max_transfer_retries: 0
  limits_config:
    cardinality_limit: 300000
    enforce_metric_name: false
    max_streams_per_user: 0
    reject_old_samples: false
    reject_old_samples_max_age: 168h
    retention_period: 72h
  schema_config:
    configs:
    - from: "2020-10-24"
      index:
        period: 24h
        prefix: index_
      object_store: filesystem
      schema: v11
      store: boltdb-shipper
  server:
    http_listen_port: 3100
  storage_config:
    boltdb:
      directory: /data/loki/index
    boltdb_shipper:
      active_index_directory: /data/loki/boltdb-shipper-active
      cache_location: /data/loki/boltdb-shipper-cache
      cache_ttl: 24h
      shared_store: filesystem
    filesystem:
      directory: /data/loki/chunks
  table_manager:
    retention_deletes_enabled: true
    retention_period: 72h
```

varac:
Please reopen, since #717 only fixes this for `charts/loki-distributed`, but not for `charts/loki`.

varac:
[...]

PR #717, opened 2021-09-30
[loki-distributed] R/W tmp folder for compactor

BitProcessor:
(replaces https://github.com/grafana/helm-charts/pull/713)

Fixes issue https://github.com/grafana/helm-charts/issues/609
Screenshot 2021-09-29 at 20 42 34


almson:
Why is this only for `loki-distributed`? Single-instance loki needs this change too.

PR #729, opened 2021-10-06
[enterprise-metrics] Ensure production values set ingester replication factor to three

jdbaldry:
Production values set the ingester replication factor to three to avoid data loss.
The resource calculations of these values already factored in this replication factor but did not apply it in the configuration.
If you have not reduced the compute resources in these values then this change should have no impact besides increased resilience to ingester failure.

Also, upgrade to [Grafana Enterprise Metrics v1.5.1](https://grafana.com/docs/metrics-enterprise/latest/downloads/#v151----september-21st-2021).


techgnosis (changes requested):
The ingester config is correct but you have added a second ingester block. The replication_factor should be set in the original ingester block yeah?

techgnosis:
My request for change was based on my not reading this correctly. LGTM

PR #734, opened 2021-10-08
[grafana] fix hpa for StatefulSet

TC-robV:
If you pick StatefulSet for persistence then hpa no longer works as it tried to looks for a Deployment.


zanhsieh:
@TC-robV 
Could you bump the chart version to latest and sign DCO please?

PR #755, opened 2021-10-23
Add stable ingress apiVersion support to Loki

rtrox:
As described, when available, use apiVersion "networking.k8s.io/v1" for ingress. Otherwise, follow existing logic.

Helper function here is modified from the grafana ingress and modified for loki.

Signed-off-by: Russell Troxel <russelltroxel@gmail.com>


rtrox:
looks like this was fixed in #879

PR #769, opened 2021-11-02
[agent-operator] Update Agent Operator to v0.20.0

hjet:
Closes https://github.com/grafana/agent/issues/1024

- Update operator images
- Add toggle to `values.yaml` for `kubelet-service`

### DIFF
--- a/charts/agent-operator/templates/operator-deployment.yaml
+++ b/charts/agent-operator/templates/operator-deployment.yaml
@@ -40,6 +40,10 @@ spec:
         resources:
         {{- toYaml . | nindent 10 }}
         {{- end }}
+        args:
+          {{- if .Values.kubeletService.enabled }}
+          - --kubelet-service={{ .Values.kubeletService.namespace }}/{{ template "ga-operator.fullname" . }}-kubelet
+          {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
       {{- toYaml . | nindent 8 }}
--- a/charts/agent-operator/templates/operator-deployment.yaml
+++ b/charts/agent-operator/templates/operator-deployment.yaml
@@ -40,10 +40,10 @@ spec:
         resources:
         {{- toYaml . | nindent 10 }}
         {{- end }}
+        {{- if and .Values.kubeletService.namespace .Values.kubeletService.serviceName }}
         args:
-          {{- if .Values.kubeletService.enabled }}
-          - --kubelet-service={{ .Values.kubeletService.namespace }}/{{ template "ga-operator.fullname" . }}-kubelet
-          {{- end }}
+          - --kubelet-service={{ .Values.kubeletService.namespace }}/{{ .Values.kubeletService.serviceName }}
+        {{- end }}
       {{- with .Values.nodeSelector }}
       nodeSelector:
       {{- toYaml . | nindent 8 }}
--- a/charts/agent-operator/Chart.yaml
+++ b/charts/agent-operator/Chart.yaml
@@ -2,9 +2,9 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.1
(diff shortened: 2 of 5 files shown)


charts/agent-operator/templates/operator-deployment.yaml
```diff
@@ -40,6 +40,10 @@ spec:
         resources:
         {{- toYaml . | nindent 10 }}
         {{- end }}
+        args:
+          {{- if .Values.kubeletService.enabled }}
+          - --kubelet-service={{ .Values.kubeletService.namespace }}/{{ template "ga-operator.fullname" . }}-kubelet
```
rfratto, line 45: Hmm, will `{{ template "ga-operator.fullname" . }}` produce an obvious value for users? Should we maybe just have the whole thing be configurable? Generally you'd probably want `kube-system/kubelet` here (or similar) just to make it easier to write ServiceMonitors that consume the service.

rfratto: (And maybe if the whole value is configurable, we can just check to see if it's non-empty instead of having an enabled field? WDYT?)

charts/agent-operator/templates/operator-deployment.yaml
```diff
@@ -40,6 +40,10 @@ spec:
         resources:
         {{- toYaml . | nindent 10 }}
         {{- end }}
+        args:
```
rfratto, line 43: Does this still work if there are no arguments?

charts/agent-operator/templates/operator-deployment.yaml
hjet: > Hmm, will `{{ template "ga-operator.fullname" . }}` produce an obvious value for users? Should we maybe just have the whole thing be configurable? Generally you'd probably want `kube-system/kubelet` here (or similar) just to make it easier to write ServiceMonitors that consume the service.

I just followed the [Prom operator Helm chart](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus-operator/deployment.yaml#L42) here. In [our own docs](https://grafana.com/docs/agent/latest/operator/getting-started/#install-agent-operator-on-kubernetes) we use `default/kubelet`

charts/agent-operator/templates/operator-deployment.yaml
hjet: Will check

charts/agent-operator/templates/operator-deployment.yaml
hjet: I can make those changes tomorrow

charts/agent-operator/templates/operator-deployment.yaml
hjet: Didn't need to check!

charts/agent-operator/templates/operator-deployment.yaml
hjet: I think a loose "Helm convention" is to have things Helm manages also have the Helm release name in the K8s object name (in this case a Service). However I don't think it's that important, so I've gone ahead with the above changes.

rfratto: > I think a loose "Helm convention" is to have things Helm manages also have the Helm release name in the K8s object name (in this case a Service). However I don't think it's that important, so I've gone ahead with the above changes.

I think that's fair and completely agree FWIW, I'm just concerned about whether it'll be easy for users to find what the service name ends up being :)

rfratto (approved):
Thanks! 

One more thing: can you check if the CRDs have changed (if you haven't already?) I think there may have been one or two typo fixes.

hjet:
> Thanks!
> 
> One more thing: can you check if the CRDs have changed (if you haven't already?) I think there may have been one or two typo fixes.

Yep, checked that, there are no changes. I hope to automate all this soon as part of the agent release process...

PR #785, opened 2021-11-08
[loki-distributed] FIX: nginx dns resolve on proxy_pass

joahking:
# The problem

the current proxy_pass directives in the gateway are resolving upstream address when nginx starts/loads-configuration
so when a backend/pod changes address the logs show errors with "dns resolve timeout" 
and an nginx restart/reload is required to re-resolve addresses

# The fix

use variables instead of addresses so that nginx re-resolve domain names when TTLs expire

# Documentation of nginx 

https://www.nginx.com/blog/dns-service-discovery-nginx-plus/
- section "Using a Domain Name in an Upstream Server Group" is the current way that resolves at load time
- section "Setting the Domain Name in a Variable" is this fix here

*Note*: I am not sure if I got it correctly regarding the version, please help me fix it


joahking:
an example of error in log is:
`... [error] 9#9: *80742 xxx-loki-loki-distributed-query-frontend.logging.svc.xxx could not be resolved (110: Operation timed out),...`

zanhsieh (approved):

Issue #783, opened 2021-11-06
[promtail] chart version 3.9.0 is missing from repo index


yafanasiev:
Hi!

It looks like something broke during the release of the new Promtail 3.9.0 chart. [This commit](https://github.com/grafana/helm-charts/commit/6f3944d109da0859bc9d9fa8d9433ba8a8ef64e7) that was created via Github Actions adds it, but then the [next one](https://github.com/grafana/helm-charts/commit/1cdf1c4c791100f587a6ea765634139cffd3ef3f) removes it.

Legion2:
The CI pipeline for the release commit is failed https://github.com/grafana/helm-charts/runs/4122255988?check_suite_focus=true

stevehipwell:
@torstenwalter I think this has happened again, specifically for `loki-distributed` but potentially other charts are impacted.

yafanasiev:
Indeed - latest `promtail` 3.9.1 is up, but `loki-distributed` 0.39.1 disappeared. This is extremely annoying when using GitOps tools (we use Flux) as this generates alerts to our monitoring system. Looking at [Github Action](https://github.com/grafana/helm-charts/blob/main/.github/workflows/release.yaml) for releasing charts, I would suggest adding [concurrency](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency) to workflow to ensure there are no race conditions. I can make a PR for this.

stevehipwell:
@yafanasiev concurrency sounds sensible. I'm trying to figure out how the chart releaser tooling is overwriting commits when the code is meant to be doing a plain commit.

PR #793, opened 2021-11-08
ci: run only one concurrent release workflow

yafanasiev:
Ensures that only one concurrent `release` workflow is running at the moment to prevent race condition when updating Helm repository index file.

Fixes https://github.com/grafana/helm-charts/issues/783.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -5,6 +5,8 @@ on:
     branches:
       - main
 
+concurrency: release
+
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


yafanasiev:
@stevehipwell

torstenwalter:
@yafanasiev  we had the same idea. It's better than now, but still not a perfict solution see my comments in #794

yafanasiev:
Yep. Closing in favor of https://github.com/grafana/helm-charts/pull/794

PR #800, opened 2021-11-09
[loki] Make entire tmp dir writable

alexnederlof:
Because `securityContext.readOnlyRootFilesystem: true` boltdb cannot write to `tmp` which it seems to do to update its markers. The log if filled with errors that it cannot perform a write operation in `tmp`. By changing the scratch directory to the entire `tmp` folder, the problem is fixed because the empty folder is writable by the loki user.


almson:
This is a real issue eg for persistence, but this PR is wrong since it only mounts tmp if there's `alerting_groups`.

zanhsieh:
Closed in favour of https://github.com/grafana/helm-charts/pull/967.

PR #807, opened 2021-11-11
[grafana] make datasource sidecar watchMethod configurable

D3luxee:
This PR makes the watchMethod for the datasource sidecar configurable like the dashboard sidecar.
The current default behavior is kept.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,8 +1,8 @@
 apiVersion: v2
 name: grafana
-version: 6.17.5
+version: 6.17.6
 appVersion: 8.2.3
-kubeVersion: '^1.8.0-0'
+kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
 home: https://grafana.net
 icon: https://raw.githubusercontent.com/grafana/grafana/master/public/img/logo_transparent_400x.png
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -87,7 +87,7 @@ initContainers:
     imagePullPolicy: {{ .Values.sidecar.imagePullPolicy }}
     env:
       - name: METHOD
-        value: LIST
+        value: {{ .Values.sidecar.datasources.watchMethod }}
       - name: LABEL
         value: "{{ .Values.sidecar.datasources.label }}"
       {{- if .Values.sidecar.datasources.labelValue }}
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -643,6 +643,8 @@ sidecar:
     # Otherwise the namespace in which the sidecar is running will be used.
     # It's also possible to specify ALL to search in all namespaces.
     searchNamespace: null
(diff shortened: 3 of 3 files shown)


zanhsieh:
@D3luxee 
Could you bump the chart version and sign DCO please?

D3luxee:
Hi, i wasn't sure if i should bump the chart or if you guys do it only on releases.
Now i've bumped it from 6.17.5 to 6.17.6 and signed off the DCO for all commits.

zanhsieh (approved):

xeor:
I'm not sure if this works as intended. Setting `sidecar.datasources.watchMethod` to `WATCH` for me on chart `6.18.1` makes grafana not able to start..

zanhsieh:
@D3luxee 
Could you look into @xeor 's case please ?

D3luxee:
I can confirm this issue, it seems i missed the fact that the datasource sidecar is an init container and the k8s-sidecar will not exit in watch mode.

I was using this chart as part of kube prometheus stack on friday this was finally updated to use the latest version of the grafana chart. I've run into the same issue as @xeor and spend the friday with debugging and figuring out how to resolve it.

There are a few things to consider in trying to make the chart able to dynamically provision datasources.

Grafana does not watch for new datasources in the provisioning folder, it does poll new dashboards but not new datasources!
This means that Grafana will only load datasources at startup. https://grafana.com/docs/grafana/latest/administration/provisioning/#data-sources

This might be the reason why this was on watchMethod LIST and as an init container.. but there is still a way to get it working!

Grafana supports reloading the provisioned datasources via api and the k8s sidecar supports sending requests when a config map changed. 
https://grafana.com/docs/grafana/latest/http_api/admin/#reload-provisioning-configurations
https://github.com/kiwigrid/k8s-sidecar#configuration-environment-variables

I can provide a configuration example here for others to find and we could discuss if this should be added to the chart. 
The k8s-sidecar would need to be configured with grafana admin credentials to allow it reloading via admin api.

D3luxee:
@xeor 
The settings below will define a second container which will watch for new datasources.
What happens is:
1. original datasource sidecar will start as init container and provision datasources
2. the extra datasource sidecar will start in WATCH mode and add any new datasources and trigger a reload via api
```
    grafana:
      extraContainers: |
        - name: grafana-dynamic-datasources
          image: "quay.io/kiwigrid/k8s-sidecar:1.14.2"
          imagePullPolicy: IfNotPresent
          env:
            - name: METHOD
              value: WATCH
            - name: LABEL
              value: "dynamic_datasource"
            - name: FOLDER
              value: "/etc/grafana/provisioning/datasources"
            - name: RESOURCE
              value: "both"
            - name: NAMESPACE
              value: ALL
            - name: REQ_URL
              value: http://grafana/api/admin/provisioning/datasources/reload
            - name: REQ_METHOD
              value: POST
            - name: REQ_USERNAME
              value: admin
            - name: REQ_PASSWORD
              value: YourGrafanaAdminPassword
          resources: {}
          volumeMounts:
            - name: sc-datasources-volume
              mountPath: "/etc/grafana/provisioning/datasources"
```

Issue #808, opened 2021-11-11
Querier deployment has unused volume mount when index-gateway is enabled


naveedyahyazadeh:
Describe the bug:
I noticed that, after enabling the index-gateway, enabling persistence for the index-gateway, and moving querier from STS to deployment with no persistence, that the pod was still using ephemeral storage on its node. After searching through the pod's files, I noticed that the `/var/loki` volume mount is still active on the querier but it is not being used. 

Code for querier deployment's volume mount: https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/templates/querier/deployment-querier.yaml#L80

Code for index-gateway's volume mount https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/templates/index-gateway/statefulset-index-gateway.yaml#L73

`ls -lsa` on /var/loki in querier pod:
![image](https://user-images.githubusercontent.com/54322109/141338131-7ebf9bd9-7ca4-4dca-ab70-d7dbd2a8ec1c.png)

`ls -lsa` on /var/loki in index-gateway pod:
![image](https://user-images.githubusercontent.com/54322109/141338323-92594a6d-8efc-476f-99e5-dbe82ced09cd.png)

I believe that we can remove the volume mount from the querier deployment when the index gateway is enabled.

sboschman:
Change #809 breaks the out-of-the-box usage of the querier fifo cache:
```
msg="error running loki" err="mkdir /var/loki: read-only file system\nerror creating index client
```

From a still running pod form chart v0.39.3
```
$ du -ch /var/loki
4.2G    /var/loki/cache
4.2G    /var/loki
4.2G    total
```

patsevanton:
Same issue

PR #809, opened 2021-11-11
[loki-distributed] Remove /var/loki volume mount from querier deployment

naveedyahyazadeh:
Re: #808, 

This PR removes the `/var/loki` A.K.A `data`  volume mount from the querier deployment when the index-gateway is enabled.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.1
-version: 0.39.3
+version: 0.39.4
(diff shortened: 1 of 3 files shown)


charts/loki-distributed/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.1
-version: 0.39.2
+version: 0.39.3
```
Whyeasy, line 6: Please update the chart version.

naveedyahyazadeh: Done!

unguiculus (approved):

PR #815, opened 2021-11-12
[grafana] Remove duplicate use of extraEmptyDirMounts

4orty:
related to #362 
extraVolumeMounts contains extraEmptyDirMounts so extraEmptyDirMounts is unnecessary.

It's my first time to PR. Please tell me if there's anything against Contributing Guidelines.

### DIFF
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -120,7 +120,6 @@ This version requires Helm >= 3.1.0.
 | `extraSecretMounts`                       | Additional grafana server secret mounts       | `[]`                                                    |
 | `extraVolumeMounts`                       | Additional grafana server volume mounts       | `[]`                                                    |
 | `extraConfigmapMounts`                    | Additional grafana server configMap volume mounts | `[]`                                                |
-| `extraEmptyDirMounts`                     | Additional grafana server emptyDir volume mounts | `[]`                                                 |
 | `plugins`                                 | Plugins to be loaded along with Grafana       | `[]`                                                    |
 | `datasources`                             | Configure grafana datasources (passed through tpl) | `{}`                                               |
 | `notifiers`                               | Configure grafana notifiers                   | `{}`                                                    |
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -309,10 +309,6 @@ containers:
         subPath: {{ .subPath | default "" }}
         readOnly: {{ .readOnly }}
     {{- end }}
-    {{- range .Values.extraEmptyDirMounts }}
-      - name: {{ .name }}
-        mountPath: {{ .mountPath }}
-    {{- end }}
     ports:
       - name: {{ .Values.service.portName }}
         containerPort: {{ .Values.service.port }}
@@ -511,10 +507,6 @@ volumes:
     emptyDir: {}
     {{- end }}
 {{- end }}
-{{- range .Values.extraEmptyDirMounts }}
-  - name: {{ .name }}
-    emptyDir: {}
-{{- end -}}
 {{- if .Values.extraContainerVolumes }}
 {{ toYaml .Values.extraContainerVolumes | indent 2 }}
 {{- end }}
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -103,12 +103,6 @@ extraConfigmapMounts: []
   #   configMap: certs-configmap
   #   readOnly: true
 
-
-extraEmptyDirMounts: []
-  # - name: provisioning-notifiers
-  #   mountPath: /etc/grafana/provisioning/notifiers
-
-
 # Apply extra labels to common labels.
 extraLabels: {}
 


zanhsieh:
@4orty 
I cannot justify what you proposed to change. The volume and directory are two different things in k8s. Empty directories mostly is for cache or some temporary files. Removing them I don't see how you could get it back from volume.

4orty:
> @4orty I cannot justify what you proposed to change. The volume and directory are two different things in k8s. Empty directories mostly is for cache or some temporary files. Removing them I don't see how you could get it back from volume.

Thanks for your replying!
In #362, we can add emptyDir in extraVolumeMounts right above extraEmptyDirMounts.
So I thought these two were overlapping.
Well.. do these two mean something different?? ( extraVolumeMounts's emptyDir and extraEmptyDirMounts's emptyDir )

```
{{- range .Values.extraVolumeMounts }}
  - name: {{ .name }}
    {{- if .existingClaim }}
    persistentVolumeClaim:
      claimName: {{ .existingClaim }}
    {{- else if .hostPath }}
    hostPath:
      path: {{ .hostPath }}
    {{- else }}
    emptyDir: {}
    {{- end }}
{{- end }}
```

zanhsieh:
> > @4orty I cannot justify what you proposed to change. The volume and directory are two different things in k8s. Empty directories mostly is for cache or some temporary files. Removing them I don't see how you could get it back from volume.
> 
> Thanks for your replying! In #362, we can add emptyDir in extraVolumeMounts right above extraEmptyDirMounts. So I thought these two were overlapping. Well.. do these two mean something different?? ( extraVolumeMounts's emptyDir and extraEmptyDirMounts's emptyDir )
> 
> ```
> {{- range .Values.extraVolumeMounts }}
>   - name: {{ .name }}
>     {{- if .existingClaim }}
>     persistentVolumeClaim:
>       claimName: {{ .existingClaim }}
>     {{- else if .hostPath }}
>     hostPath:
>       path: {{ .hostPath }}
>     {{- else }}
>     emptyDir: {}
>     {{- end }}
> {{- end }}
> ```

They are two different things. What if someone would like to mount a volume and have a temporary directory at the same time?

4orty:
> > > @4orty I cannot justify what you proposed to change. The volume and directory are two different things in k8s. Empty directories mostly is for cache or some temporary files. Removing them I don't see how you could get it back from volume.
> > 
> > 
> > Thanks for your replying! In #362, we can add emptyDir in extraVolumeMounts right above extraEmptyDirMounts. So I thought these two were overlapping. Well.. do these two mean something different?? ( extraVolumeMounts's emptyDir and extraEmptyDirMounts's emptyDir )
> > ```
> > {{- range .Values.extraVolumeMounts }}
> >   - name: {{ .name }}
> >     {{- if .existingClaim }}
> >     persistentVolumeClaim:
> >       claimName: {{ .existingClaim }}
> >     {{- else if .hostPath }}
> >     hostPath:
> >       path: {{ .hostPath }}
> >     {{- else }}
> >     emptyDir: {}
> >     {{- end }}
> > {{- end }}
> > ```
> 
> They are two different things. What if someone would like to mount a volume and have a temporary directory at the same time?

I tested in the following way.

# values.yaml
```
extraVolumeMounts:
   - name: extra-volume-0
     mountPath: /mnt/volume0
     readOnly: true
     existingClaim: volume-claim
   - name: extra-volume-1
     mountPath: /mnt/volume1
     readOnly: true
     hostPath: /usr/shared/
   - name: extra-empty-volume
     mountPath: /mnt/volume2
```

And helm template result is
# Source: grafana/templates/deployment.yaml
```
      containers:
        - name: grafana
          image: "grafana/grafana:8.2.3"
          imagePullPolicy: IfNotPresent
          volumeMounts:
            - name: extra-volume-0
              mountPath: /mnt/volume0
              subPath: 
              readOnly: true
            - name: extra-volume-1
              mountPath: /mnt/volume1
              subPath: 
              readOnly: true
            - name: extra-empty-volume
              mountPath: /mnt/volume2
              subPath: 
              readOnly: 
        volumes:
          - name: extra-volume-0
            persistentVolumeClaim:
              claimName: volume-claim
          - name: extra-volume-1
            hostPath:
            path: /usr/shared/
          - name: extra-empty-volume
            emptyDir: {}
```

At first, i want to just delete 'else' phrase on extraVolumeMounts but according to k8s api/types.go, emptyDir and pvc and hostPath are all classified as volume.
So i thought emptyDir is right to be in extraVolumeMounts rather than in extraVolumeMounts.

[kubernetes types.go](https://github.com/kubernetes/api/blob/master/core/v1/types.go)

zanhsieh:
@4orty 
How do you ensure the backward compatibility? For example, somebody already defined extraEmptyDirMounts.

If you really think this is a must-go feature, then please bump minor Chart version.

4orty:
> @4orty How do you ensure the backward compatibility? For example, somebody already defined extraEmptyDirMounts.
> 
> If you really think this is a must-go feature, then please bump minor Chart version.

The backward compatibility is one of the things that i was most concerned about.
In fact, this PR is more close to clean code(not add features) as just delete duplicate feature and not impact on the usage of chart or the result of the kubernetes resource(deployment).
In current chart, one of the two(clean code and backward compatibility) canot be maintained. (To keep both of them, we should have deleted else phrase in #362 PR.
 
I don't know which philosophy is this chart values ​​more. If this chart value the backward compatibility more, please close this PR but if this chart value the claen code more, i'll bump minor Chart version.

zanhsieh:
Close this PR because no good solution for handling backward compatibility.

PR #452, opened 2021-05-26
[loki-distributed] Add relabelings to ServiceMonitor

kamilhristov:
Example:

```yaml
serviceMonitor:
  enabled: true
  relabelings:
  - separator: /
    sourceLabels:
    - namespace
    - __meta_kubernetes_pod_label_app_kubernetes_io_component
    targetLabel: job
```

PR #818, opened 2021-11-13
[tempo-distributed] Add relabeling configuration to ServiceMonitors

KingJ:
Similar to https://github.com/grafana/helm-charts/pull/452, this adds support for configuring a ServiceMonitor's relabelling behaviour. This is especially useful for anyone who uses Grafana's various mixins for dashboards, as these expect a job dimension that includes the namespace (see https://github.com/grafana/helm-charts/pull/452 for a sample `relabelings` config that achieves this).

This also needs to be done for the Promtail chart, but there is an existing open PR for this - https://github.com/grafana/helm-charts/pull/766.


mapno (approved):
Thanks for the contribution! LGTM

PR #820, opened 2021-11-14
Set namespace for all tempo non-clustered resources

willthames:
Ensure that namespace gets correctly set during resource creation

### DIFF
--- a/charts/tempo-distributed/templates/compactor/service-compactor.yaml
+++ b/charts/tempo-distributed/templates/compactor/service-compactor.yaml
@@ -2,6 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: {{ template "tempo.compactorFullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "tempo.compactorLabels" . | nindent 4 }}
 spec:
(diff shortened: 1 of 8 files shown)


Whyeasy:
Hi @willthames ,

Thank you for your pull request. Adding the namespace inside the Helm templates is a bad practice. Deploying to the correct namespace should be taken care of by your deployment method/system. 

This was requested before in Loki and we reverted this.

Will close the PR.

willthames:
@whyeasy is that documented anywhere? As I know of at least two deployment systems that I work with heavily that fail if namespace isn't set in a manifest file (and if both are wrong, I can probably get both fixed).

I usually try and avoid setting namespace outside of the resource manifest files, and relying on current namespace when using e.g. kubectl seems very hazardous.

There are 50+ files currently using `namespace: {{ .Release.Namespace }}` - is there a plan to remove that from those files?

Whyeasy:
So I don't think it's anywhere properly documented. But I'm not the only one who think this a bad thing:

https://github.com/grafana/helm-charts/pull/627#issuecomment-913450661

Regarding the other files. I'm not a maintainer on all charts and I haven't been active the last couple of months. So I need to check the charts and will try to revert the ones I maintain. 

I understand that relying on current namespace can be a pain. But in most cases pipelines or applications like Flux/Gitops toolkit and Argo are used to apply these resources. 

Sorry for the inconvenience, but I would like to keep the namespace out of the templates.

@unguiculus anything to add?

Issue #487, opened 2021-06-09
Tempo: enable more custom fields


guyguy333:
Hello,

Currently, we can't pass custom values to tempo server in chart. Field is currently limited to http listener port. Maybe adding something like `{{- toYaml .Values.tempo.server | nindent 6 }}` below `http_listen_port` will be good enough ?

Thanks

jleloup:
I would also like to be able to provide more custom configuration to the Tempo server (`grpc_server_max_recv_msg_size` comes to mind there).

I can provide a PR quite soon to do so.

PR #822, opened 2021-11-15
[tempo] Allow custom configuration of Tempo server

jleloup:
As described in issue [#487](https://github.com/grafana/helm-charts/issues/487) it would be nice to be able to provide some custom configuration to Tempo server in the single binary chart.

In order to solve that we can use a similar approach to what is already available for other Tempo sub components (like ingester) so a `.Values.tempo.server ` could be used to provide YAML configuration to the server.

Note: to make this change consistent the previous value `httpListenPort: 3100` is modified to be set as any other server configuration.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.10.0
+version: 0.10.1
 appVersion: 1.2.0
 engine: gotpl
 home: https://grafana.net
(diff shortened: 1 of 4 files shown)


joe-elliott:
Thanks for the contribution!

This is technically a breaking change. Since our major version < 1 we don't technically have to update it, but maybe just update the minor rev instead of patch?

jleloup:
> Thanks for the contribution!
> 
> This is technically a breaking change. Since our major version < 1 we don't technically have to update it, but maybe just update the minor rev instead of patch?

Sure I'll go for a minor rev.
I wonder if I need to also update some documentation ?

joe-elliott:
Apologies for the ci checks. I think the last thing you need to do is just run helm-docs. I do this:
```
```
Thank you for sticking with this.

jleloup:
You're welcome :)
I think the Helm documentation is now up to date

joe-elliott (approved):

PR #826, opened 2021-11-16
[enterprise-logs] Fix dev cluster MinIO endpoints in default configuration.

chaudum:
Services where incorrectly configured when using the default MinIO configuration,
which pointed to `{{ .Release.Name }}-enterprise-logs-minio` instead of `{{ .Release.Name }}-minio`.

Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

### DIFF
--- a/charts/enterprise-logs/CHANGELOG.md
+++ b/charts/enterprise-logs/CHANGELOG.md
@@ -11,7 +11,7 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
-## Unreleased
+## 1.2.1
 
 - [BUGFIX] Fixed the dev cluster MinIO endpoints in the default configuration. #826
 
--- a/charts/enterprise-logs/CHANGELOG.md
+++ b/charts/enterprise-logs/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## Unreleased
+
+- [BUGFIX] Fixed the dev cluster MinIO endpoints in the default configuration. #826
+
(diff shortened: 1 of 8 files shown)


charts/enterprise-logs/CHANGELOG.md
```diff
@@ -11,6 +11,12 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## Unreleased
+
+## 1.2.1
+
+- [BUGFIX] Fixed the MinIO endpoints in the resultiung `config.yaml`, which pointed incorrectly to `{{ .Release.Name }}-enterprise-logs-minio` instead of `{{ .Release.Name }}-minio`. #826
```
jdbaldry, line 18: There's a typo in "resulting". 

nit: I would consider something simpler like:
```suggestion
- [BUGFIX] Fixed the dev cluster MinIO endpoints in default configuration. #826
```
or just replacing the reference to `config.yaml` with something like "GEL configuration file" since the latter is more likely to be understood by users.

charts/enterprise-logs/templates/_helpers.tpl
```diff
@@ -75,3 +75,10 @@ Create the app name of enterprise-logs clients. Defaults to the same logic as "e
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Create the service URL including port for MinIO.
```
jdbaldry, line 80: nit: I think a URL must have a scheme. I think this endpoint is more technically correct.

jdbaldry (approved):
Nice fix, much neater too!

jdbaldry (approved):
LGTM!

PR #827, opened 2021-11-16
[tempo-distributed] allow Tempo server custom configuration

jleloup:
I would like to solve the same issue as [#487](https://github.com/grafana/helm-charts/issues/487) though for the distributed Tempo chart.

Similarly to https://github.com/grafana/helm-charts/pull/822:

* introduce a `server` value section to be included in the overall `config`
* also add log level & format parameters as they are quite useful

To make it more consistent with the single binary chart: we could also move this server section under `tempo.server` instead though it seems to me the distributed chart favors atomic part of the configuration as the YAML root level.


charts/tempo-distributed/values.yaml
```diff
@@ -449,6 +447,13 @@ config: |
         service: memcached-client
         timeout: 500ms
 
+# Set Tempo server configuration
+# Refers to https://grafana.com/docs/tempo/latest/configuration/#server
+server:
+  http_listen_port: 3100
+  log_level: info
+  log_format: logfmt
```
annanay25, line 455: FYI these two are defaults already. Not sure if you intended to include them anyway

jleloup: It was intentional to explicitly let new users know that they can tweak logging through these values.

Though I can remove it if you prefer.

annanay25: OK. Might be good to add a comment that these are defaults but can be changed. 

Also CI is complaining about the docs. Once that is fixed we can merge this!

joe-elliott (approved):
Thanks for the contribution! Let's go!

PR #828, opened 2021-11-16
[agent-operator] Fix agent-operator tests setting an empty MetricsInstance spec

cippaciong:
Because:

* Having a null `spec:` definition in the agent-operator MetricsInstance
  used for test, causes the test to fail with the following error during
  the apply of the manifest: 'The MetricsInstance "primary" is invalid:
  spec: Invalid value: "null": spec in body must be of type object:
  "null"'

This commit:

* Sets the value to an empty object `spec: {}` rather than null

Signed-off-by: Tommaso Sardelli <lacapannadelloziotom@gmail.com>

### DIFF
--- a/charts/agent-operator/Chart.yaml
+++ b/charts/agent-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.2
+version: 0.1.3
 appVersion: "0.20.0"
 home: https://grafana.com/docs/agent/latest/
 icon: https://raw.githubusercontent.com/grafana/agent/v0.20.0/docs/assets/logo_and_name.png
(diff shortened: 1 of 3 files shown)


cippaciong:
@hjet @rfratto Sorry for the direct mention, I hoped you could have a look at this.

Unfortunately when you deploy the helm chart using [terraform-provider-helm](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#disable_webhooks) it will automatically run the test and mark the installation as failed even if it's not, so this is kind of a blocker for me 😓

hjet:
Thanks for this! Looks like both the test CI cluster and k3s both don't support OpenAPIV3 schema validation...

Can you run `helm-docs` and commit the changes? That'll get the last check passing. 

We didn't really test this with the terraform provider so there may be some other weirdness (usually objects deployed by `helm test` don't get cleaned up) - please file any other issues in https://github.com/grafana/agent and we'll take a look.

hjet:
This may also solve the `helm test` cleanup issue: https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies

cippaciong:
Thanks for the feedback @hjet 🙂 
I updated the README with `helm-docs` and added the `hook-delete-policy` annotations, let me know if there is anything else I can do.

hjet (approved):
lgtm!

hjet:
Thanks so much! I tested locally and for some reason the resources were not cleaned up after the `helm test` hook passed (this was the case before as well). But fixing this is out of scope for this CL so I will merge - just be careful in your own clusters with straggler resources!

cippaciong:
Thanks 🙂

PR #829, opened 2021-11-16
[tempo] Fix affinity toYaml call

renatomjr:
Fix `<toYaml>: wrong number of args for toYaml: want 1 got 2` error when setting affinity.


joe-elliott (approved):
Nice! Thanks for the fix.

PR #831, opened 2021-11-17
[tempo-distributed] allow configuration of the Kafka receiver

jleloup:
I would like to be able to configure a Kafka receiver in Grafana Tempo.

To do so I propose to use a set of values in the current `traces` map & include it in the overall `config` value.

While this approach helps to configure all possible Kafka parameters without duplicating them as Helm values, it breaks consistency with other receivers which are boolean only.


joe-elliott:
Yeah, I don't know why the other receivers are just true/false. I much prefer this way. Let's get the conflicts cleaned up and we'll merge.

joe-elliott (dismissed):

annanay25 (commented):
@jleloup some other PRs were merged recently and are now conflicting with these changes. Please resolve and we will merge!

jleloup:
> @jleloup some other PRs were merged recently and are now conflicting with these changes. Please resolve and we will merge!

Sure thing, I have resolved those conflicts. I think it is now up to date.

mapno (approved):
lgtm, thanks!

PR #832, opened 2021-11-18
[enterprise-logs] Bump enterprise-logs version to 1.3.0

chaudum:
This version bump is necessay, because in a recent change (commit
1428e05458fb8f180b9d71199d4fa916ea0ff0cb) the GEL image version was
bumped to v1.2.0 by accident.

Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

PR #840, opened 2021-11-22
[enterprise-logs] Fix template rendering bug

chaudum:
Commit d75c89a164ff96f3a1b400ab59c0d2bac67896cf introduced a bug that caused an error when MinIO was disabled. It was caused by the `"enterprise-logs.minio"` template macro, which tried to sus the service port configuration even when `minio.enabled` was set to `false`.

---

This PR includes PR #832

### DIFF
--- a/charts/enterprise-logs/CHANGELOG.md
+++ b/charts/enterprise-logs/CHANGELOG.md
@@ -11,7 +11,7 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
-## Unreleased
+## 1.3.1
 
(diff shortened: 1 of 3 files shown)


jdbaldry (approved):
Sorry for missing this on the first review. I didn't test with minio disabled. This looks like it will absolutely solve the problems described.

I should point out that I haven't yet been able to spend the time testing out the 1.3.0 upgrade yet.

chaudum:
Seems like rebasing dismisses existing reviews @jdbaldry

jdbaldry:
It does, taking another browse.

jdbaldry (approved):

PR #844, opened 2021-11-23
Remove spurious port definition on Grafana Pods.

mt-inside:
The Grafana Pods declare two ports. One is hard-coded to 3000 (translated to 80 by the default value for the Service port). The other adds the _Service_ port details to the _Pods_. This looks like a copy-and-paste error?

This is causing me real issues - leaving `Values.service.port` set to `80` just results in a spurious port definition, but setting it to 3000 (where a lot of systems expect grafana to be) causes TWO ports 3000 to be declared on the grafana Deployment. kubectl is happy to apply those manifests, but some things eg Flux 2 do more rigours checks and reject the resource.

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -314,9 +314,6 @@ containers:
         mountPath: {{ .mountPath }}
     {{- end }}
     ports:
-      - name: {{ .Values.service.portName }}
-        containerPort: {{ .Values.service.port }}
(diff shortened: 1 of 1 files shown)


craigb-zencore:
Have same issue when using GCP Anthos Config Manager (ACM). It complains about the duplicate port and will not deploy the resource. ACM uses Kustomize and Knative under the hood. It's possible one of those is responsible for the error.

PR #848, opened 2021-11-25
[tempo] Align service with tempo default config

Samuron:
I find it confusing that now tempo is listening for OTLP http traces on port 4318 by default, but this port is not exposed by the service


annanay25 (commented):
Thanks @Samuron. Please also bump the charts minor version.

Samuron:
@annanay25 done

zanhsieh (approved):

PR #855, opened 2021-11-26
[tempo-distributed] Switch otlp http port from 55681 => 4318

tete17:
I based this MR on reading the opentelementry protocol specification, in particular [here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-default-port).

I found this issue while manually adding an ingress to expose the http endpoint of tempo and realising the port's did not match

I consider this a bug fix and in particular not a breaking change for both helm charts for the following reasons:

# Standalone
You may consider this a breaking change since some users may depend on the port number but the default configuration listens on `4318` already. This means, that unless the user has manually override the port number on the configuration, the http service port doesn't work since the pod is not listening there. TLDR it was already broken.

We only risk breaking users who already noticed this bug and have either not reported it or I am the first one to notice. In which case a simple removal of the specific port configuration should suffice (like I am planning to do).

# Distributed
In this case we have an ingress and a nginx instance as a proxy. The ports numbers are only internal so this change should be invisible for anyone upgrading the chart.

I am also open to consider these changes more breaking changes or add extra services like `-legacy` prefixes but given the reasons above I wanted to avoid doing that in favour of cleanliness.


joe-elliott:
CI will only allow a PR to touch one chart at a time. Perhaps drop the `tempo` part of this PR as it looks like someone attempted something similar here:

https://github.com/grafana/helm-charts/pull/848

zanhsieh (approved):

Issue #856, opened 2021-11-26
Make isDefault configurable in loki-stack data sources


jdechicchis:
Grafana data sources can be configured to be the default: https://github.com/grafana/helm-charts/blob/main/charts/grafana/values.yaml#L459

However, when using loki-stack the `isDefault` value cannot be configured. It would be nice for it to be configurable https://github.com/grafana/helm-charts/blob/main/charts/loki-stack/templates/datasources.yaml#L17

jdechicchis:
I'm happy to make a PR to add this feature as well

PR #857, opened 2021-11-26
[loki-stack] Make isDefault configurable in loki-stack data sources

jdechicchis:
#856


trevorwhitney (commented):
What happens if you set more than one datasource `isDefault: true`? It would be a bit nicer if there was just one datasource that could be selected here.

zanhsieh (approved):

Issue #853, opened 2021-11-26
[enterprise-logs] ingester can not start using default configuration


vlad-diachenko:
If we try to install `enterprise-logs` based on `docker.io/grafana/enterprise-logs:v1.2.0` chart with this configuration(default with small changes) :
<details>
  <summary>values.yaml</summary>
  
  ```yaml
# -----------------------------------------
# Configuration for `enterprise-logs` chart
# -----------------------------------------

# -- Overrides the chart's name
nameOverride: null

# -- Overrides the chart's computed fullname
fullnameOverride: null

# -- Definition of the Docker image for Grafana Enterprise Logs
# If any of these values are changed, then they also need to be reflected in
# `loki-distributed.loki.image`
image:
  # -- The container registry to use
  registry: docker.io
  # -- The image repository to use
  repository: grafana/enterprise-logs
  # -- The version of Grafana Enterprise Logs
  tag: v1.2.0
  # -- Defines the policy how and when images are pulled
  pullPolicy: IfNotPresent
  # -- Additional image pull secrets
  pullSecrets: []

# -- Definition of the ServiceAccount for containers
# Any additional configuration of the ServiceAccount has to be done in
# `loki-distributed.serviceAccount`.
serviceAccount:
  # -- Specifies whether a ServiceAccount should be created
  # If this value is changed to `false`, it also needs to be reflected in
  # `loki-distributed.serviceAccount.create`.
  create: false

# -- External config.yaml
# A GEL configuration file may be provided as Kubernetes Secret outside of this Helm chart.
useExternalConfig: false
externalConfigName: enterprise-logs-config
externalConfigVersion: "0"

# -- External license.jwt
# A GEL license file may be provided as Kubernetes Secret outside of this Helm chart.
useExternalLicense: false
externalLicenseName: enterprise-logs-license
externalLicenseVersion: "0"

# -- Grafana Enterprise Logs license
# In order to use Grafana Enterprise Logs features, you will need to provide
# the contents of your Grafana Enterprise Logs license, either by providing the
# contents of the license.jwt, or the name Kubernetes Secret that contains your
# license.jwt.
# To set the license contents, use the flag `--set-file 'license.contents=./license.jwt'`
license:
  contents: "NOTAVALIDLICENSE"

# -- Grafana Enterprise Logs configuration file
config:
  auth:
    type: enterprise
  auth_enabled: true

  license:
    path: "/etc/enterprise-logs/license/license.jwt"
[...]

PR #860, opened 2021-11-29
[enterprise-logs] bumped version of loki-distributed chart

vlad-diachenko:
fixes the issue #853 

> in GEL:v1.2.0 , WAL is enabled by default and the default WAL path is wal that makes the app to create a folder wal in working directory (/ect/enterprise-logs/) that is read-only.

the latest version of loki-distributed defines correct path for the WAL.

### DIFF
--- a/charts/enterprise-logs/CHANGELOG.md
+++ b/charts/enterprise-logs/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 1.3.3
+
+- [BUGFIX] Bumped version of `loki-disctributed` chart to 0.39.3 that defines default WAL location. #863
+
 ## 1.3.2
 
 - [BUGFIX] Fixed issue that caused GEL version not to be updated in the templates of the loki-distributed child chart. #863
--- a/charts/enterprise-logs/Chart.lock
+++ b/charts/enterprise-logs/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: loki-distributed
   repository: https://grafana.github.io/helm-charts
-  version: 0.37.3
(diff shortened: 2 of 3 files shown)


charts/enterprise-logs/values.yaml
```diff
@@ -97,6 +97,8 @@ config:
     chunk_block_size: 262144
     chunk_encoding: snappy
     chunk_retain_period: 1m
+    wal:
+      dir: /var/loki/wal
```
chaudum, line 101: Is it a good idea to put the WAL on ephemeral storage?

chaudum: Also, I think this should be fixed in the loki-distributed chart, if it isn't already.

vlad-diachenko: > Is it a good idea to put the WAL on ephemeral storage?

it will be PVC if persistence is enabled.
```yaml
 {{- if not .Values.ingester.persistence.enabled }}
        - name: data
          emptyDir: {}
  {{- else }}
  volumeClaimTemplates:
    - metadata:
        name: data
      spec:
        accessModes:
          - ReadWriteOnce
        {{- with .Values.ingester.persistence.storageClass }}
        storageClassName: {{ if (eq "-" .) }}""{{ else }}{{ . }}{{ end }}
        {{- end }}
        resources:
          requests:
            storage: {{ .Values.ingester.persistence.size | quote }}
  {{- end }}
```

chaudum: :see_no_evil: I've only looked at my local deployment where persistence was not enabled. Should we do that by default, maybe?

vlad-diachenko: updated version of loki-distributed . the latest version already contains this fix

chaudum (approved):

PR #867, opened 2021-11-30
[enterprise-metrics] Remove non-relevant default limits

simonswine:
I think these limits shouldn't be set in this way, as they are likely to
come from copy-pasting the loki helm chart.

- `enforce_metric_name: false`: Is required for Loki, but not really helpful for Metrics.
- `reject_old_samples: true`/`reject_old_samples_max_age: 168h` are not relevant for TSDB storage engine, as the limiting factor is the TSDB storage engine for ingestion older samples, rathern than these defaults.

### DIFF
--- a/charts/enterprise-metrics/CHANGELOG.md
+++ b/charts/enterprise-metrics/CHANGELOG.md
@@ -10,6 +10,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 1.7.1
+
+* [BUGFIX] Remove chunks related default limits. #867
+
 ## 1.7.0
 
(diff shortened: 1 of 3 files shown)


09jvilla:
@simonswine should we also remove these limits from mimir so that they are not even exposed?

simonswine:
> @simonswine should we also remove these limits from mimir so that they are not even exposed?

I think so, they only make sense once we have support for out of order ingestion for TSDB.

09jvilla:
great - I created https://github.com/grafana/mimir/issues/554

jdbaldry (approved):
I think it is likely that these limits come from the Cortex chart from way-back-when.

Looks good to me. I just added a CHANGELOG entry.

PR #869, opened 2021-12-01
[loki] update Ingress API version networking.k8s.io/v1

jdompablo:
`W1201 14:22:28.788947   51565 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress`


MurzNN:
We need to check version and keep `v1beta` for k8s versions older than 1.22, I've implemented this in my PR https://github.com/grafana/helm-charts/pull/879

zanhsieh:
Close in favour of PR #879

PR #871, opened 2021-12-02
[tempo] Add overrides to Tempo

gabrieloandco:
We needed to add overrides to Tempo to change some configurations.

PR #874, opened 2021-12-03
[tempo] Add overrides to Tempo

annanay25:
Follow up of #871

Signed-off-by: Annanay Agarwal <annanay.agarwal@grafana.com>


mapno (approved):

ognjenVlad:
This is not working as expected, overrides are never used in tempo configuration. overrides.yaml should be moved to tempo.yaml or read from /conf/override.yaml

PR #875, opened 2021-12-03
[loki] allow for an empty value in extraArgs

walkafwalka:


trevorwhitney (changes requested):
Would you mind providing a description to help me understand your use case a little bit more? Thanks!

PR #879, opened 2021-12-07
[loki] Update loki ingress template to support kubernetes 1.22+

MurzNN:
When I'm installing loki helm chart to Kubernetes 1.22+, I got the error:
```
Error: UPGRADE FAILED: unable to recognize "": no matches for kind "Ingress" in version "networking.k8s.io/v1beta1"
```
This is because `apiVersion: networking.k8s.io/v1beta1` [becomes deprecated in Kubernetes 1.22+](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#ingress-v122), with replacing to stable `networking.k8s.io/v1`.

Also in `path` part of the template there are some changes needed.

P.S. `grafana` chart installs succesfully on k8s 1.22.4.

_And there are several warnings about compatibility with v1.25+:_
```
W1207 15:08:07.725807 1034065 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
```
_I can fix them in next PR._

### DIFF
--- a/charts/loki/templates/ingress.yaml
+++ b/charts/loki/templates/ingress.yaml
@@ -1,12 +1,7 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "loki.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- $apiVersion := "extensions/v1" -}}
-{{- if semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion -}}
-{{- $apiVersion = "networking.k8s.io/v1" -}}
-{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-{{- $apiVersion = "networking.k8s.io/v1beta1" -}}
-{{- end }}
+{{- $apiVersion := include "ingress.apiVersion" . -}}
 apiVersion: {{ $apiVersion }}
 kind: Ingress
 metadata:
--- a/charts/loki/templates/ingress.yaml
+++ b/charts/loki/templates/ingress.yaml
@@ -4,8 +4,7 @@
 {{- $apiVersion := "extensions/v1" -}}
 {{- if semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion -}}
 {{- $apiVersion = "networking.k8s.io/v1" -}}
-{{- end }}
-{{- if and (semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion) (semverCompare "<1.22-0" .Capabilities.KubeVersion.GitVersion) -}}
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 {{- $apiVersion = "networking.k8s.io/v1beta1" -}}
 {{- end }}
 apiVersion: {{ $apiVersion }}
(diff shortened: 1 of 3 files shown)


jtackaberry:
Tested this patch with microk8s 1.22 and Helm v3.7.2 and it works for me.

Though this perhaps a bit cleaner:

```
{{- if semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion -}}
{{- $apiVersion = "networking.k8s.io/v1" -}}
{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
{{- $apiVersion = "networking.k8s.io/v1beta1" -}}
{{- end }}
```

MurzNN:
@jtackaberry are you sure that 'if else' should work? Cuz I also try to go that way, but found this https://stackoverflow.com/questions/16985469/switch-or-if-elseif-else-inside-golang-html-templates

jtackaberry:
> @jtackaberry are you sure that 'if else' should work? Cuz I also try to go that way, but found this https://stackoverflow.com/questions/16985469/switch-or-if-elseif-else-inside-golang-html-templates

It's working for me with latest Helm.  Comments on that SO post circa 2018 say this is possible.  And there are plenty of occurrences of "else if" in the grafana/helm-charts repo, so I think we're ok.

zanhsieh:
@MurzNN 
Please bump the chart version to latest. Without it, it can't be merged.

MurzNN:
I've reworked to `if else` version and rebased to latest commits on main, @zanhsieh can you please review?

zanhsieh:
@MurzNN 
"Bump the chart version to latest" means updating charts/loki/Chart.yaml version. Plus, you have to sign DCO; see below "Details" next to DCO in this page for instruction.

MurzNN:
Sorry for misunderstanding, now I've bumped `loki` chart from version: 2.8.1 to 2.8.2.

charts/loki/templates/ingress.yaml
```diff
@@ -1,11 +1,13 @@
...
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-apiVersion: networking.k8s.io/v1beta1
-{{- else -}}
-apiVersion: extensions/v1beta1
+{{- $apiVersion := "extensions/v1" -}}
+{{- if semverCompare ">=1.22-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- $apiVersion = "networking.k8s.io/v1" -}}
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- $apiVersion = "networking.k8s.io/v1beta1" -}}
```
zanhsieh, line 8: Could you move this under templates/_helpers.tpl ?

zanhsieh (changes requested):

charts/loki/templates/ingress.yaml
MurzNN: I've moved in last commit, but seems we still need to assign it to variable, to work inside `if` comparisons like this:
```
            {{- if eq $apiVersion "networking.k8s.io/v1" }}
            pathType: Prefix
              {{- end }}
```
Because it doesn't want to work with template values:
```
{{- if eq (template "ingress.apiVersion" .) "networking.k8s.io/v1" }}
```
produces error `unexpected <template> in parenthesized pipeline`, and
```
{{- if eq (include "ingress.apiVersion" .) "networking.k8s.io/v1" }}
```
produces other error: `can't evaluate field Capabilities in type string`.
Maybe you can give me some workaround to this?

zanhsieh (approved):

PR #883, opened 2021-12-08
Add ingress for distributor ring ui

cphu-jt:


Whyeasy:
> This can be achieved by updating the nginx config. Closing this PR for now as it's in complete and it can be achieved in the nginx config as mentioned or port forward to the pod.

cphu-jt:
Aim was to give user opportunities to split exposure using for example two ingress class:
- gateway can be public
- distributor should remain private

PR #884, opened 2021-12-08
Enable wal as default

cphu-jt:


Whyeasy:
This shouldn't be enabled by default and needs to be configured per user basis. As it makes the installation a little bit harder (supplying persistent volume etc. ) for the initial installation. So it's easier for users to try this chart if it's not enabled by default.

Closing this PR.

cphu-jt:
Ok seems fair but not production ready. 

Should i make another pull request to explicitly says that the wal is not enabled by default ?
``        enabled: false
            dir: /var/loki/wal
 ``

Whyeasy:
Hi @cphu-jt I, no this is not needed either. As this is the default value according to the documentation

```
wal:
  # Enables writing to WAL.
  # CLI flag: -ingester.wal-enabled
  [enabled: <boolean> | default = false]
```

https://grafana.com/docs/loki/latest/configuration/#ingester_config

PR #908, opened 2021-12-17
[enterprise-logs] remove fields that make it hard to do helm updates

trevorwhitney:
I'm no helm expert, but I found by setting the selector to explicitly `null`, and using the `checksum/config` annotation, it made it hard to do a helm upgrade of the enterprise logs chart. Since the tokengen is a job, `spec.selector` and `spec.template` are immutable, thus updates always fail when `selector` is trying to be set to `null`, and also fail whenever theres a change to config. I think in the case of the tokenjob it is expected you will have to delete it an recreate it to make changes.

### DIFF
--- a/charts/enterprise-logs/CHANGELOG.md
+++ b/charts/enterprise-logs/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 1.3.4
+
+- [CHANGE] Remove selector and config hash annotations from the tokengen job that make it hard to update the helm chart after deploying that job
+
 ## 1.3.3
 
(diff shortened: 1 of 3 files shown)


zanhsieh:
@trevorwhitney 
Please bump chart to latest version.

chaudum (commented):
LGMT
Please also add a changelog entry and update the version of the chart.

zanhsieh:
@trevorwhitney 
Please sign DCO too. Thank you.

trevorwhitney:
@zanhsieh so, I have signed the DCO. That check seems to fail whenever I don't do `--signoff` with my commits. Why is this necessary? I am signing commits with GPG, and as far as I can tell, the `--signoff` flag only adds the string `Signed-off-by: Trevor Whitney <trevorjwhitney@gmail.com>` to the commit message. Seems like GPG should be sufficient in confirming my identity, yeah?

zanhsieh (approved):

PR #909, opened 2021-12-17
[loki] Fix correct ingress for >= 1.20

tarioch:
The helper returns "networking.k8s.io/v1" and not networking.k8s.io/v1. This fails the if comparison and always generates the else part, which is not valid for the new api.


zanhsieh:
@tarioch 
Please bump chart to latest version. Thank you.

tarioch:
Done, I did it but then forgot to push it

zanhsieh (approved):

Issue #730, opened 2021-10-06
How to solve issue about Promtail extraScrapeConfigs not being picked up?


yusufkaratoprak:
**Describe the bug**
It seems that dropping a namespace **kube-system** and saving a gateway **ambassador** by using the configuration below does not work:
**To Reproduce**
1.Writing configuration 👍  loki-stack-values-v2.4.1.yaml

```
loki:
  enabled: true
  config:

promtail:
  enabled: true
  extraScrapeConfigs:
  - job_name: kubernetes-pods-name
    pipeline_stages:
      - docker: {}
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - source_labels: [__meta_kubernetes_namespace]
        action: drop
        regex: .+
      - action: keep
        regex: ambassador
        source_labels:
          - __meta_kubernetes_namespace
          - __meta_kubernetes_pod_namespace
  - job_name: kubernetes-pods-app
    pipeline_stages:
      - docker: {}
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - source_labels: [__meta_kubernetes_namespace]
        action: drop
        regex: .+
      - action: keep
        regex: ambassador
        source_labels:
          - __meta_kubernetes_namespace
          - __meta_kubernetes_pod_namespace
 
  - job_name: kubernetes-pods-direct-controllers
    pipeline_stages:
      - docker: {}
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - source_labels: [__meta_kubernetes_namespace]
        action: drop
        regex: .+
      - action: keep
        regex: ambassador
        source_labels:
          - __meta_kubernetes_namespace
          - __meta_kubernetes_pod_namespace

  - job_name: kubernetes-pods-indirect-controller
    pipeline_stages:
      - docker: {}
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - source_labels: [__meta_kubernetes_namespace]
        action: drop
        regex: .+
      - action: keep
        regex: ambassador
        source_labels:
          - __meta_kubernetes_namespace
          - __meta_kubernetes_pod_namespace

  - job_name: kubernetes-pods-static
    pipeline_stages:
      - docker: {}
    kubernetes_sd_configs:
      - role: pod
    relabel_configs:
      - source_labels: [__meta_kubernetes_namespace]
        action: drop
        regex: .+
      - action: keep
        regex: ambassador
        source_labels:
          - __meta_kubernetes_namespace
[...]

Issue #208, opened 2021-02-09
loki-stack: Default promtail chart version is far behind the current


openbob:
Even in the latest loki-stack helm chart (version: 2.3.1) "requirements.yaml" file, it has "promtail" chart version set at "^2.2.0" that is quite old.  However, the promtail chart version need to be > 3.0.0 in order to support pod Label "__meta_kubernetes_pod_label_app_kubernetes_io_name" during relabelling.

Can the promtail chart version in the "requirements.yaml" file be updated to use a more current version?

Thanks.

tomasz-kuzemko:
This also breaks the [Syslog Support](https://github.com/grafana/helm-charts/tree/main/charts/promtail#syslog-support) example. I get this error from the Promtail POD logs:

```
Unable to parse config: /etc/promtail/promtail.yaml: yaml: unmarshal errors:
  line 14: field snippets not found in type main.Config
```

After changing the `promtail` dependency version to `3.1.0` it works fine.

jtackaberry:
This threw me for a loop too.  I was about to post a someone's-wrong-on-the-internet comment on #264 but after poking at it realized I was bit by this.

If it's not going to be kept up-to-date, it might be better to remove the loki-stack chart outright?

eljoth:
Is there any statement from the maintainers ?

boniek83:
Disable promtail from loki-stack and use separate and newer promtail chart. Works well.

MikaelElkiaer:
> Disable promtail from loki-stack and use separate and newer promtail chart. Works well.

That's a good workaround.

What's the real fix though? - Would it take more than updating `requirements.txt`?

chefcook:
I would be really interested if there is another way than to install everything separately. Loki stack is already a very good package.

cilindrox:
Chiming in, since we got bit by grafana/loki#3120, which was fixed on `2.2.1` - discovered the hard way the outdated promtail dependency.

An image override was enough `promtail.image.tag: 2.2.1`, but imo the base umbrella chart should include the latest app version at a minimum.

cprivite:
Here's an additional user story. We need to override the registry prom-tail is being pulled from to avoid rate limiting, which we can't do with the old version of the promtail chart.

zanhsieh:
loki-stack updated. See: https://github.com/grafana/helm-charts/pull/1692

Feel free to close this issue if you feel satisfied.

Sheikh-Abubaker:
[...]

PR #925, opened 2021-12-30
[loki-stack] Update Promtail helm-chart dependency

peschmae:
The loki version which is referenced in the loki-stack, is broken and there is no way to properly configure the syslog receiver, since the template is invalid.

Relates to #208 and #730


zanhsieh:
Close in favour of https://github.com/grafana/helm-charts/pull/1692.

PR #933, opened 2022-01-04
[loki-distributed] Rename containers to fix some dashboard panels from loki-mixin

DaveOHenry:
All containers are currently named "loki". This causes problems when one wants to use the loki-mixin dashboards: Some panels (CPU, memory) do not show any data, because the underlying queries use container names like "querier", "ingester" and so on.


unguiculus (dismissed):
I'd suggest you bump the minor version instead and mention this change in the upgrading section in the readme, so folks are aware.

DaveOHenry:
That makes sense. Added a short description in the upgrading section and bumped the minor version. Also fixed a conflict. 

The only problematic case that came to my mind was a custom prometheus rule that uses the container name. Is there another scenario that is worth mentioning?

Whyeasy (approved):
LGTM!

PR #941, opened 2022-01-06
Update README.md

neverping:
The image renderer plugin-in was linking to a broken documentation.

After inspecting the changes on the `grafana-image-render repository`,
I found out that the change below updated most of the documentation,
where the content of the `remote_rendering_using_docker.md` was 
moved to the main `README.md` file.

https://github.com/grafana/grafana-image-renderer/pull/277


neverping:
Oh, I will open this PR with my corporate account, which signs commits.

PR #943, opened 2022-01-06
[enterprise-logs] Convert config value from structured data to a string

trevorwhitney:
This change brings the enterprise-logs chart in line with the loki-distributed chart in terms of how the config is handled. In the loki-distributed chart, the config is a string. This has the benefit of accepting the config exactly as the user provides it, without merging in other default values that are hard to keep up to date as the software changes. It has the downside of the config no longer being structured data, so we can't access nested properties like `server.http_listen_port`.

I considered exposing the http and grpc ports as top level configs, but they are hardcoded in multiple places in the loki-distributed chart, so I'd rather force that convention here as well to reduce the configuration surface area. I would be open to hardcoding `http_listen_port` to 80 as that's what's the default in Loki, but I fear that could create issues since it's a reserved port. Maybe we should change the default in Loki?

### DIFF
--- a/charts/enterprise-logs/Chart.yaml
+++ b/charts/enterprise-logs/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: "v2"
 name: "enterprise-logs"
 type: application
-version: "1.4.0"
+version: "2.0.0"
 appVersion: "v1.3.0"
 kubeVersion: "^1.10.0-0"
 description: "Grafana Enterprise Logs"
--- a/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
+++ b/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
@@ -103,10 +103,10 @@ spec:
             {{- end }}
           ports:
             - name: http-metrics
-              containerPort: {{ .Values.config.server.http_listen_port }}
+              containerPort: 3100
               protocol: TCP
             - name: grpc
-              containerPort: {{ .Values.config.server.grpc_listen_port }}
+              containerPort: 9095
               protocol: TCP
             - name: http-memberlist
               containerPort: 7946
--- a/charts/enterprise-logs/templates/admin-api/service-admin-api.yaml
+++ b/charts/enterprise-logs/templates/admin-api/service-admin-api.yaml
@@ -15,11 +15,11 @@ spec:
   type: ClusterIP
   ports:
     - name: http-metrics
-      port: {{ .Values.config.server.http_listen_port }}
+      port: 3100
       protocol: TCP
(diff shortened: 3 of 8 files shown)


jdbaldry:
I personally had inclination towards the structured configuration when we originally considered this but I'm happy to change path if we can set and document a clear position for the future.

A quick search found a number of issues and PRs that have sought to change between these two options:
- https://github.com/grafana/helm-charts/pull/554
- https://github.com/grafana/helm-charts/issues/433
- https://github.com/grafana/helm-charts/pull/588

If we choose to change we will have to document the breaking change to users and should discourage the merge of #554.

@unguiculus has provided the most comprehensive explanation for preferring string config over structured config in https://github.com/grafana/helm-charts/pull/588#pullrequestreview-728607829. They have much more experience with Helm than I do.

jdbaldry:
With regards to the hard coded ports, if the loki-distributed chart is also doing the same, I'm happy to extend that and reduce surface area as you mentioned.

trevorwhitney:
@jdbaldry thanks for the background, I will get up to speed on previous discussions. In the meantime, l thought it might be useful for me to to document my motivation for going to string config values:

1) With the introduction of the common config in 2.4 or Loki, the structured config presents a more specific issue. The common config valuea are only used if there is not a more specific value present (ie. the common ring is only used for the ingester if there is not an ingester specific ring config.) With the structured default config, these more specific subsections need to be specifically `null`'d out, and in some cases even that was not working correctly for me. I think it is confusing for users who may try to add a simpler config, to then end up with an actual config much larger than what they provided, which may in fact be overriding the values they intended (ie. using a more specific default value rather than the common config value provided by the user).

2) I think structured config requires more effort to keep these charts up to date with new versions. Large configuration changes might require significant re-work of the structured defaults, to make sure the defaults will not interact with user provided config in an unexpected way. I think it is likely this will be missed in the future since it requires us to consider all permutations of combinations of user provided values with defaults to determine what may or may not break.

trevorwhitney:
After reading through the previous discussions, I agree with the points @unguiculus made in #588 as well, though they are different than the issues I mentioned above.

unguiculus:
Thanks, @trevorwhitney. I fully second this change. However, since this is a breaking change, it deserves a major version bump. You should also add an upgrading section to the readme that documents this.

trevorwhitney:
@unguiculus bumped chart to version 2.0

jdbaldry (dismissed):
LGTM!

zanhsieh:
@trevorwhitney 
Please sign DCO. Instructions should be located on the bottom of this PR page. Thank you.

zanhsieh (approved):

Issue #952, opened 2022-01-11
Make the loki-ingester a replicaset


applike-ss:
Hello,

we want to scale the loki ingester in a horizontal way.
Currently the ingester in the distributed helm chart (which we are using) is a statefulset.
I would like to suggest either adjusting it to a replicaset or to let the chart user decide about the resource kind.
If a PR for this would be welcome, then i'd create one.

best,
Sebastian S.

PR #953, opened 2022-01-11
[loki-distributed] added option to use replicaset for ingesters

applike-ss:
This implements the issue stated in #952.
While it works, we are actually aiming for a different solution.

### DIFF
--- a/charts/loki-distributed/templates/ingester/deployment-ingester.yaml
+++ b/charts/loki-distributed/templates/ingester/deployment-ingester.yaml
@@ -1,6 +1,6 @@
-{{- if eq .Values.ingester.kind "StatefulSet"}}
+{{- if eq .Values.ingester.kind "Deployment"}}
 apiVersion: apps/v1
-kind: StatefulSet
+kind: Deployment
 metadata:
   name: {{ include "loki.ingesterFullname" . }}
   labels:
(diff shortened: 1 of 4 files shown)


charts/loki-distributed/templates/ingester/deployment-ingester.yaml
```diff
@@ -0,0 +1,107 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "loki.ingesterFullname" . }}
+  labels:
+    {{- include "loki.ingesterLabels" . | nindent 4 }}
+    app.kubernetes.io/part-of: memberlist
+spec:
+  replicas: {{ .Values.ingester.replicas }}
+  minReadySeconds: 60 # Should we define this variable or is a static value for the ingesters to register in the ring fine?
```
Whyeasy, line 11: Maybe drop this? And replace it with a liveness and readiness probe. If I take a look at the ingester deployment type in the cortex Helm, 'minreadyseconds' is not defined.

applike-ss: there's a readyness probe in place, so i removed the `minReadySeconds`. Didn't check how to do a liveness probe for loki. Since there is none for the statefulset, i assume we don't require it here?

Whyeasy (approved):
LGTM!

Whyeasy:
@applike-ss could you please run helm docs? The rest looks good.

applike-ss:
Done that, thanks for the hint

Whyeasy (approved):
LGTM!

PR #915, opened 2021-12-21
[enterprise-logs] Use correct configuration for the compactor's persistent volumes

chaudum:
Closes #913 

Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

PR #957, opened 2022-01-12
[enterprise-logs] multiple fixes

chaudum:
This PR contains multiple changes and fixes:

- [BUGFIX] Use correct subPath configuration for the compactor's storage mount.
- [BUGFIX] Fixed issue that prevented users from mouting extra persistent volumes for the compactor.
- [CHANGE] Configure `securityContext.fsGroup` value for Admin API pod based on the value `adminApi.securityContext.runAsGroup`.

This PR supersedes #915 

Signed-off-by: Christian Haudum <christian.haudum@gmail.com>

### DIFF
--- a/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
+++ b/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
@@ -41,8 +41,10 @@ spec:
       {{- if .Values.adminApi.priorityClassName }}
       priorityClassName: {{ .Values.adminApi.priorityClassName }}
       {{- end }}
+      {{- if .Values.adminApi.securityContext.runAsGroup }}
       securityContext:
-        fsGroup: 10001
+        fsGroup: {{ .Values.adminApi.securityContext.runAsGroup | int }}
+      {{- end}}
       initContainers:
       # Taken from
       # https://github.com/minio/charts/blob/a5c84bcbad884728bff5c9c23541f936d57a13b3/minio/templates/post-install-create-bucket-job.yaml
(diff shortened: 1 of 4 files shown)


jdbaldry (approved):
LGTM

charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
```diff
@@ -41,8 +41,10 @@ spec:
       {{- if .Values.adminApi.priorityClassName }}
       priorityClassName: {{ .Values.adminApi.priorityClassName }}
       {{- end }}
+      {{- if .Values.adminApi.securityContext.runAsGroup }}
       securityContext:
-        fsGroup: 10001
+        fsGroup: {{ .Values.adminApi.securityContext.runAsGroup | int }}
+      {{- end}}
```
ssncferreira, line 47: In case, `runAsGroup` is not defined, the `fsGroup`, is not written. Shouldn't we default for that case?
```
securityContext:
{{- if .Values.adminApi.securityContext.runAsGroup }}
  fsGroup: {{ .Values.adminApi.securityContext.runAsGroup | int }}
{{- else }}
  runAsGroup: 10001
  fsGroup: 10001
{{- end}}
```

ssncferreira (commented):
LGTM, just a small comment 🙂

charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml
chaudum: If `runAsgroup` is not defined in the values.yaml then also the securityContext on the container will not be set, see https://github.com/grafana/helm-charts/blob/main/charts/enterprise-logs/templates/admin-api/deployment-admin-api.yaml#L118-L120
However, it probably won't work unless you use a modified Loki Docker image.

PR #961, opened 2022-01-12
[loki-simple-scalable] update to Loki 2.4.2

slim-bean:


ssncferreira (approved):
LGTM 🍰

PR #966, opened 2022-01-14
[loki-stack] Allow query limit increase over 1000 lines

gcaracuel:
Query limit is changed via _max_entries_limit_ https://grafana.com/docs/loki/latest/configuration/#limits_config which default value is 5000 but not affecting here when using Dashboard sidecar as the datasource is limited by default to 1000.


zanhsieh:
@gcaracuel 
Please split this PR to one chart per PR. Thank you.

gcaracuel:
Splitting to 2 diff new PRs

PR #974, opened 2022-01-18
[promtail] Add extra (dynamic) Manifests to chart

nxf5025:
Signed-off-by: Nick Fisher <nfisher1@its.jnj.com>

Adding the ability to add extra manifests to the deployment of the chart. This will allow for better integration with secrets add-ons such as https://github.com/external-secrets/kubernetes-external-secrets or https://github.com/kubernetes-sigs/secrets-store-csi-driver

This is essentially the same change as https://github.com/grafana/helm-charts/pull/928, but for promtail.

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.4.2
-version: 3.10.0
+version: 3.11.0
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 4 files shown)


nxf5025:
@unguiculus / @Whyeasy - Any way to get some traction on this? Currently forced to use a subchart for our secretProviderClass which isn't as clean as we'd like.

zanhsieh (approved):

dwerder:
Hello, we have a feature request for some solution like that for the fluxcd helm-chart. Maybe I am a little bit to overcautious, but that basically means that I allow someone to install everything he likes with the chart. From a security point of view that doesn't seem to be such a good idea. How do you guys think about that?

nxf5025:
@dwerder - I don't necessarily see it as a security issue. The proper RBAC and approval workflow should be put in place outside of the chart so that only the users who are supposed to deploy or make changes have the ability to do so.

dwerder:
@nxf5025 Thanks for the feedback

PR #977, opened 2022-01-19
[tempo] authEnabled -> multitenancyEnabled

flokli:
authEnabled created the auth_enabled key in the config file.

This has been renamed a while ago in
https://github.com/grafana/tempo/pull/646, so update the helm chart
accordingly.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.12.2
+version: 0.12.3
 appVersion: 1.2.1
 engine: gotpl
 home: https://grafana.net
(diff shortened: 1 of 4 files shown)


flokli:
Even with `authEnabled` set to `true`, tempo spammed the logs with lines like these:

```
tempo ts=2022-01-19T14:38:20.653574419Z caller=middleware.go:66 msg="failed to extract org id" err="no org id"
```

flokli:
Hmm, it could be I ran into https://github.com/grafana/tempo/issues/495 (as I'm not sending over GRPC). Any pointers?

joe-elliott:
If you set  `auth_enabled` to false then it should allow ingestion/queries without setting the `X-Scope-OrgID` header.

Alternatively try: `multitenancy_enabled` to false

flokli:
Okay, but independent of this, the helm chart should configure `multitenancy_enabled` instead of the deprecated ` auth_enabled`, right?

annanay25 (approved):
This change is fine, but not sure about you getting `no-org-id` when setting `auth_enabled` to false. Unless `multitenancy_enabled` is set to true Tempo will not look for `X-Scope-OrgID` header

PR #979, opened 2022-01-20
[loki-simple-scalable] add hpas for read/write loki simple scalable services

wondersd:
Adding hpa resources for read/write loki-simple-scaling services. read looks like it was intended to have this but wasnt finished (values with no template files) and write didnt have either.

I think this may need a warning about enabling autoscaling on writes when not using an object_store backend.


zanhsieh (approved):

charts/loki-simple-scalable/values.yaml
```diff
@@ -173,6 +173,17 @@ prometheusRule:
 write:
   # -- Number of replicas for the write
   replicas: 1
+  autoscaling:
+    # -- Enable autoscaling for the write
+    enabled: false
+    # -- Minimum autoscaling replicas for the write
+    minReplicas: 1
```
trevorwhitney, line 180: can we make the `minReplicas` 3? One of the goals of simple-scalable is to use Loki defaults as much as possible. I'd like to be able to use the default replication_factor of 3, which would require `minReplicas` to be 3. Furthermore, anyone who has a big enough cluster that they need to enable autoscaling probably also wants a replication factor of 3 so they have failure tolerance.

charts/loki-simple-scalable/values.yaml
```diff
@@ -173,6 +173,17 @@ prometheusRule:
...
+    # -- Maximum autoscaling replicas for the write
+    maxReplicas: 3
```
trevorwhitney, line 182: and we could then bump `maxReplicas` to 6?

trevorwhitney (changes requested):
I have a [PR](https://github.com/grafana/helm-charts/pull/1118) open to max the default replicas 3, in which case I think it would also make sense to bump the `min` and `max` replica values for the autoscaling. The goal here is to make the helm defaults work with loki defaults as much as possible. WDYT?

charts/loki-simple-scalable/values.yaml
zanhsieh: I support minReplicas = 3.

charts/loki-simple-scalable/values.yaml
zanhsieh: I support setting it to maxReplicas = 3 initially so no immediately scaling for testing. The user could later change maxReplicas value once they feel confident / full tested to handle scaling.

zanhsieh (dismissed):

zanhsieh (approved):

trevorwhitney:
closing this in favor of #1188 since we can't get the DCO signed here.

Issue #918, opened 2021-12-22
[grafana] usage of recommended kubernetes labels


yeplaa:
Hi,

Helm lists the recommended kubernetes labels from https://helm.sh/docs/chart_best_practices/labels/ 
Currently two labels are missing in Grafana Chart Helm : 
   - **app.kubernetes.io/component**  : Maybe it will be interesting to add this label with *grafana* value, what do you think? 
   - **app.kubernetes.io/part-of** : my opinion is that it is a label to add by the user
  
What do you think?

PR #994, opened 2022-01-25
[grafana] Add app.kubernetes.io/component label

yeplaa:
Signed-off-by: Loïc Fouray <loic.fouray@orange.com>

Helm lists the recommended kubernetes labels from https://helm.sh/docs/chart_best_practices/labels/

Currently two labels are missing in Grafana Chart Helm :

app.kubernetes.io/component : it will be interesting to add this label
app.kubernetes.io/part-of : my opinion is that it is a label to add by the user

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.21.0
+version: 6.21.1
 appVersion: 8.3.4
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_helpers.tpl
+++ b/charts/grafana/templates/_helpers.tpl
@@ -82,6 +82,7 @@ Selector labels
 {{- define "grafana.selectorLabels" -}}
 app.kubernetes.io/name: {{ include "grafana.name" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/component: grafana
 {{- end -}}
 
 {{/*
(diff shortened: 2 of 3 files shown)


zanhsieh (approved):

billimek:
FYI this change appears to be breaking for existing deployments of Grafana.  The label field seems to be immutable:

```
Helm upgrade failed: cannot patch "grafana" with kind Deployment: Deployment.apps "grafana" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/component":"grafana", "app.kubernetes.io/instance":"grafana", "app.kubernetes.io/name":"grafana"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

johanfleury:
Yeah, this lead to downtime on our side as the helm release was not updated with `--atomic` so the service ended up with a selector that was not matching pods’ labels.

Helm doesn’t handle immutable fields (see https://github.com/helm/helm/issues/7082), those should be handled with extra care.

Also, to be honest, I don’t understand the point of this PR. Grafana is a single component service, why does it need the `app.kubernetes.io/component` label?

yeplaa:
The usage of app.kubernetes.io/component label identifies Grafana and Grafana-renderer objects with the same label. i have created an issue on this subject https://github.com/grafana/helm-charts/issues/918
Indeed it fails on update ; label field immutable. Could you rollback my commit, or should i created a new PR to revert?

zanhsieh:
> The usage of app.kubernetes.io/component label identifies Grafana and Grafana-renderer objects with the same label. i have created an issue on this subject #918 Indeed it fails on update ; label field immutable. It's better to do a rollback? Could you rollback my commit, or should i created a new PR to revert with 6.21.2 tag?

Yeap. That will be easier. As the rest of this chart maintainers are not active.

PR #1019, opened 2022-02-08
[tempo-distributed] Enable security context for pods and containers

warrenhodg:
Enable security context for pods and containers as per the style of
the loki-distributed chart.

Signed-off-by: Warren Hodgkinson <warren.hodgkinson@gmail.com>


zanhsieh:
@warrenhodg 
Could you bump the chart version to latest? Please bump minor version since previously this chart does not have security context.

joe-elliott:
Thanks for the contribution. As @zanhsieh noted it needs a version bump. Also please run helm-docs or the linter won't pass:

```
```

One question:
Will this break upgrades? If we change the use that Tempo is running as I'm concerned the ingesters will not be able to access all the files it created when it comes back up.

zanhsieh:
@warrenhodg 
Could you bump Chart.yaml version to latest please?

joe-elliott:
Does this recently merged PR accomplish what you want?

https://github.com/grafana/helm-charts/pull/1443

Issue #104, opened 2020-12-03
Prepend default port names with "http-"


snuggie12:
This will make istio work out of the box and save anyone some time who is unaware of the rules behind [protocol selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#automatic-protocol-selection).

It will also reduce number of overrides someone needs to provide to the chart (i.e. `.Values.podPortName` and `.Values.service.portName`)

PR #1021, opened 2022-02-09
[tempo] specify grpc appProtocol for tempo service

walkafwalka:
Istio uses port name prefixes or `appProtocol` to determine the protocol.
Currently, the names have the protocol as a suffix, but using appProtocol felt like a smaller change.

Relates to https://github.com/grafana/helm-charts/issues/104.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.13.2
+version: 0.14.0
 appVersion: 1.3.1
 engine: gotpl
 home: https://grafana.net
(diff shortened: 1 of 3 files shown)


joe-elliott:
Thanks for the contribution. Looks like you need to bump the chart version and run helm-docs:

```
```

One question. What versions of k8s supports this field? Will this cut off users on earlier versions?

walkafwalka:
> One question. What versions of k8s supports this field? Will this cut off users on earlier versions?

`appProtocol` was introduced in Kubernetes version 1.18. As an alternative, we could prefix port names with `grpc` instead of as a suffix.

walkafwalka:
> Add the prefix like you're suggesting

I went with the prefix for simplicity.

joe-elliott (approved):
Thanks for the improvement!

kmai:
@joe-elliott this wasn't merged onto `tempo-distributed`, right?

PR #1026, opened 2022-02-09
[grafana] Use `with` for serviceMonitor interval and scrapeTimeout

xbglowx:
I ran into a problem with kube-prometheus-stack setting interval to "", which gets rendered as `null`

Signed-off-by: Brian Glogower <xbglowx@gmail.com>


zanhsieh (approved):

xbglowx:
Fixed via https://github.com/grafana/helm-charts/pull/1030

PR #1031, opened 2022-02-10
[enterprise-logs] Use the same ServiceAccount for tokengen

patrickabrennan:
Adjusted serviceAccount and ServiceAccountName so that it can be provided from the values.yaml file. This is needed for K8s on GKE, OpenShift, etc...


jdbaldry (approved):
LGTM. @patrickabrennan tested this out in GCS and everything was working as expected.

PR #1033, opened 2022-02-10
[enterprise-logs] Re-add WAL dir configuration lost in 2.0.0 update

patrickabrennan:
Need to add the following under the ingester section of `config: |` of the values.yaml file for enterprise-logs chart. Right now this setting is only in the loki-distributed helm chart values.yaml file 

```
wal:
     dir: /var/loki/wal
```

This results in the ingester pods crash looping with an error of readonly filesystem


jdbaldry (approved):
Thanks @patrickabrennan! I really appreciate your contribution.

I have pushed some commits that update the chart version and CHANGELOG as well as other housekeeping. Once this is merged, the 2.0.1 release of this Helm chart will have your fix in :)

PR #1000, opened 2022-01-27
[loki-distributed] add global and per-component pod labels

aslafy-z:
Add global and per-component `podLabels` field to the `loki-distributed` chart.

PR #1043, opened 2022-02-15
[tempo-distributed] tempo podlabels

rlex:
Adds podlabels for tempo-distributed, global and per-component


rlex:
please also take a look at #1000, similar to this but for loki

joe-elliott (approved):
Awesome! Thanks for the improvement.

PR #667, opened 2021-09-06
[loki-distributed] Add autoscaling support

stevehipwell:
Fixes https://github.com/grafana/helm-charts/issues/664.

PR #1047, opened 2022-02-17
[loki-distributed] fix - add support for autoscaling ingester

joshuasimon-taulia:
I believe if you set replication_factor: 3, you can autoscale the ingester up and down. memory load seems to be evenly distributed to all of my replicas with a max of 6. mostly copied from https://github.com/grafana/helm-charts/pull/667

### DIFF
--- a/charts/loki-distributed/templates/ingester/statefulset-ingester.yaml
+++ b/charts/loki-distributed/templates/ingester/statefulset-ingester.yaml
@@ -7,7 +7,9 @@ metadata:
     {{- include "loki.ingesterLabels" . | nindent 4 }}
     app.kubernetes.io/part-of: memberlist
 spec:
+{{- if not .Values.ingester.autoscaling.enabled }}
   replicas: {{ .Values.ingester.replicas }}
+{{- end }}
   podManagementPolicy: Parallel
   updateStrategy:
     rollingUpdate:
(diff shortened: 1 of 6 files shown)


Whyeasy:
@joshuasimon-taulia Just to double check. You run your ingester with the deployment kind? Or with statefulset?

joshuasimon-taulia:
i'm currently running my ingester as a statefulset, but the included vpa logic will work with a deployment.

charts/loki-distributed/templates/ingester/statefulset-ingester.yaml
```diff
@@ -7,7 +7,9 @@ metadata:
     {{- include "loki.ingesterLabels" . | nindent 4 }}
     app.kubernetes.io/part-of: memberlist
 spec:
+{{- if not .Values.ingester.autoscaling.enabled }}
```
trevorwhitney, line 10: for consistency with the querier templates (which can also be run as either a deployment or a stateful set), I would advocate we not make replicas conditional on autoscaling here.

trevorwhitney (dismissed):
Can we remove the condition on # of replicas from the statefulset? This will match the what we have for the querier.

zanhsieh (approved):

charts/loki-distributed/values.yaml
```diff
@@ -278,6 +278,17 @@ ingester:
...
+    # -- Enable autoscaling for the ingester
+    enabled: false
+    # -- Minimum autoscaling replicas for the ingester
+    minReplicas: 1
+    # -- Maximum autoscaling replicas for the ingester
+    maxReplicas: 3
+    # -- Target CPU utilisation percentage for the ingester
+    targetCPUUtilizationPercentage: 60
+    # -- Target memory utilisation percentage for the ingester
+    targetMemoryUtilizationPercentage:
```
zanhsieh, line 291: What is the default value here for memory?

zanhsieh (changes requested):

zanhsieh (approved):

PR #1049, opened 2022-02-17
[loki] feat: Add optional prometheusRule value/template

eyenx:
Adds an additional prometheusRule template for adding custom Rules which can be used by alertmanager to alert us.

It makes sure prometheusRules are only deployed if serviceMonitor is enabled

---

Signed-off-by: Toni Tauro <toni.tauro@adfinis.com>

### DIFF
--- a/charts/loki/Chart.yaml
+++ b/charts/loki/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: loki
-version: 2.10.0
+version: 2.10.1
 appVersion: v2.4.2
 kubeVersion: "^1.10.0-0"
 description: "Loki: like Prometheus, but for logs."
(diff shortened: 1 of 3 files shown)


zanhsieh:
@eyenx 
You need to bump the chart to the latest version, otherwise lint test won't pass.

zanhsieh (approved):

charts/loki/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: loki
-version: 2.9.1
+version: 2.10.0
```
zanhsieh, line 3: 2.10.1 ?

zanhsieh (changes requested):

charts/loki/Chart.yaml
eyenx: @zanhsieh pls rereview

zanhsieh (approved):

PR #1057, opened 2022-02-18
[tempo-distributed] feat: support for autoscaling

joshuasimon-taulia:
mostly copied from https://github.com/grafana/helm-charts/pull/667/files.

In the `loki-distributed` chart, autoscaling the `querier` service is dependent on `.Values.indexGateway` being enabled, but I noticed that there's no `indexGateway` in the `tempo-distrubuted` chart. Is there an equivalent? Does this mean that the tempo querier is not scalable?


joe-elliott:
Thanks for the improvement!

> In the loki-distributed chart, autoscaling the querier service is dependent on .Values.indexGateway being enabled, but I noticed that there's no indexGateway in the tempo-distrubuted chart. Is there an equivalent?

There is no equivalent.

> Does this mean that the tempo querier is not scalable?

Tempo querier is stateless and scaleable.

So I really like this improvement, but I have two concerns:
- Ingesters are stateful and easy to scale up but currently need to be scaled down manually by calling the `/shutdown` endpoint to force all data to the backend. So I think we should remove the auto scaling options for ingesters.
- Compactors are stateless and can be scaled automatically but if there's a lot of churn then compactors will waste effort by doing partial jobs as well as duplicate work by attempting the same jobs.

Can we remove the ingester scaling and maybe add some kind of cooldown or throttling configuration for the compactors? Maybe default to only autoscaling compactors once every 6 hours or something?

joshuasimon-taulia:
> Ingesters are stateful and easy to scale up but currently need to be scaled down manually by calling the /shutdown endpoint to force all data to the backend. So I think we should remove the auto scaling options for ingesters.
anecdotally, using hpa to autoscale both loki-distributed and tempo-distributed ingesters is working for me.

> Compactors are stateless and can be scaled automatically but if there's a lot of churn then compactors will waste effort by doing partial jobs as well as duplicate work by attempting the same jobs.

personally, [we are on GKE which uses the default ](https://cloud.google.com/kubernetes-engine/docs/concepts/horizontalpodautoscaler#preventing_thrashing)`--horizontal-pod-autoscaler-downscale-stabilization` of 5 minutes. before k8s 1.23, this was the only **stable** way to customize hpa autoscaling cooldown behavior. k8s `v1.21.6-gke` seems to support `autoscaling/v2beta2`. I can add hpa configs for the compactors and play with 
```
    behavior:
        scaleDown:
            stabilizationWindowSeconds: 21600
```
on the compactors and report back

secustor:
@joe-elliott 
> ingesters are stateful and easy to scale up but currently need to be scaled down manually by calling the /shutdown endpoint to force all data to the backend. So I think we should remove the auto scaling options for ingesters.

Is there a reason that no  preStop hook is used or is it simply not yet implemented?

joe-elliott:
> Is there a reason that no preStop hook is used or is it simply not yet implemented?

We wouldn't want to run this everytime on shutdown. Generally an ingester flushes data to disk, shuts down, restarts and reloads the data. Force flushing all data to the backend will take a bit longer and requires reviewing logs to confirm it has completed.

joe-elliott:
I appreciate this work, but do not feel comfortable merging it as long as it has an HPA on ingesters. If we remove that I would be glad to merge.

zanhsieh (approved):

PR #1069, opened 2022-02-23
[agent-operator] Topology Spread Constraints, Replicas, Kubelet Log Level Best Practices

gus-reis-dojo:
Topology Spread Constraints, Replicas, Kubelet Log Level Best Practices. Also agent version updated to 0.23

Signed-off-by: Gus Reis <gustavo.reis@dojo.tech>

Signed-off-by: Gus Reis <gustavo.reis@dojo.tech> working for Dojo / Paymentsense

### DIFF
--- a/charts/agent-operator/Chart.yaml
+++ b/charts/agent-operator/Chart.yaml
@@ -2,9 +2,9 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.5
-appVersion: "0.22.0"
+version: 0.1.6
+appVersion: "0.23.0"
 home: https://grafana.com/docs/agent/latest/
-icon: https://raw.githubusercontent.com/grafana/agent/v0.22.0/docs/assets/logo_and_name.png
+icon: https://raw.githubusercontent.com/grafana/agent/v0.23.0/docs/assets/logo_and_name.png
 sources:
-  - https://github.com/grafana/agent/tree/v0.22.0/pkg/operator
+  - https://github.com/grafana/agent/tree/v0.23.0/pkg/operator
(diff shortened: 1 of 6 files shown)


hjet:
hey there @gus-reis-dojo!! thank you for your contribution and for using agent-operator :)

i have a couple of comments:

1. we purposefully omitted the `namespace` parameter as you can set this when installing / updating the chart with the `--namespace` flag. there is [debate/disagreement](https://github.com/helm/helm/issues/5465) on whether this should be included in the chart/template, so we have decided to defer to developers to add this should they need it. can you share your use case here?
2. the operator team (@rfratto and @captncraig) do not think agent operator would support more than one replica at the moment, perhaps you can file an issue in the [agent repo](https://github.com/grafana/agent) with your use case to discuss this?
3. i am not super familiar with `topologySpreadConstraints` but it seems to apply to >1 operator pod use cases, which per `2.` is not supported at the moment. can you share your use case here as well?
4. i am not sure that we want the default logging level to be `warn`, although parametrizing it seems fine to me

cc @rfratto @captncraig @aengusrooneygrafana

aengusrooneygrafana:
Thanks for the review @hjet  - i'll pick this up with @gus-reis-dojo tomorrow!

gus-reis-dojo:
Hi @hjet ! thanks for the feedback/review, hope I get all the answers right:

1. Ok, will pass namespace as -n when applying helm chart if that 100% works, else I'll send you feedback here for sure. My use case is to make sure things are not deployed across different namespaces on K8s clusters.
2 & 3. Are linked, so if we don't have support for more than 1 replica, we won't set topologySpreadConstraints - that's fine.
4. I'm currently having to hack things using Kustomize patches, to make sure log level is set to warn instead of info (it makes it less noisy/avoid flooding logs) if that logging level is parameterised it's one less thing to hack around when using the operator helm chart 👍 

cc @rfratto @captncraig @aengusrooneygrafana

Hit me up if I should send just the parameterised log level as a change, I'll update this PR or create a new one

PR #1070, opened 2022-02-23
[tempo-distributed] Add possibility to enable RBAC and PodSecurityPolicy

trutty:
Add possibility to enable RBAC features and create a PodSecurityPolicy.
Used the [Loki helm chart](https://github.com/grafana/helm-charts/tree/main/charts/loki) as template for the manifests and `values.yaml`.

Yes, [Pod Security Policies](https://kubernetes.io/docs/concepts/policy/pod-security-policy/) are deprecated since Kubernetes v1.21 and will be removed with v1.25, but until then they are in use nonetheless. That's why I think this PR will be a useful addition.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.16.2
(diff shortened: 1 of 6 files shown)


joe-elliott (approved):
Thanks! I appreciate these improvements to the chart.

Agree that pod security policy can still be useful for some folks even as it gets phased out.

PR #1073, opened 2022-02-24
[enterprise-logs] Update values.yaml

patrickabrennan:
Change the values.yaml file, so one can choose pspEnabled for K8s that use PodSecurityPolicy (i.e. AKS, EKS, and GKE) when pspEnbabled set to 'true'. For OpenShift, which doesn't support / use PodSecurityPolicy, but instead uses SecurityContextContraints we can just set pspEnabled to 'false' and sccEnabled to 'true' and have the role, rolebinding and serviceAccount use the proper one (i.e. PSP or SCC).

### DIFF
--- a/charts/enterprise-logs/values.yaml
+++ b/charts/enterprise-logs/values.yaml
@@ -466,8 +466,9 @@ loki-distributed:
 
   # RBAC configuration
   rbac:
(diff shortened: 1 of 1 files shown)


zanhsieh:
@patrickabrennan 
Please sign DCO and bump the chart version.

PR #1074, opened 2022-02-24
[loki-distributed] Update serviceaccount.yaml

patrickabrennan:
Add if conditions to either create a PodSecurityPolicy (used by AKS, EKS, GKE, etc...) or SecurityContextContraints which is what OpenShift uses. This also means we do not need the role.yaml and rolebinding.yaml files


patrickabrennan:
Found a better way to do this. Closing and will open a new PR

PR #1080, opened 2022-02-26
Add files via upload

patrickabrennan:
This adds support for OpenShift by defining the SCC policies required. This goes file securitycontextconstraints.yaml (OCP SCC version of podsecuritypolicy along with the values.yaml change for rbac.sccEnabled  and the role.yaml changes as well as rolebinding.yaml


zanhsieh:
@patrickabrennan 
Which chart?

patrickabrennan:
This is for Loki-distributed chart. This is the OpenShift SCC settings

PR #1085, opened 2022-02-28
Add open shift support to gel

patrickabrennan:
Necessary file for GEL to work on OCP


charts/loki-distributed/templates/role.yaml
```diff
@@ -1,10 +1,11 @@
-{{- if .Values.rbac.pspEnabled }}
+{{- if or .Values.rbac.pspEnabled .Values.rbac.sccEnabledi true }}
```
zanhsieh, line 1: Please fix typo and be sure the boolean logic correctness. Or true with anything will never evaluate as false.

zanhsieh (changes requested):

charts/loki-distributed/templates/role.yaml
patrickabrennan: fixed. Thanks for catching.

vlad-diachenko (approved):
I have tested the changes on OpenShift and it was deployed successfully and GEL worked correctly.

zanhsieh (approved):

PR #1088, opened 2022-03-01
[tempo-distributed] compactor is installed by default

kmcrawford:
There is no `compactor.enabled` flag


annanay25:
Hi @kmcrawford, thanks for the contribution.

Please bump the chart version and update docs. You can use the following command to do so
```
```

zanhsieh (approved):

PR #1089, opened 2022-03-02
[loki-distributed] loki.rulerRulesDirName should be lower case

tpokki:
ConfigMap names cannot be upper case. So for example `Example-Tenant` does not work in ruler configuration:

```
ruler:
  directories:
    Example-Tenant:
       rules1.txt: |
          ...
```
... as the name is used in the ConfigMap name. 

This change will impact only the name of the configmap (e.g. `loki-ruler-rules-example-tenant`), the directory in filesystem be written as it should (e.g. `/etc/loki/rules/Example-Tenant`).

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.2
-version: 0.45.0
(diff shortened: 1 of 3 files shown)


zanhsieh (approved):

tpokki:
Hello @unguiculus , @Whyeasy, 
Could you approve/merge this PR, or is there some open questions or issues to resolve still? Thanks!

PR #1101, opened 2022-03-10
[enterprise-metrics] GEM Files modified for OCP support plus the securitycontextconstraints.ya…

patrickabrennan:
These are modification to the `role.yaml`, `rolebinding.yaml `and `values.yaml `file for OpenShift Support. This also adds a file called `securitycontextconstraints.yaml` which is the equivalent of the `podsecuritypolicy.yaml `file, but for OolpenShift.


krajorama (changes requested):
We'll have to redo this on mimir-distributed as we've deprecated this chart in favour of having a common Mimir/GEM chart.

PR #1104, opened 2022-03-16
[tempo-distributed] Add annotation for query-frontend-discovery

lazam:
The default behavior of a headless service is to get the IP addresses of the pods backing the service. This behavior causes issue when adding annotations such as **external-dns** to expose the `query-frontend-service` . 

Adding a separate annotations for `query-frontend-discovery` will solve this issue.

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -313,7 +313,9 @@ queryFrontend:
     annotations: {}
     # -- Type of service for the queryFrontend
     type: ClusterIP
-  # -- The name of the PriorityClass for query-frontend pods
+  serviceDiscovery:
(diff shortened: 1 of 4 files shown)


charts/tempo-distributed/values.yaml
```diff
@@ -313,7 +313,9 @@ queryFrontend:
     annotations: {}
     # -- Type of service for the queryFrontend
     type: ClusterIP
-  # -- The name of the PriorityClass for query-frontend pods
```
annanay25, line 316: Let's re-add this comment about `priorityClassName`

annanay25 (commented):
Thanks for the contribution!

Docs can be fixed by running:
```
```

annanay25 (approved):
lgtm, thanks!

Issue #849, opened 2021-11-25
Make ruler workload optionally a StatefulSet


stribb:
Per https://grafana.com/docs/loki/latest/operations/recording-rules/#kubernetes it is recommended that the ruler workload should be in a StatefulSet rather than a Deployment.

I propose that we do this conditional on the value of `ruler.persistence.enabled`.

Let me know your thoughts. If we have broad agreement, I can craft a PR.

paalkr:
Please let me know if you would like any help testing such a feature. I'm more then happy to contribute.
Currently I have a fully functional loki 2.4.2 helm microservices deployment running in a test cluster.

In production we still use Kustomize to deploy both Cortex and Loki, and the capability to horizontally scale the ruler is one of the main reason it's not migrate to helm yet.

PR #1109, opened 2022-03-17
[loki-distributed] Allow to configure ruler as statefulset

RomainBelorgey:
Useful if using recording rules and need to scale ruler as recommended by https://grafana.com/docs/loki/latest/operations/recording-rules/#kubernetes

Inspired by the configuration of `ingester`.

Should fix https://github.com/grafana/helm-charts/issues/849

Let me know if I miss something :)

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.2
-version: 0.45.1
+version: 0.45.2
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.2
-version: 0.45.2
+version: 0.46.0
 home: https://grafana.github.io/helm-charts
 sources:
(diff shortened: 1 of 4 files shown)


RomainBelorgey:
Make changes to normally fix the CI (chart version bump + `helm-docs`)

charts/loki-distributed/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.2
-version: 0.45.1
+version: 0.45.2
```
Whyeasy, line 6: please use 0.46.0 for this release.

charts/loki-distributed/values.yaml
```diff
@@ -881,6 +881,8 @@ compactor:
 ruler:
   # -- Specifies whether the ruler should be enabled
   enabled: false
+  # -- Kind of deployment [StatefulSet/Deployment]
+  kind: StatefulSet
```
Whyeasy, line 885: I think we should keep this as a Deployment by default. So that the current deployments of the Ruler won't change for other users.

RomainBelorgey: Makes sense :+1:

zanhsieh:
@RomainBelorgey 
"Documentation not up to date. Please run helm-docs and commit changes!"

RomainBelorgey:
> @RomainBelorgey "Documentation not up to date. Please run helm-docs and commit changes!"

Ah yes because I changed to `Deployment` :sweat_smile: 
Will rerun that

RomainBelorgey:
Sorry, found some issues I missed regarding this change :facepalm: 
Will make a quick PR for that

PR #1115, opened 2022-03-22
[loki-distributed] Fixing statefulset configuration for Ruler

RomainBelorgey:
Forgot some changes, it should fix the previous PR https://github.com/grafana/helm-charts/pull/1109


Whyeasy (approved):
Thanks for the quick fix.

PR #1117, opened 2022-03-23
[loki-distributed] Bump to 0.46.2 to redeploy chart

RomainBelorgey:
As 0.46.1 is not available and the github action is not redeploying it, bumping
the version to 0.46.2 to fix it.


Whyeasy (approved):
LGTM

PR #1118, opened 2022-03-23
[loki-simple-scalable] make ssd default to 3 replicas

trevorwhitney:
Change the defaults for SSD chart to default to 3 replicas for read and write.

### DIFF
--- a/charts/loki-simple-scalable/values.yaml
+++ b/charts/loki-simple-scalable/values.yaml
@@ -187,7 +187,18 @@ prometheusRule:
 # Configuration for the write
 write:
   # -- Number of replicas for the write
-  replicas: 1
+  replicas: 3
+  autoscaling:
+    # -- Enable autoscaling for the read, this is only used if `queryIndex.enabled: true`
+    enabled: false
+    # -- Minimum autoscaling replicas for the read
+    minReplicas: 1
+    # -- Maximum autoscaling replicas for the read
+    maxReplicas: 3
+    # -- Target CPU utilisation percentage for the read
+    targetCPUUtilizationPercentage: 60
+    # -- Target memory utilisation percentage for the read
+    targetMemoryUtilizationPercentage:
   image:
     # -- The Docker registry for the write image. Overrides `loki.image.registry`
     registry: null
@@ -250,7 +261,7 @@ write:
 # Configuration for the read node(s)
 read:
   # -- Number of replicas for the read
-  replicas: 1
+  replicas: 3
   autoscaling:
     # -- Enable autoscaling for the read, this is only used if `queryIndex.enabled: true`
     enabled: false
--- a/charts/loki-simple-scalable/values.yaml
+++ b/charts/loki-simple-scalable/values.yaml
@@ -188,17 +188,6 @@ prometheusRule:
 write:
   # -- Number of replicas for the write
   replicas: 3
-  autoscaling:
-    # -- Enable autoscaling for the read, this is only used if `queryIndex.enabled: true`
-    enabled: false
-    # -- Minimum autoscaling replicas for the read
-    minReplicas: 1
-    # -- Maximum autoscaling replicas for the read
-    maxReplicas: 3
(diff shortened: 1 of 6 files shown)


charts/loki-simple-scalable/values.yaml
```diff
@@ -172,7 +172,7 @@ prometheusRule:
 # Configuration for the write
 write:
   # -- Number of replicas for the write
-  replicas: 1
+  replicas: 3
```
zanhsieh, line 175: Could you explain why autoscaling does not fulfil most common use? Or just your personal local environment require so?

charts/loki-simple-scalable/values.yaml
```diff
@@ -235,18 +235,7 @@ write:
...
-    enabled: false
-    # -- Minimum autoscaling replicas for the read
-    minReplicas: 1
-    # -- Maximum autoscaling replicas for the read
-    maxReplicas: 3
-    # -- Target CPU utilisation percentage for the read
-    targetCPUUtilizationPercentage: 60
-    # -- Target memory utilisation percentage for the read
-    targetMemoryUtilizationPercentage:
+  replicas: 3
```
zanhsieh, line 238: Same as above.

zanhsieh:
@trevorwhitney 
Also please sign DCO and bump the chart version to latest. Without these, this PR cannot be merged.

charts/loki-simple-scalable/values.yaml
zanhsieh: I think it's better to just comment out the autoscaling value, otherwise how does the new users know where to go when they need those syntax? e.g. min 3 and max 7 in some big / wide cluster.

trevorwhitney: @zanhsieh I'm confused by this comment. I don't see anywhere in the chart that the auto-scaling value is being used? I don't think the `write` pod ever had this section, and I think that section for `read` was just copied over by accident when deriving this chart from `loki-distributed`. I think the chart would need a `HorizontalPodAutoscaler` in order to enable auto-scaling. Instead of adding that, it seems a lot simpler to me to just bump the replicas from 1 to 3 so it works with the default replication factor of 3. WDYT?

zanhsieh: @trevorwhitney 
Could your own case represent 75% of the existing loki-simple-scalable user cases? Could you prove it? 
If the syntax for autoscaling lost and one day the existing user needs those, what should they do? Go back to this commit history and dig it up what should it be? Are you sure this public repo would keep all pieces of history?

trevorwhitney: @zanhsieh sorry for the confusion, but I don't understand your question? The simple scalable helm chart has never used autoscaling, so I'm not sure what history you're referring to? Removing this config will work with 100% of `loki-simple-scalable` deployments that already exist. If someone needs to find this syntax, they can look at the `loki-distributed` chart which includes an autoscaling configuration as well as templates for `HorizontalPodAutoscaler`.

As for the replica change, the driving force behind SSD mode is simplicity, which is the motivation for not including autoscaling and for including defaults that work with Loki defaults as much as possible. If someone would prefer using 1 replica, those values are easy to change, and they can look at the included `ci` configurations for examples on how to do that.

trevorwhitney:
@zanhsieh I know see there is an open [PR](https://github.com/grafana/helm-charts/pull/979) to add autoscaling to `loki-simple-scalable`. If there is a need for this in the community, then I'm not opposed to it, though I do think we should make the min 3, and we should still remove that config in this PR and add it back in #979 so it's all in the same PR.

trevorwhitney:
@zanhsieh sounds good, so are you suggesting we close this in favor of #979, and add the min and max of 3 to that PR?

zanhsieh:
> @zanhsieh sounds good, so are you suggesting we close this in favor of #979, and add the min and max of 3 to that PR?

@trevorwhitney 
Yes.

PR #1119, opened 2022-03-23
[promtail] make extra client configs an array of configs

trevorwhitney:
This makes it easier to support multiple client configs in the promtail helm chart. The provided lokiAddress is used for the first client config, but I've changed is from `client:` with a single config to `clients:` to support multiple, with the rest being optionally provided as an array of client configs in the `extraClientConfigs` section.

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.4.2
-version: 3.11.0
+version: 4.0.0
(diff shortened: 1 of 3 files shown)


zanhsieh:
@trevorwhitney 
Please bump the chart version to latest.

zanhsieh (approved):

aeimer:
@trevorwhitney is there a reason why the templating was removed? We used templating for that config.

aeimer:
@trevorwhitney How am I supposed to add eg this config to the default fixed array entry?

image

Your change does not allows us to edit the default client entry as before, was this intentionally?

PR #1152, opened 2022-03-30
[mimir-distributed] Fix the naming of minio configmap and secret in the parent chart.

krajorama:


jdbaldry (approved):
LGTM, assuming CI passes.

PR #1159, opened 2022-03-31
[loki-distributed] Fix metric duplication

### LINKED ISSUE #597: [loki-distributed] ServiceMonitor causes metrics duplication
Because ingester and querier have both a regular and a headless service, the metrics are scraped twice for each pod.
This causes the same metrics to be stored twice with the only difference being the `-headless` suffix in service label.

I assume this is not indended as it seems both wasteful and confusing because when e.g. summing data you might count the data from these components twice.

I think `PodMonitor` might be a better fit for this scenario

stevehipwell:
This PR fixes the metric duplication caused by having an additional headless service for some of the Loki components by adding a label to the headless service to disable the `ServiceMonitor` using it to discover targets.

Closes #597.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.4.2
-version: 0.47.0
+version: 0.47.1
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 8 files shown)


trevorwhitney (approved):
The headless services are only used for memberlist, so ignoring them for metrics seems reasonable.

stevehipwell:
@trevorwhitney the service is used by the ServiceMonitor to capture the endpoints to be translated into a Prometheus scrape config so it doesn't really matter which one is used (the service name is labelled so I guess it matters a bit).

PR #1163, opened 2022-04-01
enterprise-metrics 2.0.0

krajorama:
This is work in progress, do not pick


charts/enterprise-metrics/templates/_helpers.tpl
```diff
@@ -59,3 +59,10 @@ Create the app name of enterprise-metrics clients. Defaults to the same logic as
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Calculate the config from structured and unstructred text input
```
jdbaldry, line 64:
```suggestion
Calculate the config from structured and unstructured text input
```

charts/enterprise-metrics/values.yaml
```diff
@@ -6,9 +6,9 @@
 
 # Container image settings.
 # Since the image is unique for all microservices, so are image settings.
-image:
+image: &image
   repository: grafana/metrics-enterprise
-  tag: v1.7.0
+  tag: r176-00920039
```
jdbaldry, line 11: I'm sure you're away, but just flagging to make sure that we update this on release.

charts/enterprise-metrics/values.yaml
```diff
@@ -42,132 +52,175 @@ tokengenJob:
...
+          url: http://{{ template "enterprise-metrics.fullname" . }}-admin-api.{{ .Release.Namespace }}.svc:{{ include "mimir.serverHttpListenPort" . }}            
+        admin_api:
+          url: http://{{ template "enterprise-metrics.fullname" . }}-admin-api.{{ .Release.Namespace }}.svc:{{ include "mimir.serverHttpListenPort" . }}
+        alertmanager:
+          url: http://{{ template "enterprise-metrics.fullname" . }}-alertmanager.{{ .Release.Namespace }}.svc:{{ include "mimir.serverHttpListenPort" . }}
+        compactor:
+          url: http://{{ template "enterprise-metrics.fullname" . }}-compactor.{{ .Release.Namespace }}.svc:{{ include "mimir.serverHttpListenPort" . }}
+        distributor:
+          url: http://{{ template "enterprise-metrics.fullname" . }}-distributor.{{ .Release.Namespace }}.svc:{{ include "mimir.serverHttpListenPort" . }}
+        graphite_querier:
```
jdbaldry, line 94: I wonder if we should remove the graphite proxy rules since we don't yet deploy the resources for these proxies.

charts/enterprise-metrics/values.yaml
```diff
@@ -42,132 +52,175 @@ tokengenJob:
...
+      file: /var/mimir/runtime.yaml
+
+    memberlist:
+      abort_if_cluster_join_fails: false
+      bind_port: 7946
+      compression_enabled: false
+      join_members:
+      - {{ include "mimir.fullname" . }}-gossip-ring
+
+    # This configures how the store-gateway synchronizes blocks stored in the bucket. It uses Minio by default for getting started (configured via flags) but this should be changed for production deployments.
```
jdbaldry, line 172:
```suggestion
    # This configures how the store-gateway synchronizes blocks stored in the bucket. It uses MinIO by default for getting started but this should be changed for production deployments.
```

charts/enterprise-metrics/values.yaml
```diff
@@ -42,132 +52,175 @@ tokengenJob:
...
+        bucket_name: enterprise-metrics-ruler
+        access_key_id: {{ .Values.mimir.minio.accessKey }}
+        secret_access_key: {{ .Values.mimir.minio.secretKey }}
+        insecure: true
+    {{- end }}
+
+    instrumentation:
+      enabled: true
+      distributor_client:
+        address: 'dns:///{{ template "mimir.fullname" . }}-distributor.{{ .Release.Namespace }}.svc:{{ include "mimir.serverHttpListenPort" . }}'
```
jdbaldry: I believe this should be the GRPC port.

jdbaldry: Oh you fixed this, whilst I was reviewing :)

charts/enterprise-metrics/values.yaml
krajorama: yes, just got RC

charts/enterprise-metrics/values.yaml
krajorama: ack

charts/enterprise-metrics/templates/_helpers.tpl
krajorama: we've removed structured anyway, rewriting this

Logiraptor:
@krajorama Should this PR be closed now after #1203 ?

PR #1167, opened 2022-04-01
[grafana] tpl support

chlumper:
# Description of the change

Add support for template directives at initContainers and envFrom.

# Benefits

Allows to use values for generation of e.g. secret names which is especially handy if chart is used within umbrella chart.

sample `values.yaml`:
```yaml
extraInitContainers:
  - name: create-postgresql-db
    image: bitnami/postgresql:latest
    command: ["createdb"]
    args: [ "mydb" ]
    env:
      - name: PGPASSWORD
        valueFrom:
          secretKeyRef:
            name: '{{ include "grafana.fullname" . }}-env-secret'
            key: GF_DATABASE_PASSWORD

envValueFrom:
  ENV_NAME:
    configMapKeyRef:
      name: '{{ include "grafana.fullname" . }}-env'
      key: value_key

envFromConfigMaps:
  - name: '{{ include "grafana.fullname" . }}-env'
    optional: true
```

Signed-off-by: Christian Lumper <c.lumper@tributech.io>

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.24.1
+version: 6.25.1
 appVersion: 8.4.2
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.25.1
+version: 6.25.0
 appVersion: 8.4.2
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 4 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.24.1
+version: 6.25.1
```
Xtigyro, line 3: Thanks for the contribution, @chlumper! 🥇 

Could you please fix the version so that it is equal to `6.25.0`?

Xtigyro (changes requested):

charts/grafana/Chart.yaml
chlumper: I changed the version to `6.25.0` and I also added some missing doc for the values in the readme.

Xtigyro (approved):

PR #1174, opened 2022-04-04
WIP: test rendering mimir templates in GEM context

krajorama:
When rendering kube manifests in the Mimir subchart context, it is not
possible to generate the correct checksum for the checksum/config
annotation. This is a regression since GEM used to restart services when
configuration was changed. The issue is that in the subchart context
the configuration is in the parent chart and cannot be reached.

This patch makes the manifest templates global named templates and thus
able to be rendered in the parent context.


krajorama:
obsolete, see discussion in #1182

PR #1175, opened 2022-04-04
[enterprise-logs] Ensure each container name matches the component name

jdbaldry:
This is useful to ensure monitoring mixins work.

### DIFF
--- a/charts/enterprise-logs/Chart.yaml
+++ b/charts/enterprise-logs/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: "v2"
 name: "enterprise-logs"
 type: application
-version: "2.0.4"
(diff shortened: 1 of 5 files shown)


charts/enterprise-logs/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "v2"
 name: "enterprise-logs"
 type: application
-version: "2.0.4"
+version: "2.0.5"
```
zanhsieh, line 4: Should bump to minor version, e.g. 2.1.0, since not really backward compatible.

zanhsieh (changes requested):

zanhsieh (approved):

PR #1181, opened 2022-04-05
[mimir-distributed] add standard prometheus pod annotations

jmcarp:
By convention, prometheus scrape configs often use the
`prometheus.io/scrape` and `prometheus.io/port` pod annotations to
configure kubernetes pod scrape targets. Similar to prometheus pod
annotations for other charts in this repo, this patch adds these
annotations to the relevant pods in the mimir chart.

Based on https://github.com/cortexproject/cortex-helm-chart/pull/288.

Signed-off-by: Josh Carp <jm.carp@gmail.com>

### DIFF
--- a/charts/mimir-distributed/CHANGELOG.md
+++ b/charts/mimir-distributed/CHANGELOG.md
@@ -11,6 +11,9 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 2.0.6
+* [ENHANCEMENT] Add standard prometheus pod annotations. #1181
+
 ## 2.0.5
 
 * [BUGFIX] Use new component name system for gateway ingress. This regression has been introduced with #1203. #1260
--- a/charts/mimir-distributed/Chart.yaml
+++ b/charts/mimir-distributed/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 2.0.5
+version: 2.0.6
 appVersion: 2.0.0
 description: "Grafana Mimir"
 engine: gotpl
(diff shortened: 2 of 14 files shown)


krajorama (changes requested):
Hi, thank you for the contribution. I'd kindly ask you to create a template called "mimir.podAnnotations" in _helpers.tpl and have common annotations and these there. And then include the new template in the Deployment/StatefulSet definitions Or you can leave it to me and I'll open an issue. The reason being that values.yaml is meant for user settings and if the user sets podAnnotations themselves, they can easily overwrite these
annotations by accident.

jmcarp:
Updated. Is this roughly what you had in mind?

krajorama:
Yes, exactly, I just need to double check if this interferes with the "ServiceMonitor" setup I have going (and have a PR on), might have to add an "if"

krajorama (approved):
LGTM, please bump the version (0.1.5 at the moment) and run helm-docs (1.4.0) to update the readme, or just follow what the CI says. Let me know if you have trouble.

ryan-dyer-sp:
Shouldnt these be hidden behind a values variable?  This is just adding them to all pods and not all users need or want these annotations.  kube-prometheus-stack deprecated the use of these annotations in favor of servicemonitors a while back.

jmcarp:
@krajorama any preference on this? I'm happy to resolve conflicts or refactor a bit.

krajorama:
Since it's deprecated by prom, I wouldn't add it with it's own named value. Seems like another use case for #1284 on the other hand. @jmcarp so something like `global.podAnnotations` would be acceptable? Note that right now these wouldn't apply to subcharts (memcached and minio). Although it came up in another issue to change to memcached managed by our chart directly (#1315).

zanhsieh:
Mimir-distrubuted repo moved. See:
https://github.com/grafana/helm-charts/blob/main/charts/mimir-distributed/README.md
Close this PR.

PR #1182, opened 2022-04-06
[mimir-distributed] Allow rendering mimir templates in other contexts

krajorama:
When `mimir-distributed` is used as a subchart, the rendering of the
annotation "checksum/config" on Pods is wrong. This is important to
trigger service restart when the config is changed.

There are two options:

a) config is in `mimir-distributed` or shared with `mimir-distributed` in
which case it can contain templates and values of the parent chart, making
it impossible to evaluate in the subchart.

b) config lives in the parent chart, but then it is not possible to reach
it in `mimir-distributed` as subchart and the checksum is the checksum of
an empty string. See: "few important details" in
[Helm subcharts and globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/)

This patch makes the manifest templates global named templates and thus
able to be rendered in the parent context. It is one step short of
merging the mimir-distributed chart with enterprise-metrics.


charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
-# Notes on testing mimir-distributed
+# Developer notes
+
+## Notes on the kube manifest templates and configuration
+
+Why do we generate Kubernetes resource definitions from named templates, and not directly ? That is, why do we have
```
jdbaldry, line 5: nit: Typically there is not a space before sentence ending punctuation.
```suggestion
Why do we generate Kubernetes resource definitions from named templates, and not directly? That is, why do we have
```

charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
...
+```yaml
+{{- if not .Values.skipResourceRendering -}}
+{{- include "mimir.querierDeployment" . -}}
+{{- end -}}
+```
+instead of just writing the Deployment template directly ?
```
jdbaldry, line 11: nit: Typically there is not a space before sentence ending punctuation.
```suggestion
instead of just writing the Deployment template directly?
```

charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
...
+
+[PR #1182](https://github.com/grafana/helm-charts/pull/1182) commit message, quote:
+
+When Mimir is used as a subchart, the rendering of the annotation "checksum/config" on PODs is wrong. This is important to trigger service restart when the config is changed.
```
jdbaldry, line 15: nit: Chart is `mimir-distributed` and the Kubernetes resource is a Pod.
```suggestion
When `mimir-distributed` is used as a subchart, the rendering of the annotation "checksum/config" on Pods is wrong. This is important to trigger service restart when the config is changed.
```

charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
...
+
+There are two options: config is in Mimir or shared with Mimir in which case it can contain templates and values of the parent chart, making it impossible to evaluate in the subchart. Or the config lives in the parent chart, but then it is not possible to reach it in Mimir as subchart and the checksum is checksum of empty string. See: "few important details" in [Helm subcharts and globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/)
+
+This patch makes the manifest templates global named templates and thus able to be rendered in the parent context. It is one step short of merging the mimir-distributed chart with enterprise-metrics.
```
jdbaldry, line 19:
```suggestion
```
I would consider removing this line as I think "This patch ..." will be confusing to readers post-merge.

charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
...
+{{- include "mimir.querierDeployment" . -}}
+{{- end -}}
+```
+instead of just writing the Deployment template directly ?
+
+[PR #1182](https://github.com/grafana/helm-charts/pull/1182) commit message, quote:
+
+When Mimir is used as a subchart, the rendering of the annotation "checksum/config" on PODs is wrong. This is important to trigger service restart when the config is changed.
+
+There are two options: config is in Mimir or shared with Mimir in which case it can contain templates and values of the parent chart, making it impossible to evaluate in the subchart. Or the config lives in the parent chart, but then it is not possible to reach it in Mimir as subchart and the checksum is checksum of empty string. See: "few important details" in [Helm subcharts and globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/)
```
jdbaldry, line 17:
```suggestion
There are two options:
- config is in `mimir-distributed` or shared with `mimir-distributed` in which case it can contain templates and values of the parent chart, making it impossible to evaluate in the subchart.
- config lives in the parent chart, but then it is not possible to reach it in `mimir-distributed` as subchart and the checksum is checksum of empty string. See: "few important details" in [Helm subcharts and globals](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/)
```

charts/mimir-distributed/INTERNAL.md
jdbaldry: Ah, I actually missed that this is quoting the PR description. Perhaps we can use the markdown `>` to indicate that this is a quote instead of the start and end quote sentences?

charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
...
+
+This patch makes the manifest templates global named templates and thus able to be rendered in the parent context. It is one step short of merging the mimir-distributed chart with enterprise-metrics.
+
+End quote.
+
+### Background
+
+Why do we need to have `checksum/config` ? In order to keep this mimir-distributed chart up to date and tested, it is important that it can be reused in the enterprise-metrics chart since Grafana uses Grafana Enterprise Metrics (GEM) internally and for its cloud offering. This means that the chart must support all required features of future and past GEM chart releases. One such feature is that in case of configuration change the affected services must be restarted when `helm upgrade` is executed. The [best practice](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) to achieve that is to use the `checksum/config` annotation.
```
jdbaldry, line 25: nit: Punctuation.
```suggestion
Why do we need to have `checksum/config`? In order to keep this mimir-distributed chart up to date and tested, it is important that it can be reused in the enterprise-metrics chart since Grafana uses Grafana Enterprise Metrics (GEM) internally and for its cloud offering. This means that the chart must support all required features of future and past GEM chart releases. One such feature is that in case of configuration change the affected services must be restarted when `helm upgrade` is executed. The [best practice](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) to achieve that is to use the `checksum/config` annotation.
```

charts/mimir-distributed/INTERNAL.md
```diff
@@ -1,8 +1,36 @@
...
+
+Why not just merge the Mimir and GEM charts ? Having to use an enterprise-metrics chart for the open source Grafana Mimir would sow confusion about it's licensing status and how it can be used and what features are available. On the other hand, having to use mimir-distributed for the enterprise version would potentially slow the development of the open source version as enterprise has stricter rules - not to mention confusion again.
```
jdbaldry, line 27: nit: punctuation, some styling preferences, and grammar fix for "its" as a possessive ("it's" is only a contraction for "it is")
```suggestion
Why not just merge the Mimir and GEM charts? Having to use an `enterprise-metrics` chart for the open source Grafana Mimir would sow confusion about its licensing status and how it can be used and what features are available. On the other hand, having to use `mimir-distributed` for the enterprise version would potentially slow the development of the open source version as enterprise has stricter rules - not to mention confusion again.
```

jdbaldry (commented):
Is it only by convention that we have all the named templates in the `_helpers.*.tpl` files? Perhaps it makes sense for each named template to be in the template file? For example, having the compactor PDB named template defined and used in the `compactor.-pdb.yaml`. 

As I wrote this I recalled that there was different Helm behavior for YAML and TPL files, perhaps this is why we have things as they are?

jdbaldry:
I really appreciate the included explanation in `INTERNAL.md`. I have avoided doing a full docs review of the content but I have included some nits on some things that jumped out as I read through. It was comprehensive enough for me to understand the reasons for this change.

I approve of the direction of this PR as I cannot currently think of a better way of handling this parent and subchart relationship.

krajorama:
> Is it only by convention that we have all the named templates in the `_helpers.*.tpl` files? Perhaps it makes sense for each named template to be in the template file? For example, having the compactor PDB named template defined and used in the `compactor.-pdb.yaml`.
> 
> As I wrote this I recalled that there was different Helm behavior for YAML and TPL files, perhaps this is why we have things as they are?

Yes, YAMLs are directly rendered to output so one cannot define templates in those.  Except _*.yaml which is are partial templates. https://helm.sh/docs/topics/library_charts/

Note: even if we use partial templates (as in _compactor-pdb.yaml) , it needs to be invoked in Mimir and GEM by name (same as named template) and has to be in a different file - because the context issue is exactly the same. Maybe referencing it would be slightly different, so I'll check out that, maybe we can avoid templates in global namespace with them.

krajorama:
> > Is it only by convention that we have all the named templates in the `_helpers.*.tpl` files? Perhaps it makes sense for each named template to be in the template file? For example, having the compactor PDB named template defined and used in the `compactor.-pdb.yaml`.
> > As I wrote this I recalled that there was different Helm behavior for YAML and TPL files, perhaps this is why we have things as they are?
> 
> Yes, YAMLs are directly rendered to output so one cannot define templates in those. Except _*.yaml which is are partial templates. https://helm.sh/docs/topics/library_charts/
> 
> Note: even if we use partial templates (as in _compactor-pdb.yaml) , it needs to be invoked in Mimir and GEM by name (same as named template) and has to be in a different file - because the context issue is exactly the same. Maybe referencing it would be slightly different, so I'll check out that, maybe we can avoid templates in global namespace with them.

Never mind, partial template is just a different name for a "named template" https://helm.sh/docs/chart_template_guide/named_templates/ . Anyway, I'll test your suggestion to have each template in its own file, maybe it will make the diff nicer at least

trevorwhitney:
@krajorama curious about this comment:

> It is one step short of merging the mimir-distributed chart with enterprise-metrics.

What is the argument for not just having one chart? There could be a boolean that controls which image to use (enterprise or OSS), as well as whether or not to deploy the enterprise templates.

krajorama:
> @krajorama curious about this comment:
> 
> > It is one step short of merging the mimir-distributed chart with enterprise-metrics.
> 
> What is the argument for not just having one chart? There could be a boolean that controls which image to use (enterprise or OSS), as well as whether or not to deploy the enterprise templates.
> 
> The enterprise image is publicly available, and the char itself is Apache 2 licensed for either version, if someone accidentally used the wrong value for the boolean, GEM with enterprise auth turned off, or without a valid token, should behave just like Mimir, right?

My main argument is that it would be confusing for people and hinder adoption. What can you expect from the chart? Would people be comfortable using an enterprise chart for free, would they always wonder whether it's allowed, whether they needed a license or not for feature. Which documentation to look at and when.

CC @09jvilla

RichiH:
The target audience is engineers; I believe those actually prefer the name "Mimir" over "GEM" -- I know that the same is true for "GEL" and "Loki".

As long as the boolean is made clear in the GEM docs alongside a note with "For the GEM stuff, please use Mimir charts" in nicer words, it should be fine.

If you're really concerned, you could build out a GEM chart in a subdirectory to make it easier for GEM customers to point their deployment tooling to, but beyond that, merging seems to be the best course of action as it saves us time _and_ lets the engineers run what they most likely want: Mimir (which happens to have a license and extra features).

krajorama:
The more I think about it the better a single chart sounds (well, single per deployment model I guess - mimir-distributed, mimir-simple-scaleable, mimir-readwrite-scaleable, right?). Also naming it mimir-something fixes my issue with adoption. And the fact that we fail if you try something enterprise without license also makes it safe. So I'm on board. @jdbaldry  @09jvilla

trevorwhitney:
I like this plan. In agreement with everything you mentioned @09jvilla. Let's plan to do the same with Loki. As for the separate repo plan I've already been having doubts about that. I agree CODEOWNERS solves some of those problems and a second repo probably costs us more in effort and confusion than it gains.

@krajorama sorry for all the back and forth on how to solve this templating issue followed by this totally new direction. Thanks for your effort and for spurring this conversation, I'm happy with where it got us!

krajorama:
Agreement reached. I'll move my GEM2.0 from the other chart to this `mimir-distributed` one - guarded by an `enterprise.enabled` value.

osg-grafana:
cc @KMiller-Grafana for context

PR #1186, opened 2022-04-06
[loki-simple-scalable] added OpenShift support

vlad-diachenko:
added SecurityContextConstraints that is used in OpenShift instead of PodSecurityPolicy


cyriltovena (approved):
LGTM

chaudum (approved):

PR #1188, opened 2022-04-06
[loki-simple-scalable] replace PR #979

zanhsieh:
This replacement for PR #979 since unable to get DCO sign-off.

Original quote:
----
Adding hpa resources for read/write loki-simple-scaling services. read looks like it was intended to have this but wasnt finished (values with no template files) and write didnt have either.

I think this may need a warning about enabling autoscaling on writes when not using an object_store backend.

### DIFF
--- a/charts/loki-simple-scalable/README.md
+++ b/charts/loki-simple-scalable/README.md
@@ -1,6 +1,6 @@
 # loki-simple-scalable
 
-![Version: 0.4.0](https://img.shields.io/badge/Version-0.4.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.5.1](https://img.shields.io/badge/AppVersion-2.5.1-informational?style=flat-square)
+![Version: 0.4.1](https://img.shields.io/badge/Version-0.4.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.5.1](https://img.shields.io/badge/AppVersion-2.5.1-informational?style=flat-square)
 
 Helm chart for Grafana Loki in simple, scalable mode
 
--- a/charts/loki-simple-scalable/README.md
+++ b/charts/loki-simple-scalable/README.md
@@ -1,6 +1,6 @@
 # loki-simple-scalable
 
-![Version: 0.3.3](https://img.shields.io/badge/Version-0.3.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.4.2](https://img.shields.io/badge/AppVersion-2.4.2-informational?style=flat-square)
+![Version: 0.3.4](https://img.shields.io/badge/Version-0.3.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.4.2](https://img.shields.io/badge/AppVersion-2.4.2-informational?style=flat-square)
 
 Helm chart for Grafana Loki in simple, scalable mode
 
--- a/charts/loki-simple-scalable/README.md
+++ b/charts/loki-simple-scalable/README.md
@@ -80,6 +80,7 @@ helm repo add grafana https://grafana.github.io/helm-charts
 | global.image.registry | string | `nil` | Overrides the Docker registry globally for all images |
 | global.priorityClassName | string | `nil` | Overrides the priorityClassName for all pods |
 | imagePullSecrets | list | `[]` | Image pull secrets for Docker images |
+| loki.commonConfig | object | `{"path_prefix":"/var/loki","replication_factor":1,"storage":{"filesystem":{"chunks_directory":"/var/loki/chunks","rules_directory":"/var/loki/rules"}}}` | Check https://grafana.com/docs/loki/latest/configuration/#common_config for more info on how to provide a common configuration |
 | loki.config | object | See values.yaml | Config file contents for Loki |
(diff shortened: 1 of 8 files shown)


charts/loki-simple-scalable/README.md
```diff
@@ -80,9 +80,9 @@ helm repo add grafana https://grafana.github.io/helm-charts
 | global.image.registry | string | `nil` | Overrides the Docker registry globally for all images |
 | global.priorityClassName | string | `nil` | Overrides the priorityClassName for all pods |
 | imagePullSecrets | list | `[]` | Image pull secrets for Docker images |
-| loki.commonConfig | object | `{"path_prefix":"/var/loki","replication_factor":1,"storage":{"filesystem":{"chunks_directory":"/var/loki/chunks","rules_directory":"/var/loki/rules"}}}` | Check https://grafana.com/docs/loki/latest/configuration/#common_config for more info on how to provide a common configuration |
```
trevorwhitney, line 83: we don't want to lose this section

charts/loki-simple-scalable/values.yaml
```diff
@@ -59,68 +59,53 @@ loki:
   existingSecretForConfig: ""
   # -- Config file contents for Loki
   # @default -- See values.yaml
-  config: |
+  config:
```
trevorwhitney, line 62: this should be a string

charts/loki-simple-scalable/values.yaml
```diff
@@ -59,68 +59,53 @@ loki:
...
       http_listen_port: 3100
 
     memberlist:
       join_members:
-        - {{ include "loki.fullname" . }}-memberlist
+        - "{{ include \"loki.fullname\" . }}-memberlist"
 
-    {{- if .Values.loki.commonConfig}}
     common:
-    {{- toYaml .Values.loki.commonConfig | nindent 2}}
```
trevorwhitney, line 74: don't want to lose this either

charts/loki-simple-scalable/values.yaml
```diff
@@ -188,6 +173,17 @@ prometheusRule:
 write:
   # -- Number of replicas for the write
   replicas: 1
+  autoscaling:
+    # -- Enable autoscaling for the write
+    enabled: false
+    # -- Minimum autoscaling replicas for the write
+    minReplicas: 1
```
trevorwhitney, line 180: let's set minReplicas to 3 so it will work with the default RF of 3

trevorwhitney (changes requested):
this looks like it's pretty out of date with the current state of main. we lost a bunch of changes, like making the config a string and extracting the common config section

charts/loki-simple-scalable/values.yaml
zanhsieh: this one won't fix due to line 63 auth_enabled: false is not string.

charts/loki-simple-scalable/values.yaml
zanhsieh: this one have dependencies and the value has been copied over. won't fix.

charts/loki-simple-scalable/README.md
zanhsieh: this part has been moved to under loki.config.

zanhsieh:
@trevorwhitney
Could you approve this PR please? Or you have some idea?

charts/loki-simple-scalable/templates/configmap.yaml
```diff
@@ -7,5 +7,5 @@ metadata:
     {{- include "loki.labels" . | nindent 4 }}
 data:
   config.yaml: |
-    {{- tpl (mergeOverwrite (tpl .Values.loki.config . | fromYaml) .Values.loki.structuredConfig | toYaml) . | nindent 4 }}
+    {{- tpl (toYaml .Values.loki.config) . | nindent 4 }}
```
trevorwhitney, line 10: this is going to effectively remove the `structuredConfig` section, is that desired?

charts/loki-simple-scalable/templates/read/statefulset-read.yaml
```diff
@@ -74,8 +76,6 @@ spec:
           volumeMounts:
             - name: config
               mountPath: /etc/loki/config
-            - name: tmp
```
trevorwhitney, line 77: why are we removing the `/tmp` volume? I'm not actually sure what it's used for, but I can't tell from this PR why it's being removed either.

charts/loki-simple-scalable/values.yaml
trevorwhitney: we are moving in the direction of string config fields for all the charts. I don't understand what you mean about line 63, and us not being able to include `auth_enabed: false` in the config string?

without this fix this PR cannot be merged.

charts/loki-simple-scalable/values.yaml
trevorwhitney: again I don't understand. what dependencies prevent us from having a common config section?

charts/loki-simple-scalable/values.yaml
```diff
@@ -59,68 +59,54 @@ loki:
...
   storageConfig: {}
-  # -- Uncomment to configure each storage individually
-  # boltdb_shipper: {}
-  # filesystem: {}
-  # azure: {}
-  # gcs: {}
-  # s3: {}
-
-  # -- Structured loki configuration, takes precedence over `loki.config`, `loki.schemaConfig`, `loki.storageConfig`
-  structuredConfig: {}
```
trevorwhitney, line 123: again, why are we removing this.

trevorwhitney (changes requested):
This PR is diverging pretty strongly from some common patterns we have across our helm charts, most notably being using strings for the config field. There needs to be a much wider discussion before we make the following changes:

1. Remove the structured config
2. Change the config from a string to a structured value
3. Remove the common config section

If the goal is to get autoscaling merged, I would split this PR to only include those sections, as that could be merged.

Issue #1178, opened 2022-04-05
[mimir-distributed] Option to disable AlertManager


paalkr:
The Cortex helm chart has the ability to disable AlertManager.
https://github.com/cortexproject/cortex-helm-chart/blob/b69318fb683d4c56ca3e1658b87024596928c221/values.yaml#L161-L162

I think Mimir should have this capability as well. We do run an external highly available AlertManager, so the AlertManager that ships with Mimir isn't needed.

krajorama:
please reopen if it is still an issue, thank you

PR #1191, opened 2022-04-07
[mimir-distributed] mimir service flags

krajorama:
Fix missing if compactor.enabled
Add alertmanager.enabled flag for the use case of having an external alertmanager

Implements #1178


charts/mimir-distributed/templates/compactor/compactor-servmon.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.compactor.enabled -}}
```
pracucci, line 1: Compactor is a required component in Mimir. It shouldn't be disabled.

krajorama: artifact from starting from GEM chart I think, removed, will add to GEM changelog

56quarters (approved):
LGTM

PR #1201, opened 2022-04-08
[mimir-distributed] Bump PodDisruptionBudget to policy/v1

micborens:
# Description
Hello,

We updated our Kubernetes cluster to 1.21 and we checked some deprecated like `PodSecurityPolicy` and `PodDisruptionBudget`.

To avoid _warnings_ during deployment, I updated `PodDisruptionBudget` from `v1beta` to `v1` following Kubernetes changelogs.

Readme is updated following helm-docs this command: `docker run --rm --volume "$(pwd):/helm-docs" -u $(id -u) jnorwood/helm-docs:v1.4.0`

Thx

### DIFF
--- a/charts/mimir-distributed/Chart.yaml
+++ b/charts/mimir-distributed/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 0.1.6
+version: 0.1.7
 appVersion: 2.0.0
 description: "Grafana Mimir"
 engine: gotpl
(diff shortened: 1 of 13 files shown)


krajorama:
Hi,  1.25+ kubernetes is some time out still as far as I can see so I don't think this is urgent - plus we might make it fail for some existing users.

micborens:
let's close this PR as it's done in the other PR linked

PR #1203, opened 2022-04-08
[mimir-distributed] mimir add enterprise deployment as option

krajorama:
In this PR we implement the minimum needed to enable Grafana Enterprise Metrics through the `enterprise.enabled` value.

Should not change anything when `enterprise.enabled` is false.


krajorama:
@zanhsieh please leave merging to me

charts/mimir-distributed/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 0.1.6
+version: 2.0.0
```
trevorwhitney, line 2: oh, I like keeping the chart version in line with the mimir version, but what happens if the chart requires a breaking change that does not require a corresponding app version update?

charts/mimir-distributed/templates/admin-api/admin-api-dep.yaml
```diff
@@ -0,0 +1,120 @@
+{{- if .Values.enterprise.enabled -}}
```
trevorwhitney, line 1: :+1:  like it

charts/mimir-distributed/templates/gateway/gateway-dep.yaml
```diff
@@ -0,0 +1,115 @@
+{{- if .Values.enterprise.enabled -}}
```
trevorwhitney, line 1: Do we need a gateway and an nginx? What features is the gateway currently providing that nginx isn't? An option here could just be changing the value of the `nginx.conf` based on whether or not enterprise is enabled (ie. to add the `/admin` routes). WDYT?

charts/mimir-distributed/templates/gateway/gateway-pdb.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.enterprise.enabled -}}
```
trevorwhitney, line 1: Does the gateway need a PDB? Failed writes will buffer and retry, failed queries can just be retried. Not sure I see the need like I do with the ingesters/admin api.

charts/mimir-distributed/templates/ingester/ingester-statefulset.yaml
```diff
@@ -16,7 +16,7 @@ spec:
       {{- include "mimir.ingesterSelectorLabels" . | nindent 6 }}
   updateStrategy:
     {{- toYaml .Values.ingester.statefulStrategy | nindent 4 }}
-  serviceName: {{ template "mimir.fullname" . }}-ingester-headless
+  serviceName: {{ template "mimir.fullname" . }}-ingester{{- if not .Values.useGEMLabels -}}-headless{{- end -}}
```
trevorwhitney, line 19: it's only headless for OSS? did I read that write, not sure I understand why?

charts/mimir-distributed/templates/nginx/ingress.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if not .Values.enterprise.enabled -}}
```
trevorwhitney, line 1: see comment about gateway, but WDYT about using nginx for both enterprise and OSS?

charts/mimir-distributed/templates/admin-api/admin-api-dep.yaml
```diff
@@ -0,0 +1,120 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    {{- toYaml .Values.admin_api.annotations | nindent 4 }}
+  labels:
+    {{- include "mimir.adminApiLabels" . | nindent 4 }}
+    {{- if .Values.useGEMLabels }}{{- include "mimir.gemExtraLabels" . | nindent 4 }}{{- end }}
```
trevorwhitney, line 9: what's the relationship between `Values.enterprise.enabled` and `Values.useGEMLabels`? Can we assume that we always want `Values.useGEMLabels` when in a template guarded by `Values.enterprise.enabled`?

trevorwhitney (commented):
Looks good, thanks @krajorama, really happy with this direction. I have a few thoughts, let me know what you think.

charts/mimir-distributed/templates/gateway/gateway-pdb.yaml
krajorama: I added it just for consistency, maybe no-one will actually set it

charts/mimir-distributed/Chart.yaml
krajorama: I think enterprise customers need 2.0.0 for continuity from 1.8.1 + this has breaking changes

charts/mimir-distributed/templates/ingester/ingester-statefulset.yaml
krajorama: upgrade does not work from 1.8.1 GEM otherwise, since serviceName is immutable

charts/mimir-distributed/templates/nginx/ingress.yaml
krajorama: see answer above :)

charts/mimir-distributed/templates/admin-api/admin-api-dep.yaml
krajorama: I think the two are orthogonal. The `Values.enterprise.enabled` turns on/off licensed features, and `Values.useGEMLabels` is selects between kubernetes standard labels and labels matching GEM 1.8.1 - the later required for the upgrade to work.

`useGEMLabels` is probably not the best name,  something like `enterprise.legacyLabels` would be better. IMHO the current way we generate labels is a bit complicated and has lots of copy paste, not to mention it generates the labels in a different order from GEM 1.8.1, so `helm diff` output is too long. I'm writing a PR to solve these issues.

krajorama: See https://github.com/grafana/helm-charts/pull/1212

09jvilla:
@krajorama will let the team review, but noticed (not sure if this is for a separate PR) that there is no update to the readme. I'd suggest we have some instructions on the steps needed to make the chart install GEM instead of Mimir. 

I'd like to move more of the docs on the Helm chart into grafana.com, but we probably need a step 1 to move most of the readme content in the mimir docs (grafana.com/mimir) and then a step 2 to start importing the mimir docs into the GEM docs, and then a step 3 to make the GEM necessary changes to the Helm chart install instructions. All of that is gonna be out of scope for the GEM 2.0 release, so we can just stick with the readme for now. cc @jdbaldry re: my comment on future docs

charts/mimir-distributed/scripts/create-servmon
```diff
@@ -34,22 +62,33 @@ function snake_case {
   sed -E -e 's/-/_/' <<<"$1"
 }
 
-camel_cased="$(camel_case "$1")"
-snake_cased="$(snake_case "$1")"
+# camel_cased="$(camel_case "${component}")"
```
dimitarvdimitrov, line 65:
```suggestion
```

charts/mimir-distributed/scripts/create-pdb
```diff
@@ -34,23 +62,45 @@ function snake_case {
   sed -E -e 's/-/_/' <<<"$1"
 }
 
-component="$1"
-camel_cased="$(camel_case "$1")"
-snake_cased="$(snake_case "$1")"
+# camel_cased="$(camel_case "${component}")"
```
dimitarvdimitrov, line 65:
```suggestion
```

dimitarvdimitrov: I guess you don't need the function above either

dimitarvdimitrov (commented):
Left a few minor comments around the docs you've added. Afraid I can't be of much more help

charts/mimir-distributed/scripts/create-pdb
dimitarvdimitrov: same issue in the other 3 scripts

charts/mimir-distributed/CHANGELOG.md
```diff
@@ -0,0 +1,239 @@
...
+- [BUGFIX]
+
+Entries should include a reference to the Pull Request that introduced the change.
+
+## 2.0.0
+
+* [CHANGE] **Breaking** for existing users of `mimir-distributed`: the naming convention is changed to have shorter resource names, as in `<release>-mimir-distributed-store-gateway` is now just `<release>-mimir-store-gateway`. To have the previous names, please specify `nameOverride: mimir-distributed` in the values. #1203
+* [CHANGE] The chart `enterprise-metrics` is renamed to `mimir-distributed`. #1203
+* [CHANGE] **Breaking** Configuration for Grafana Enterprise Metrics is now in the value `mimir.config` as a helm template **string**.
+  Please consult the [Grafana Enterprise Migration Guide](https://grafana.com/docs/enterprise-metrics/latest/) to learn more about how to upgrade the configuration.
```
jdbaldry, line 19:
```suggestion
  Please consult the [Grafana Enterprise Migration Guide](https://grafana.com/docs/enterprise-metrics/latest/migrating-from-gem-1.7/) to learn more about how to upgrade the configuration.
```

charts/mimir-distributed/README.md
```diff
@@ -19,28 +19,107 @@ Kubernetes: `^1.10.0-0`
...
+
+To install the chart with licensed features enabled, using a local Grafana Enterprise Metrics license file called `license.jwt`, provide the license as a value and set the `enterprise.enabled` value to `true`.
+
+```bash
+helm install <cluster name> grafana/mimir-distributed --set 'enterprise.enabled=true' --set-file 'license.contents=./license.jwt'
+```
+
+### Upgrade from a previous version of Grafana Enterprise Metrics
+
+Please consult the [migration guide](https://grafana.com/docs/enterprise-metrics/latest/) for details on how to prepare the configuration. Prepare a custom values file, with the contents:
```
jdbaldry, line 93:
```suggestion
Please consult the [migration guide](https://grafana.com/docs/enterprise-metrics/latest/migrating-from-gem-1.7/) for details on how to prepare the configuration. Prepare a custom values file, with the contents:
```

jdbaldry (commented):
I provided some feedback on the README that actually needs to be reflected in the template. Sorry about that!

Looks really great to me!

charts/mimir-distributed/CHANGELOG.md
```diff
@@ -0,0 +1,239 @@
...
+  Except for the following parameters specified as command line parameters in the Pod templates,
+  everything is now set in this string-typed value, giving a definitive source of configuration.
+  Exceptions:
+    > The `-target=` must be provided individually.\
+    The `-config.file=` obviously.\
+    The `-activity-tracker` for convenience.\
```
09jvilla, line 25: I'm curious why this one.

charts/mimir-distributed/CHANGELOG.md
```diff
@@ -0,0 +1,239 @@
...
+    User defined arguments from `.<service>.extraArgs`.
+* [CHANGE] **Breaking** Kubernetes object labels now follow the [kubernetes standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) (e.g. `app.kubernetes.io/component=ingester`). To enable smooth upgrade and compatibility with previous Grafana Enterprise Metrics helm chart, the value `enterprise.legacyLabels` should be set to `true`.
```
09jvilla, line 27:
```suggestion
* [CHANGE] **Breaking** Kubernetes object labels now follow the [kubernetes standard](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) (e.g. `app.kubernetes.io/component=ingester`). To enable smooth upgrade and compatibility with previous Grafana Enterprise Metrics Helm chart, the value `enterprise.legacyLabels` should be set to `true`.
```

charts/mimir-distributed/CHANGELOG.md
```diff
@@ -0,0 +1,239 @@
...
+* [CHANGE] **Breaking** Ingesters only support `StatefulSet` from now on as chunks storage was removed in favour of blocks storage.
+* [CHANGE] **Breaking** Compactor is a required component, the value `compactor.enabled` is removed.
+* [CHANGE] **Breaking** The configuration parameter `server.http_listen_port` and `server.grpc_listen_port` cannot be changed from their defaults.
+* [CHANGE] The default for `ingester.ring.replication_factor` is now 3 and there will be 3 ingesters started even with the default `values.yaml`.
+  On the other hand, Pod anti affinity is turned off by default to allow single node deployment.
+* [FEATURE] Upgrade to [Grafana Enterprise Metrics v2.0.0](https://grafana.com/docs/enterprise-metrics/latest/)
```
09jvilla, line 33:
```suggestion
* [FEATURE] Upgrade to [Grafana Enterprise Metrics v2.0.0](https://grafana.com/docs/enterprise-metrics/v2.0.x/)
```

charts/mimir-distributed/CHANGELOG.md
```diff
@@ -0,0 +1,239 @@
...
+* [FEATURE] Reworked chart to enable installing Grafana Mimir open source software version without licensed features.
+* [FEATURE] Added the value `nameOverride` to enable migration from Cortex helm chart.
+* [FEATURE] The alertmanager can be disabled with `alertmanager.enabled: false`, to support the use case of external alertmanager.
+* [FEATURE] Added definitions of `ServiceMonitor` objects for Prometheus monitoring. Configuration via the `serviceMonitor` values. This enables partial functionality of Grafana Mimir dashboards out of the box - without alerts and recording rules pre-loaded.
```
09jvilla, line 37:
```suggestion
* [FEATURE] Added definitions of `ServiceMonitor` objects for Prometheus monitoring. Configuration is done via the `serviceMonitor` values. This enables partial functionality of Grafana Mimir dashboards out of the box - without alerts and recording rules pre-loaded.
```

charts/mimir-distributed/CHANGELOG.md
```diff
@@ -0,0 +1,239 @@
...
+* [BUGFIX] Remove unused value `.useExternalLicense`.
+
+## Entries from enterprise-metrics chart
+
+## 1.8.1
+
+* [ENHANCEMENT] Support Grafana Mimir monitoring mixin labels by setting container names to the component names.
+  This will make it easier to select different components in cadvisor metrics.
+  Previously, all containers used "enterprise-metrics" as the container name.
+  Now, the ingester Pod will have a container name "ingester".
```
09jvilla, line 50:
```suggestion
  Now, for example, the ingester Pod will have a container name "ingester" rather than "enterprise-metrics".
```

09jvilla:
@krajorama left a few minor suggestions on the README and Changelog but overall both LGTM.

charts/mimir-distributed/CHANGELOG.md
krajorama: I wanted to have a single configuration for this and keep it in the config, but I got a comment that I should set it for services that don't use it (i.e. anything on write path).

krajorama: It turns out to be incompatible with the mimirtool convert, so this is now removed and fixed.

charts/mimir-distributed/templates/validate.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.config }}
-{{- fail "Top level 'config' is not allowed, configuration should be provided as a string in 'mimir.config' and can contain template expressions. Alternatively configuration can be provided as an external secret." }}
+{{- fail "Top level 'config' is not allowed. In 'mimir.config', provide the configuration as a string that can contain template expressions. Alternatively, you can provide the configuration as an external secret." }}
```
osg-grafana, line 2: Thank you! Looks good. :)

osg-grafana (approved):
Approving the change to the `charts/mimir-distributed/templates/validate.yaml` file.

Logiraptor (approved):
This PR is huge but it's not going to get any smaller. :shipit: 

I say we keep an eye on things and address any further issues with smaller PRs as quickly as possible over the coming days + weeks.

PR #1210, opened 2022-04-08
[tempo-distributed] Add ingester block configurations

svennjegac:
Hi,

These configurations are [present in grafana tempo official docs](https://grafana.com/docs/tempo/latest/configuration/#ingester), however those fields are not exposed through helm chart.

I tried to add these fields manually to `tempo-distributed` config map, and afterwards I checked `ingester/status/config` endpoint. Configurations were successfully consumed by ingester pods.

I would like to expose these fields, because with some tweaking, ingesters memory footprint can be altered significantly.
Also, if this PR gets accepted, I would be glad to add few more contributions to the repository, and to expose few additional fields.

Thanks


mapno (approved):
Hi! Thanks for the contribution.

Generally, we try to avoid adding specific configs for options that can be configured via [flags](https://github.com/grafana/tempo/blob/main/modules/ingester/config.go#L43-L45). But these are important configs, I think it's good to have them there.

PR #1212, opened 2022-04-09
[mimir-distributed] Implement labels, podLabels, selectors and resourceName in a common way

krajorama:
Remove the individual helpers from each service, there was no
difference in the labels/selectors.
Fixes wrong part-of/memberlist label for non memberlist services.
Rename useGEMLabels to enterprise.legacyLabels.
Fix order of labels so it's the same as in the enterprise-metrics chart and similar to cortex resulting in a much shorter helm diff for migrations.
Fix toplevel object labels.
Make license secret name and config secret name a template to align with the rest of the object names.


dimitarvdimitrov (approved):
Trusting the diff you found, LGTM

> Ok, `helm diff` from previous version of chart shows the expected changes:
> * reorder of labels (cortex diff reduction)
> * removed part-of memberlist label from query-frontend and overrides exporter
> * fixed dashboard labels:
>    ```console
>       -           replacement: "dev/query_frontend"
>       +           replacement: "dev/query-frontend"
>       -           replacement: "dev/store_gateway"
>       +           replacement: "dev/store-gateway"
>    ```
> * name of config  runtime config objects calculated, not hardcoded anymore
> * correct labels for:
>    * PodSecurityPolicy
>    * Role
>    * RoleBinding
>    * ServiceAccount
>    * config Secret
>    * runtime ConfigMap

krajorama:
Tested install , upgrade from Mimir and migration from Cortex for regression.

PR #1225, opened 2022-04-13
[mimir-distributed] fix prod values capped-small and capped-large

krajorama:
Removes wrong, redundant config snippet and adds affinity rules - should be the same as small and large yaml.


jdbaldry (approved):
LGTM

PR #1230, opened 2022-04-13
[mimir-distributed] Bump helm-docs version to not force contributors on older vesion

krajorama:
Start updating charts for helm-docs 1.8.1. We can only update and release one chart at a time , so put ones that fail linting on helmdocsignore for now.

Old version renders chart dependencies differently:
old: | https://helm.min.io/ | minio | 8.0.10 |

new: | https://helm.min.io/ | minio(minio) | 8.0.10 |

Old version missing chart alias.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Check Docs
         run: |
-          docker run --rm --volume "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.4.0
+          docker run --rm --volume "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.8.1
           if ! git diff --exit-code; then
               echo "Documentation not up to date. Please run helm-docs and commit changes!" >&2
               exit 1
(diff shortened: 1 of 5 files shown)


.github/workflows/linter.yml
```diff
@@ -15,7 +15,32 @@ jobs:
...
+          ./charts/enterprise-logs-simple\n
+          ./charts/promtail\n
+          ./charts/fluent-bit\n
+          ./charts/enterprise-metrics\n
+          ./charts/loki\n
+          ./charts/agent-operator\n
+          ./charts/loki-stack\n
+          ./charts/tempo\n
+          ./charts/tempo-vulture\n
+          ./charts/loki-simple-scalable\n"
```
zanhsieh, line 33: Why we need this? Doesn't add charts to [.helmdocsignore](https://github.com/grafana/helm-charts/blob/main/.helmdocsignore) would help?

krajorama: nope, I want to introduce helm-docs v1.8.1 so people don't have to have an old version around, but simply switching means that all charts need to be modified, but we can only modify one chart at a time, so we need some mechanism to gradually change the helm-docs version. So far I couldn't fix it, I'll try again tomorrow my time (UTC+2)

krajorama: oh, wait I think I understand what you mean - that seems like a pretty good idea if it works :) going to try tomorrow, or feel free to try it yourself

vlad-diachenko (approved):
LGTM

PR #1231, opened 2022-04-13
[loki] max_entries_limit_per_query to it's default value

gcaracuel:
This PR relates to https://github.com/grafana/helm-charts/pull/966

Query limit is changed via max_entries_limit https://grafana.com/docs/loki/latest/configuration/#limits_config which default value is 5000 but not affecting here when using Dashboard sidecar as the datasource is limited by default to 1000.

Signed-off-by: Guillermo Caracuel <guillermo.caracuel@predictiva.io>

### DIFF
--- a/charts/loki/Chart.yaml
+++ b/charts/loki/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: loki
-version: 2.11.0
+version: 2.11.1
 appVersion: v2.5.0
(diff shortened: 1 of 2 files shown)


gcaracuel:
This does not really change a lot the Helm Chart but moves Grafana defaults 1000 rows to 5000 rows that is default value for Loki, I still think we should merge this.

zanhsieh (approved):

PR #1232, opened 2022-04-13
[loki-stack] max_entries_limit_per_query to it's current value

gcaracuel:
This PR relates to https://github.com/grafana/helm-charts/pull/966

Query limit is changed via max_entries_limit https://grafana.com/docs/loki/latest/configuration/#limits_config which default value is 5000 but not affecting here when using Dashboard sidecar as the datasource is limited by default to 1000.

### DIFF
--- a/charts/loki-stack/templates/datasources.yaml
+++ b/charts/loki-stack/templates/datasources.yaml
@@ -21,6 +21,8 @@ data:
       url: http://{{(include "loki.serviceName" .)}}:{{ .Values.loki.service.port }}
       version: 1
       isDefault: {{ .Values.loki.isDefault }}
+      jsonData:
+        maxLines: {{ .Values.grafana.sidecar.datasources.enabled }}
 {{- end }}
 {{- if .Values.prometheus.enabled }}
     - name: Prometheus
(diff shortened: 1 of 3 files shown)


charts/loki-stack/templates/datasources.yaml
```diff
@@ -21,6 +21,8 @@ data:
       url: http://{{(include "loki.serviceName" .)}}:{{ .Values.loki.service.port }}
       version: 1
       isDefault: {{ .Values.loki.isDefault }}
+      jsonData:
+        maxLines: {{ .Values.grafana.sidecar.datasources.enabled }}
```
zanhsieh, line 25: Shouldn't be:
{{ .Values.grafana.sidecar.datasources.maxLines }}

zanhsieh (changes requested):

charts/loki-stack/templates/datasources.yaml
gcaracuel: totally, my bad. Migrated this from another PR and copy pasted

zanhsieh (approved):

PR #1233, opened 2022-04-13
[mimir-distributed] naginx regex location blocks

dimitarvdimitrov:
* `/distributor`
* `/alertmanager`
* `/multitenant_alertmanager`
* `/api/v1/rules`
* `/prometheus/rules`
* `/prometheus/config/v1/rules`

are all prefixes to endpoints, so the nginx proxy rules need to reflect that

This was the source of this bug report https://grafana.slack.com/archives/C039863E8P7/p1649793055645919

### DIFF
--- a/charts/mimir-distributed/values.yaml
+++ b/charts/mimir-distributed/values.yaml
@@ -1260,13 +1260,13 @@ nginx:
           location {{ template "mimir.prometheusHttpPrefix" . }}/api/v1/rules {
             proxy_pass      http://{{ template "mimir.fullname" . }}-ruler.{{ .Release.Namespace }}.svc.cluster.local:{{ include "mimir.serverHttpListenPort" . }}$request_uri;
           }
-          location = /api/v1/rules {
+          location /api/v1/rules {
             proxy_pass      http://{{ template "mimir.fullname" . }}-ruler.{{ .Release.Namespace }}.svc.cluster.local:{{ include "mimir.serverHttpListenPort" . }}$request_uri;
           }
           location {{ template "mimir.prometheusHttpPrefix" . }}/api/v1/alerts {
             proxy_pass      http://{{ template "mimir.fullname" . }}-ruler.{{ .Release.Namespace }}.svc.cluster.local:{{ include "mimir.serverHttpListenPort" . }}$request_uri;
           }
-          location = {{ template "mimir.prometheusHttpPrefix" . }}/rules {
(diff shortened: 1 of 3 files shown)


charts/mimir-distributed/values.yaml
```diff
@@ -1235,38 +1235,38 @@ nginx:
           {{- end }}
 
           # Distributor endpoints
-          location /distributor {
```
krajorama, line 1238: this is already a prefix match (= would be exact match)

charts/mimir-distributed/values.yaml
```diff
@@ -1235,38 +1235,38 @@ nginx:
...
+          location ~ /distributor {
             proxy_pass      http://{{ template "mimir.fullname" . }}-distributor.{{ .Release.Namespace }}.svc.cluster.local:{{ include "mimir.serverHttpListenPort" . }}$request_uri;
           }
           location = /api/v1/push {
             proxy_pass      http://{{ template "mimir.fullname" . }}-distributor.{{ .Release.Namespace }}.svc.cluster.local:{{ include "mimir.serverHttpListenPort" . }}$request_uri;
           }
 
           # Alertmanager endpoints
-          location {{ template "mimir.alertmanagerHttpPrefix" . }} {
```
krajorama, line 1246: already a prefix match

charts/mimir-distributed/values.yaml
```diff
@@ -1235,38 +1235,38 @@ nginx:
...
+          location ~ {{ template "mimir.alertmanagerHttpPrefix" . }} {
             proxy_pass      http://{{ template "mimir.fullname" . }}-alertmanager.{{ .Release.Namespace }}.svc.cluster.local:{{ include "mimir.serverHttpListenPort" . }}$request_uri;
           }
-          location = /multitenant_alertmanager/status {
+          location ~ /multitenant_alertmanager {
```
krajorama, line 1249: what other endpoints are there under multitenant_alertmanager?

dimitarvdimitrov: you can find them here https://github.com/grafana/mimir/blob/7f5bf4b7247af57279380226293946f535c43467/pkg/api/api.go#L189-L199

krajorama: I see some other, but our own document says not to expose to users: https://grafana.com/docs/mimir/latest/operators-guide/reference-http-api/#alertmanager-configs

krajorama (approved):
LGTM, tested locally that it works

Issue #1214, opened 2022-04-11
[Promtail] extraClientConfigs breaking change


dungdm93:
Before, I have one `lokiAddress` and I use `extraClientConfigs` to add extra config/labels before send logs to Loki server.
![image](https://user-images.githubusercontent.com/6848311/162719115-0384c82f-0782-4341-8908-73ae16bcc1f8.png)
![image](https://user-images.githubusercontent.com/6848311/162720008-a375bb3c-ce88-4809-a919-e76d652bbf47.png)

After #1119, `extraClientConfigs` seem used to add extra clients, instead of extra config of existing client. So that will breaking my use-case:
![image](https://user-images.githubusercontent.com/6848311/162720464-98468a91-c81c-4953-8c26-6747ed387c23.png)

dungdm93:
I can make a PR for this, but I don't known which one you prefer:
1. Rollback #1119 and add another key for this like `extraClients`
2. make `lokiAddress` become a map, instead of string
```
lokiAddress:
   url: http://loki-gateway/loki/api/v1/push
   // more config can be added here
```
then:
```
    clients:
      - {{ toYaml .Values.config.lokiAddress }}
```

tobygale:
We've hit the same issue.
We use `extraClientConfigs` to set the `tenant_id` in places where we cannot inject the `X-Scope-OrgID` header.
Just had to rollback to the previous version.

jimgus:
Hit this issue as well when trying to add `tls_config` in `extraClientConfigs `

secustor:
We have ran into this today too. We use it to set the external labels and authentication methods .

I suggest the introduction of a new variable instead of replacing `lokiAddress` with a map. 
```
lokiClientConfig:
  external_labels:
     aLabel: aValue
```

```
    clients:
      - url: {{ .Values.config.lokiAddress }}
        {{ toYaml .Values.config.lokiClientConfig | nindent 4 }}
      {{- with .Values.config.snippets.extraClientConfigs }}
      {{- toYaml . | nindent 2 }}
      {{- end }}
```

fabienpelletier:
same, I'm using the `extraClientConfigs` to set `basic_auth` and this change breaks it.

ambis:
> same, I'm using the `extraClientConfigs` to set `basic_auth` and this change breaks it.

Same here. I have to re-write the whole file, which is not update proof.

```yaml
    extraEnv:
      - name: PROMTAIL_PASSWORD
[...]

PR #1242, opened 2022-04-14
[promtail] fix: rework of #1119 to preserve backwards compatability/features

ts-mini:
A potential solution for the problem in #1214

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.5.0
-version: 4.2.1
+version: 5.0.0
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
--- a/charts/promtail/README.md
+++ b/charts/promtail/README.md
@@ -24,6 +24,10 @@ A major chart version change indicates that there is an incompatible breaking ch
 
 ### From Chart Versions < 3.0.0
 
+#### Upgrading to >5.0.0
+
+* [breaking] lokiAddress and associated configuation in snippets.extraClientConfigs are removed and replaced with a single config.clients array.
+
 #### Notable Changes
 
 * Helm 3 is required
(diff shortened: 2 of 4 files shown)


secustor (approved):

ducnm0711:
Hi any update on merging this?
This seems to be a downgrade rather than upgrade to me :
- With chart version 3.11.0
```yaml
image:
  tag: 2.5.0
config:
  lokiAddress: https://loki/api/v1/push
  snippets:
    extraClientConfigs: |
      tenant_id: tenant-a
```
- For version 4.2.0, you need to parse full `file: |` + client ` {{- toYaml . | nindent 4 }}`
```yaml
config:
  lokiAddress: https://loki/api/v1/push
  snippets:
    extraClientConfigs:
      tenant_id: tenant-a

  file: |
    server:
      log_level: {{ .Values.config.logLevel }}
      http_listen_port: {{ .Values.config.serverPort }}
    clients:
      - url: {{ tpl .Values.config.lokiAddress . }}
      {{- with .Values.config.snippets.extraClientConfigs }}
      {{- toYaml . | nindent 4 }}
      {{- end }}
    positions:
      filename: /run/promtail/positions.yaml
    scrape_configs:
      {{- tpl .Values.config.snippets.scrapeConfigs . | nindent 2 }}
      {{- tpl .Values.config.snippets.extraScrapeConfigs . | nindent 2 }}
```

eviln1:
Hi :wave: 
#1119 was correctly marked as a breaking change ( and bumped the major version ) because `.Values.config.snippets.extraClientConfigs` 's type was changed from `string` to `list`, and chart's users need to know about that and adjust their config if needed.

However, I think that `4.0.0` is broken in the sense where the template doesn't render a valid - or at least, expected - `promtail.yaml` config, when used as intended (ie. using a list of `extraClientConfigs`).

If this PR changes the type of `.Values.config.snippets.extraClientConfigs` to string back again, then I believe it should also be marked as a breaking change, and bumped up a major version.

ts-mini:
Sorry everyone, was a bit selfish and published my version locally to unblock.  I've merged in some changes (the major version bump) + an indent fix I never pushed. 

Apologies that this got lost on my plate of things to do.

aceat64:
> WDYT of removing `lokiAddress` and just exposing an array of clients?

I like this, the default in `values.yaml` could still have the Loki address clearly set, but also show an inline example of additional options. For example:
```yaml
# -- You can put here any keys that will be directly added to the config file's 'client' block.
# A minimum of one client is required.
clients:
  # The Loki address to post logs to.
  - url: http://loki-gateway/loki/api/v1/push
#    external_labels:
#      label: value
#    basic_auth:
#      username: loki
#      password: loki
#  - url: http://other-client-service.local/
```

ts-mini:
I made changes given the feedback, deployed to our dev environment and seems to be :+1:.   Also took a stab at adding some verbage into the changelog to better communicate the breaking change(s).

trevorwhitney:
Ahh, sorry I just saw you made the same changes as in https://github.com/grafana/helm-charts/pull/1425, which I just merged. I'm going to close this unless there's something from #1425 we missed that you have included here?

Issue #1244, opened 2022-04-14
[mimir-distributed] Getting an error when setting Ingress enable to True in Helm chart values file. 


BitKickerBHS:
When deploying the mimir-distributed using helm chart values.yaml file if the Ingress is enabled: true I get the following error during deployment. 

Error: template: mimir-distributed/templates/nginx/ingress.yaml:46:25: executing "mimir-distributed/templates/nginx/ingress.yaml" at <include "mimir.nginxFullname" $>: error calling include: template: no template "mimir.nginxFullname" associated with template "gotpl".

Could this be caused by the _helpers.tpl not being updated in the helm chart? If this is just an issue with my config I apologize for the disruption.

agrahamlincoln:
I hit this too, I looked around and couldn't find `mimir.nginxFullname` template defined anywhere in the helm chart

BitKickerBHS:
The following needs to added to the top of charts/mimir-distributed/templates/nginx/_helpers.tpl

{{/*
nginx fullname
*/}}
{{- define "mimir.nginxFullname" -}}
{{ include "mimir.fullname" . }}-nginx
{{- end }}

Correct nginx _helpers.tpl to define  mimir.nginxFullname based off mimir.fullname .

krajorama:
fix in the works in #1260

krajorama:
please test with 2.0.5, #1260 is merged

krajorama:
should be fixed , please reopen issue if not

PR #1254, opened 2022-04-16
[mimir-distributed] Update charts/mimir-distributed/templates/nginx/_helpers.tpl

BitKickerBHS:
Correct nginx _helpers.tpl to define  mimir.nginxFullname based off mimir.fullname . Currently helm deployment will fail if ingress is enabled in value file as noted in Issue #1244 

Signed-off-by: Brandon.Abernathy@BecksHybrids.com


krajorama:
should be fixed in #1260

krajorama:
alternate fix merged

PR #1261, opened 2022-04-19
[agent-operator] Bump helm-docs use in agent-operator to 1.8.1

krajorama:
Use latest helm-docs 1.8.1 for README.md generation in agent-operator

### DIFF
--- a/.helmdocsignore
+++ b/.helmdocsignore
@@ -5,7 +5,6 @@ charts/loki-stack
 charts/enterprise-metrics
 charts/enterprise-logs
 # TODO get rid of this list after all docs migrated to helm-docs:1.8.1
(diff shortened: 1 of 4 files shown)


rfratto:
I'm not sure I understand what's actually changed here. Is 1.8.1 already in use at the repo level and we had to change files to adjust for changes to listing rules?

zanhsieh (approved):

krajorama:
> I'm not sure I understand what's actually changed here. Is 1.8.1 already in use at the repo level and we had to change files to adjust for changes to linting rules?

Yes, I've bumped helm-docs to latest recently, but I could not update all charts at once, because we can only change one chart at a time.

PR #1263, opened 2022-04-19
[grafana] Add the ability to configure lifecycle hooks

joelmeier:
Thought it might be useful to have the ability to, for example, call the Grafana API at the start of the pod so I added lifecycle hooks to the _pod.tpl file

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -561,6 +561,9 @@ containers:
 {{ toYaml .Values.livenessProbe | indent 6 }}
     readinessProbe:
 {{ toYaml .Values.readinessProbe | indent 6 }}
+{{- if .Values.grafana.lifecycleHooks }}
+    lifecycle: {{ tpl (.Values.lifecycleHooks | toYaml) . | nindent 6 }}
+{{- end }}
     resources:
 {{ toYaml .Values.resources | indent 6 }}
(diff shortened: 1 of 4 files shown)


zanhsieh (approved):

charts/grafana/templates/_pod.tpl
```diff
@@ -561,6 +561,9 @@ containers:
 {{ toYaml .Values.livenessProbe | indent 6 }}
     readinessProbe:
 {{ toYaml .Values.readinessProbe | indent 6 }}
+{{- if .Values.grafana.lifecycleHooks }}
```
zanhsieh, line 564: Should be .Values.lifecycleHooks.

==> Linting charts/grafana
Error:  templates/: template: grafana/templates/deployment.yaml:49:10: executing "grafana/templates/deployment.yaml" at <include "grafana.pod" .>: error calling include: template: grafana/templates/_pod.tpl:564:14: executing "grafana.pod" at <.Values.grafana.lifecycleHooks>: nil pointer evaluating interface {}.lifecycleHooks

zanhsieh (changes requested):

charts/grafana/templates/_pod.tpl
joelmeier: Whoops my bad :)

zanhsieh (approved):

PR #1268, opened 2022-04-20
[enterprise-logs] Bump GEL to version 1.4

trevorwhitney:
* bump GEL version to 1.4
* ~use `appVersion` from `Chart.yaml` as default image tag~ (couldn't find a way to template this into the `loki-distributed` sub chart)
* don't skip `enterprise-logs` tests on ci

### DIFF
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -44,8 +44,8 @@ jobs:
       - name: Run chart-testing (install)
         run: |
           changed=$(ct list-changed --config ct.yaml)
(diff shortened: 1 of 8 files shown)


jdbaldry:
I think for testing to be effective in CI we will need to find a way to include a license secret or modify the behavior of the tokengen to make this work.

trevorwhitney:
@jdbaldry I agree that would make testing more effective, however, even running the test as is without a license was useful as I detected a breaking config change in the default config.

jdbaldry:
Oh I actually thought it wouldn't work without a license :) I'm glad to see that is not the case!

MichelHollands (approved):
LGTM

Issue #1221, opened 2022-04-12
[mimir-distributed] Separate gateway and server ports and add validation


krajorama:
Add validation to reject if either of
```yaml
server:
  http_listen_port: 8080
  grpc_listen_port: 9095
```
is changed in `mimir.config`. Since the port numbers are reused in the configuration and manifests via a named template, it is not possible to read them from the `mimir.config` itself, as it leads to infinite loop: config -> template for port number -> config -> .....

The gateway port for Grafana Mimir use cases can already be changed via `nginx.service.port: 80`. Introduce the same for `gateway` component: `gateway.service.port: 8080`  (set to 8080 for backwards compatibility).

[x] add validation, see `validate.yaml` #1290 
[x] add `gateway.service.port` and update `gateway-svc.yaml` and follow up where needed. #1270

PR #1270, opened 2022-04-21
[mimir-distributed] Make GEM gateway svc port configurable

dimitarvdimitrov:
The non-enterprise version of the chart has an nginx working as a
reverse proxy. That can have the exposed service port overridden in the
values.yaml. The enterprise equivalent does not allow this.

This PR changes the enterprise gateway to allow custom exposed ports.

Signed-off-by: Dimitar Dimitrov <dimitar.dimitrov@grafana.com>

### DIFF
--- a/charts/mimir-distributed/CHANGELOG.md
+++ b/charts/mimir-distributed/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 2.0.7
+
+* [ENHANCEMENT] Add option to modify the port for the GEM gateway service. #1267
+
 ## 2.0.6
 
 * [ENHANCEMENT] Add option for an ingress on GEM gateway. #1266
--- a/charts/mimir-distributed/CHANGELOG.md
+++ b/charts/mimir-distributed/CHANGELOG.md
@@ -13,7 +13,7 @@ Entries should include a reference to the Pull Request that introduced the chang
 
 ## 2.0.7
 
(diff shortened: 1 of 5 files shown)


charts/mimir-distributed/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 2.0.7
+
+* [ENHANCEMENT] Add option to modify the port for the GEM gateway service. #1267
```
krajorama, line 16: this is PR 1270, not 1267
```suggestion
* [ENHANCEMENT] Add option to modify the port for the GEM gateway service. #1270
```

charts/mimir-distributed/values.yaml
```diff
@@ -1460,6 +1460,8 @@ gateway:
   service:
     annotations: {}
     labels: {}
+    # If the port is left undefined, the service will listen on the same port as the pod
+    # port: 80
```
krajorama, line 1464: nitpick: we'd like to have proper docstrings later, suggests:
```suggestion
    # -- If the port is left undefined, the service will listen on the same port as the pod
    port: null
```

krajorama (commented):
some nitpicking

charts/mimir-distributed/CHANGELOG.md
dimitarvdimitrov: ops I forgot this one

krajorama (approved):
LGTM
I've ran `helm template` on it with and without setting the port.

krajorama:
Fixes part of #1221

PR #1205, opened 2022-04-08
[mimir-distributed] Add support for multiple sts/svc in store_gateway to be used in conju…

ryan-dyer-sp:
…nction with rollout-operator

PR #1275, opened 2022-04-22
[rollout-operator] rollout op fix subchartname

krajorama:
Recalculate the chart name
    
    Commonly the chart would be used as a subchart under the
    alias rollout_operator. Making it possible to use
    .Values.rollout_operator:  But that results in a Chart name that
    is not a valid kubernetes resource name. Sanitize the chart name
    and turn _ back into -.

Also use hardcoded fix name for container for dashboards.

### DIFF
--- a/charts/rollout-operator/Chart.yaml
+++ b/charts/rollout-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: rollout-operator
 description: "Grafana rollout-operator"
 type: application
-version: 0.1.0
+version: 0.1.1
 appVersion: v0.1.1
 home: https://github.com/grafana/rollout-operator
 kubeVersion: ^1.10.0-0
--- a/charts/rollout-operator/templates/deployment.yaml
+++ b/charts/rollout-operator/templates/deployment.yaml
@@ -31,7 +31,7 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
-        - name: {{ .Chart.Name }}
+        - name: rollout-operator
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
--- a/charts/rollout-operator/README.md
+++ b/charts/rollout-operator/README.md
@@ -4,7 +4,7 @@ Helm chart for deploying [Grafana rollout-operator](https://github.com/grafana/r
 
 # rollout-operator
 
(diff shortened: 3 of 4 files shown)


charts/rollout-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: rollout-operator
 description: "Grafana rollout-operator"
 type: application
-version: 0.1.0
+version: 0.1.1
```
dimitarvdimitrov, line 5: Is it possible that this PR breaks existing Helm installs, or makes the upgrades much more painful? What would happen if someone was already using with chart name `rollout_operator`? I am not so certain how likely that scenario is either.

krajorama: It's not possible to use with rollout_operator name because you get:
```
Error: INSTALLATION FAILED: ServiceAccount "krajo-rollout_operator" is invalid: metadata.name: Invalid value: "krajo-rollout_operator": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```
This came about because of subcharting, see Chart.yaml here: https://github.com/grafana/helm-charts/pull/1205/files

```
  - name: rollout-operator
    alias: rollout_operator
    version: 0.1.1
    condition: rollout_operator.enabled
```

krajorama: The alternative is to use nameOverride instead in the chart that is using this , hmm?

charts/rollout-operator/templates/deployment.yaml
```diff
@@ -31,7 +31,7 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
-        - name: {{ .Chart.Name }}
+        - name: rollout-operator
```
dimitarvdimitrov, line 34:
```suggestion
        - name: {{ include "rollout-operator.chartName" . }}
```

charts/rollout-operator/Chart.yaml
dimitarvdimitrov: ok I see now. This PR makes using `rollout_operator` as chart name possible. I am ok with the current approach.

dimitarvdimitrov (approved):
LGTM, modulo a possibly missed replacement (see above)

charts/rollout-operator/templates/deployment.yaml
krajorama: This was intentional , otherwise it's hard to write dashboards , see commit message https://github.com/grafana/helm-charts/pull/1275/commits/81c5656730273c261233a20c13746d6239ec1fff

PR #1289, opened 2022-04-25
[grafana] bump chart to latest grafana version

solidnerd:
Bumping Grafana Helm Chart to version `8.5.0`

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 6.26.5
-appVersion: 8.4.6
+version: 6.26.6
+appVersion: 8.5.0
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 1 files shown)


zanhsieh (approved):

tjsampson:
@solidnerd This bumps the Chart and App Version, but the image.tag value is still defaulted to 8.4.6 instead of 8.5.0. Not sure if that was intentional or not. I just noticed when I updated to the latest 6.26.6 Chart version, but still running version 8.4.6 of the app. I reinstalled with image.tag value override of 8.5.0 and it worked, but I would think it should be the default for this chart version.

r0fls:
If we want to bump the Grafana image: https://github.com/grafana/helm-charts/pull/1295

solidnerd:
@tjsampson I recognized it my self and for that reason I created a new PR https://github.com/grafana/helm-charts/pull/1294

PR #1303, opened 2022-04-26
[mimir-distributed] fix null port at gateway ingress

ortuman:
Set defaults for gateway ingress service ports.

### DIFF
--- a/charts/mimir-distributed/Chart.yaml
+++ b/charts/mimir-distributed/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 2.0.8
+version: 2.0.9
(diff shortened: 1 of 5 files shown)


charts/mimir-distributed/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 2.0.8
+version: 2.0.9
```
krajorama, line 2: Please update the CHANGELOG as well

ortuman: done @ 259d65c33bcd922c12157cc530c7eb5169afc492

Issue #1307, opened 2022-04-26
[mimir-distributed] both alertmanager and alertmanager-headless services discovered by the alertmanager ServiceMonitor


agrahamlincoln:
### The Problem

When using the `ServiceMonitor`s generated by mimir-distributed helm chart, alertmanager pods are discovered twice because the labels for `alertmanager` and `alertmanager-headless` services are very similar

This caused the following issue for us:
Using `mimir-mixin` and `mimir-distributed` helm chart together (with prometheus operator ServiceMonitors) causes the `MimirGossipMembersMismatch` alert to fire incorrectly.

### Background/context

The mimir-distributed helm chart can create prometheus-operator ServiceMonitors for prometheus to discover it's components and scrape their metrics.

The mimir-distributed helm chart also creates 2 services for alertmanager, a headless one with no ClusterIP and one with a ClusterIP. When compared to the mimir jsonnet, the [mimir jsonnet only creates a headless service](https://github.com/grafana/mimir/blob/533cef6ef8a564ec3d3c434f0a0a7020c0bcc501/operations/mimir/alertmanager.libsonnet#L90)

The `GossipMembersMismatch` alert in mimir-mixin determines if there is a difference in the number of pods discovered via prometheus (kubernetes service discovery) and the number of members each mimir component expects there to be currently deployed.

When using the mimir-mixin and the mimir-distributed helm chart together, this alert fires because the alertmanager replicas are each discovered twice by prometheus.

krajorama:
Thank you @agrahamlincoln for noticing and providing the fix.

PR #1308, opened 2022-04-26
[mimir-distributed] exclude headless services from ServiceMonitors to prevent duplication of prometheus scrape targets

agrahamlincoln:
this was how i fixed https://github.com/grafana/helm-charts/issues/1307 in our setup

I'm open to other, more elegant solutions, or a rename of the service label to something else.

Because the mimir-jsonnet only has a headless service, I was tempted to remove the non-headless service from the helm chart, but decided against that.

### DIFF
--- a/charts/mimir-distributed/CHANGELOG.md
+++ b/charts/mimir-distributed/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the Pull Request that introduced the change.
 
+## 2.0.14
+
+* [BUGFIX] exclude headless services from ServiceMonitors to prevent duplication of prometheus scrape targets #1308
+
 ## 2.0.13
 
 * [ENHANCEMENT] Removed `rbac.create` option. #1317
--- a/charts/mimir-distributed/Chart.yaml
+++ b/charts/mimir-distributed/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-version: 2.0.13
+version: 2.0.14
 appVersion: 2.0.0
 description: "Grafana Mimir"
 engine: gotpl
--- a/charts/mimir-distributed/README.md
+++ b/charts/mimir-distributed/README.md
@@ -4,7 +4,7 @@ Helm chart for deploying [Grafana Mimir](https://grafana.com/docs/mimir/v2.0.x/)
 
 # mimir-distributed
 
-![Version: 2.0.13](https://img.shields.io/badge/Version-2.0.13-informational?style=flat-square) ![AppVersion: 2.0.0](https://img.shields.io/badge/AppVersion-2.0.0-informational?style=flat-square)
+![Version: 2.0.14](https://img.shields.io/badge/Version-2.0.14-informational?style=flat-square) ![AppVersion: 2.0.0](https://img.shields.io/badge/AppVersion-2.0.0-informational?style=flat-square)
 
 Grafana Mimir
 
--- a/charts/mimir-distributed/templates/alertmanager/alertmanager-servmon.yaml
+++ b/charts/mimir-distributed/templates/alertmanager/alertmanager-servmon.yaml
@@ -25,6 +25,11 @@ spec:
   selector:
     matchLabels:
       {{- include "mimir.selectorLabels" (dict "ctx" $ "component" "alertmanager" "memberlist" true) | nindent 6 }}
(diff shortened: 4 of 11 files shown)


krajorama:
I don't have a better idea, without causing a breaking change, due to both service names in use in `mimir.config`. Would you mind also adding `serviceType: headless` to ingester, query-frontend, store-gateway service and monitor the same way ?

I'm starting an epic to collect proposed breaking changes: https://github.com/grafana/helm-charts/issues/1309

krajorama:
giving it a go myself

krajorama:
giving it a go myself

agrahamlincoln:
thanks @krajorama for your feedback! I've added the label to all other headless services and updated the changelog.

Please take a look and let me know if there are any additional changes you'd like me to make

agrahamlincoln:
btw, prometheus does not discover duplicate entries for store-gateway or ingester because the headless services only have `grpc` named port, but the query-frontend also has `http-metrics` port on the headless service. I will add `serviceType: headless` to the query-frontend servicemonitor, do you think we should still add it to the others for consistency?

krajorama:
> btw, prometheus does not discover duplicate entries for store-gateway or ingester because the headless services only have `grpc` named port. The query-frontend also has `http-metrics` port on the headless service and has the same issue. I will add `serviceType: headless` to the query-frontend servicemonitor, do you think we should still add it to the others for consistency?

I think what I messed up is the direction of dependency. The ServiceMonitor does not need to know what kind of service it's monitoring (headless or not), rather a service should say that it needs monitoring. Sorry to do this, but actually I think a right label to add would be `prometheus-monitored: "true"` to the services we want to monitor. So when it comes to removing the redundant services in 3.0.0 of the chart, we'll less likely to remove something we monitor (or at least fix it). As in:

```
diff --git a/charts/mimir-distributed/templates/alertmanager/alertmanager-servmon.yaml b/charts/mimir-distributed/templates/alertmanager/alertmanager-servmon.yaml
index cfb51c14..e9e5336f 100644
--- a/charts/mimir-distributed/templates/alertmanager/alertmanager-servmon.yaml
+++ b/charts/mimir-distributed/templates/alertmanager/alertmanager-servmon.yaml
@@ -24,6 +24,7 @@ spec:
   {{- end }}
   selector:
     matchLabels:
+      prometheus-monitored: "true"
       {{- include "mimir.selectorLabels" (dict "ctx" $ "component" "alertmanager" "memberlist" true) | nindent 6 }}
   endpoints:
     - port: http-metrics
diff --git a/charts/mimir-distributed/templates/alertmanager/alertmanager-svc-headless.yaml b/charts/mimir-distributed/templates/alertmanager/alertmanager-svc-headless.yaml
index 74a5332b..9bf75a12 100644
--- a/charts/mimir-distributed/templates/alertmanager/alertmanager-svc-headless.yaml
+++ b/charts/mimir-distributed/templates/alertmanager/alertmanager-svc-headless.yaml
@@ -5,6 +5,7 @@ kind: Service
 metadata:
   name: {{ include "mimir.resourceName" (dict "ctx" . "component" "alertmanager") }}-headless
   labels:
+    prometheus-monitored: "true"
     {{- include "mimir.labels" (dict "ctx" . "component" "alertmanager" "memberlist" true) | nindent 4 }}
     {{- with .Values.alertmanager.service.labels }}
     {{- toYaml . | nindent 4 }}
```

So my reasoning to add it to all relevant services is for clarity and trying to avoid later problems.

agrahamlincoln:
sounds good, given what you described, I think it makes sense to only include these labels when `serviceMonitors` are enabled. The additional `prometheus-monitored` label is not really relevant when serviceMonitors are disabled.

I also think we should be consistent about the `prometheus-monitored` label across all components, regardless of whether they have multiple services (headless & non-headless) or not.

I'll submit another update shortly

agrahamlincoln:
actually I just discovered this nearly-identical problem was solved in the loki-distributed helm chart here https://github.com/grafana/helm-charts/pull/1159

I like the MatchExpressions use here - instead of services defining that it should be monitored, some services define that they shouldn't be monitored with this label. There are likely to be less cases where there's a headless and normal service, and in each of those cases the headless service would be the one with the "dont-monitor" label.

krajorama (approved):
LGTM
Double checked affected services.
Checked helm diff
Checked that my prometheus still receives metrics after upgrade and alertmanager is now counted once only.

PR #1311, opened 2022-04-27
[mimir-distributed] add named port memberlist to pod container

ortuman:
Signed-off-by: Miguel Ángel Ortuño <ortuman@gmail.com>

Add memberlist container named port to all components that depend on it.

### DIFF
--- a/charts/mimir-distributed/templates/alertmanager/alertmanager-dep.yaml
+++ b/charts/mimir-distributed/templates/alertmanager/alertmanager-dep.yaml
@@ -80,6 +80,9 @@ spec:
             - name: grpc
               containerPort: {{ include "mimir.serverGrpcListenPort" . }}
               protocol: TCP
+            - name: memberlist
+              containerPort: {{ include "mimir.memberlistBindPort" . }}
+              protocol: TCP
           livenessProbe:
             {{- toYaml .Values.alertmanager.livenessProbe | nindent 12 }}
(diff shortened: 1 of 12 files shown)


charts/mimir-distributed/templates/alertmanager/alertmanager-dep.yaml
```diff
@@ -80,6 +80,9 @@ spec:
             - name: grpc
               containerPort: {{ include "mimir.serverGrpcListenPort" . }}
               protocol: TCP
+            - name: memberlist
```
krajorama, line 83: alertmanager has a statefulset deployment option also!

krajorama (changes requested):
Need to add the new named template to templates/_helpers.tpl.
If we chose to hardcode the number, we should add a validation against changing via mimir.config. Otherwise should have a default hardcoded value and try to grab the bind port via mimir.calculatedConfig (see other examples). I think we can go with the second option as it's not referenced anywhere in mimir.config, so no infinite loop....

charts/mimir-distributed/templates/alertmanager/alertmanager-dep.yaml
ortuman: done @ b9d9200db8a18a37fc6340fa95a512b0371cd9db

krajorama (approved):
LGTM

PR #1312, opened 2022-04-27
[grafana] set sizeLimit for k8s-sidecar emptyDir volume

jdgeisler:
Adding an optional sizeLimit for the k8s-sidecar emptyDir volumes. This sets the total amount of local storage required for the emptyDir volume. If not set, the default is nil which means that the limit is undefined.

Example:

```yaml
datasources:
  enabled: true
  sizeLimit: 1Gi
dashboards:
  enabled: true
  sizeLimit: 2Gi
```

### DIFF
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -708,6 +708,8 @@ sidecar:
       foldersFromFilesStructure: false
     # Additional dashboard sidecar volume mounts
     extraMounts: []
+    # Sets the size limit of the dashboard sidecar emptyDir volume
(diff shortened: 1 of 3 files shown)


zanhsieh (approved):

charts/grafana/values.yaml
pierreblais, line 1: SizeLimit Should be a string, not a map[].

pierreblais (commented):
SizeLimit isn't configurable because a map is awaited while a string should be set !

Issue #1300, opened 2022-04-26
[mimir-distributed] remove rbac.create value


krajorama:
This value is redundant, without also setting rbac.pspEnabled. (See also loki-distributed).
Seems safe to simply remove and replace with if on rbacEnabled.

krajorama:
cc: @patrickabrennan

PR #1317, opened 2022-04-28
[mimir-distributed] remove rbac.create option

ortuman:
Signed-off-by: Miguel Ángel Ortuño <ortuman@gmail.com>

~~This PR merges `rbac.create` and `rbac.pspEnabled` options into a single `rbacEnabled` flag.~~

Removed `rbac.create` option.

fixes: #1300

### DIFF
--- a/charts/mimir-distributed/values.yaml
+++ b/charts/mimir-distributed/values.yaml
@@ -223,7 +223,9 @@ mimir:
 # runtimeConfig provides a reloadable runtime configuration file for some specific configuration.
 runtimeConfig: {}
 
-rbacEnabled: true
+# RBAC configuration
+rbac:
+  pspEnabled: true
 
 # ServiceMonitor configuration
 serviceMonitor:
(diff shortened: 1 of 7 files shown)


charts/mimir-distributed/values.yaml
```diff
@@ -223,9 +223,7 @@ mimir:
 # runtimeConfig provides a reloadable runtime configuration file for some specific configuration.
 runtimeConfig: {}
 
-rbac:
-  create: true
-  pspEnabled: true
+rbacEnabled: true
```
krajorama, line 226: is this how loki has it? I recall different, also this would be a breaking change since rbac.pspEnabled would be gone as well, please double check

krajorama (changes requested):

charts/mimir-distributed/values.yaml
ortuman: aside from `pspEnabled` option, loki has the extra ` sccEnabled` (SecurityContextConstraints) option, and the rolebinding resource is rendered whenever either one is enabled.

if you consider, i could add this extra option as well. otherwise i think the `rbacEnabled` should be enough.

ortuman: forget my last comment, since as you have rightly pointed out this would mean a breaking change. let me rework this.

krajorama (approved):
LGTM

PR #1323, opened 2022-04-29
[tempo-distributed] Fix ingester values

ben-natan:
In the values, ingester.config.max_block_bytes was exposed but it was ingester.config.maxBlockBytes that was used in the configmap. Same goes for max_block_duration and complete_block_timeout.


joe-elliott:
Nice! Thanks for the fix!

PR #1325, opened 2022-04-29
[loki-simple-scalable] Make enterprise a configurable option

trevorwhitney:
This PR combines `enterprise-logs-simple` and `loki-simple-scalable` with the goal of

1. Making it easier to maintain charts by having fewer to maintain
2. Making it easier to switch between OSS Loki and GEL by simply toggling a config option

Setting `enterprise.enabled = true` will change the following:
1. Use the `enterprise-logs` image instead of the `loki` image
2. Mount a license to the pods. A license can be added via the `enterprise.license` setting
3. Add additional routes to the gateway's nginx config to include Admin endpoints exposed by GEL

This is in line with the strategy we are following for Mimir chart as well (see #1203)

### DIFF
--- a/charts/loki-simple-scalable/Chart.lock
+++ b/charts/loki-simple-scalable/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: minio
+  repository: https://helm.min.io/
+  version: 8.0.9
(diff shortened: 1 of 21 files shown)


MasslessParticle (approved):
This seems like a great idea. LGTM

trevorwhitney:
PR deprecating existing `enterprise-logs-simple` once this is merged: https://github.com/grafana/helm-charts/pull/1327

PR #1326, opened 2022-04-29
[mimir-distributed] Prototype tool to compare Helm with Jsonnet

Logiraptor:
This is a big PR, so I would recommend starting with [the README](https://github.com/grafana/helm-charts/tree/logiraptor/compare-helm-with-jsonnet/scripts/k8s-diff) to get a sense for what this tool does. Then take a look at the [rules files for mimir-distributed](https://github.com/grafana/helm-charts/tree/logiraptor/compare-helm-with-jsonnet/charts/mimir-distributed/k8s-diff/rules).

Super-high-level description:

1. Adds several new tools which are useful when comparing large numbers of yaml files, particular k8s manifests and Mimir configuration
2. Adds a tanka environment with the latest released mimir jsonnet
3. Adds a minimal Helm values file to make the deployment as close as possible between Helm and Jsonnet with the current options
4. Adds configuration for the `yaml-patch` tool that results in an empty diff b/w helm and jsonnet
5. Adds a script to render the Helm and Tanka manifests, then compare them with the tools

The idea is that we can slowly remove ignore rules from number 4 as we make Helm and Jsonnet more similar. The tools in this PR will make sure we don't introduce regressions over time and point out any additional drift that occurs.

Note that the current config for `k8s-diff` ignores pretty much everything in order to make the diff clean, but the idea is we can pretty rapidly start to include important properties into the diff output.

Please don't merge until we've had a chance to talk through it.


Logiraptor:
If you're just interested in the differences b/w helm and jsonnet, here are the changes necessary to make them *identical*:

1. I made changes to the [helm](charts/mimir-distributed/k8s-diff/helm-workspace/values.yaml) and
   [jsonnet](charts/mimir-distributed/k8s-diff/jsonnet-workspace/environments/default/main.jsonnet)
   configuration in order to enable / disable high-level features like alertmanager.
2. I then ignored several resources that only exist on one side: [see ignored_objects.yml](charts/mimir-distributed/k8s-diff/rules/ignored_objects.yaml)
3. I renamed the Helm resources to match jsonnet (e.g. querier instead of mimir-querier): [see renames.yaml](charts/mimir-distributed/k8s-diff/rules/renames.yaml)
4. I fixed "cosmetic" differences like service port naming, pod labels, etc:
   - [helm_labels.yaml](charts/mimir-distributed/k8s-diff/rules/helm_labels.yaml)
   - [memberlist.yaml](charts/mimir-distributed/k8s-diff/rules/memberlist.yaml)
   - [memcached.yaml](charts/mimir-distributed/k8s-diff/rules/memcached.yaml)
   - [services.yaml](charts/mimir-distributed/k8s-diff/rules/services.yaml)
5. Finally, I removed all remaining properties with differences (these are the ones we should consider fixing): [see ignored_fields.yaml](charts/mimir-distributed/k8s-diff/rules/ignored_fields.yaml)


For the Mimir Configuration, I generated the final configuration object for each component and similarly collected all changes necessary to make them *identical*.

See [config_diffs.yaml](charts/mimir-distributed/k8s-diff/rules/config_diffs.yaml) for a complete listing of fields that differ.

Logiraptor:
Some examples of the output possible with this tooling:

1. An example where the Mimir configuration differs

Screen Shot 2022-05-05 at 3 58 06 PM

2. An example where the kubernetes configuration differs

Screen Shot 2022-05-05 at 4 00 08 PM

scripts/k8s-diff/README.md
```diff
@@ -0,0 +1,274 @@
...
+### How it works
+
+yaml-patch copies yaml files from one directory to another and optionally performs some transformation to the data in between.
+
+Internally, yaml-patch has two main phases as indicated in the diagram below.
+
+1. Preprocessing
+    - During this phase, manifests are modified in order to eliminate superfluous differences that are deemed unimportant by the user
+2. Validation
+    - During this phase, debug information collected during preprocessing is used to validate that all rules defined in the config file actually impact the final output.
```
dimitarvdimitrov, line 79:
```suggestion
1. Patching
    - During this phase, manifests are modified in order to eliminate superfluous differences that are deemed unimportant by the user
2. Validation
    - During this phase, debug information collected during patching is used to validate that all rules defined in the config file actually impact the final output.
```

scripts/k8s-diff/README.md
```diff
@@ -0,0 +1,274 @@
...
+
+```mermaid
+flowchart LR
```
dimitarvdimitrov, line 82: TIL you can do this. amazing

scripts/k8s-diff/README.md
dimitarvdimitrov: to match the flow chart

charts/mimir-distributed/k8s-diff/rules/services.yaml
```diff
@@ -0,0 +1,94 @@
...
+    value: Service
+  - op: test
+    path: /metadata/name
+    value: overrides-exporter
+  - op: test
+    path: /spec/ports/1/name
+    value: grpc
+  steps:
+  - op: remove
+    path: /spec/ports/1
```
dimitarvdimitrov, line 49: Using indexes is slightly flaky. I have't read the code, but will this fail if the grpc port is the first one not the second one?

The docs say that the validation stage of yaml-patch should do that, but I'm not sure I understand those correctly. I see some of the patch rules explicitly work only on fields that exist in helm, so they can't affect jsonnet too. Yet, obviously, the yaml-patch succeeds for them.

charts/mimir-distributed/k8s-diff/rules/memberlist.yaml
```diff
@@ -0,0 +1,56 @@
+---
+patch_rules:
+
+- name: "Adjust gossip-ring to match"
+  match:
+  - op: test
+    path: /metadata/name
+    value: gossip-ring
```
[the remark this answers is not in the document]
dimitarvdimitrov: same goes for a few tests in services.yaml

scripts/k8s-diff/README.md
```diff
@@ -0,0 +1,274 @@
...
+  match: []JsonPatchOperation (optional)
+  steps: []JsonPatchOperation
+  remove_field: jsonpointer (optional), (e.g. /metadata/labels/name)
+  rename_object: (optional)
+    from: string, name to replace in /metadata/name
+    to: string, target name
+  rename_field: (optional)
+    from: jsonpointer (optional), (e.g. /metadata/labels/name)
+    to: jsonpointer (optional), (e.g. /metadata/labels/name)
+  <jsonpointer>: []any (optional)
```
dimitarvdimitrov, line 167: having matchers in two places is slightly confusing, wdyt about something like moving the jsonpointer matchers in the `match` block?

```yaml
- name: string
  match: 
    jsonpointer: any (optional)
    jsonpointer: any (optional)
    ...
  # the rest is same as you have it
  #steps: []JsonPatchOperation
  #remove_field: jsonpointer (optional), (e.g. /metadata/labels/name)
  #rename_object: (optional)
  #  from: string, name to replace in /metadata/name
  #  to: string, target name
  #rename_field: (optional)
  #  from: jsonpointer (optional), (e.g. /metadata/labels/name)
  #  to: jsonpointer (optional), (e.g. /metadata/labels/name)
```

and optionally making the key value of the jsonpointer a single value, not an array, because I didn't see it being used as an array

I think this will simplify reading the matchers. It will improve the visual distinction between `match` items and the `steps` items, which now look identical and their parent blocks (match and steps) have the same number of letters

dimitarvdimitrov (commented):
amazing effort! I've only reviewed the non-go parts and it looks great so far. I've left a few comments and questions, most importantly a suggestion on how we write the matchers.

charts/mimir-distributed/k8s-diff/jsonnet-workspace/lib/k.libsonnet
```diff
@@ -0,0 +1 @@
+import 'github.com/jsonnet-libs/k8s-libsonnet/1.18/main.libsonnet'
```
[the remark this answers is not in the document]
Logiraptor: TBH, me neither. I created it according to the GEM jsonnet-lib readme: https://github.com/grafana/jsonnet-libs/tree/master/enterprise-metrics#getting-started

charts/mimir-distributed/k8s-diff/rules/helm_labels.yaml
```diff
@@ -0,0 +1,55 @@
+---
+patch_rules:
+
+# Helm doesn't set a namespace when running helm template
+- name: "Set default namespace"
+  match:
+  - op: remove
+    path: /metadata
```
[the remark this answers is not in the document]
Logiraptor: You're right. This is a leftover from a previous setup where Mimir config files were also processed by the same rules file. In that case I had to filter it to only apply when looking at a k8s object. I'll remove it.

charts/mimir-distributed/k8s-diff/rules/memberlist.yaml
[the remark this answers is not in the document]
Logiraptor: Yeah absolutely. The shorthand was added later and I haven't cleaned everything up yet. I should be able to replace any `test` op in the `match` section with the shorthand.

charts/mimir-distributed/k8s-diff/rules/memcached.yaml
```diff
@@ -0,0 +1,88 @@
+---
+patch_rules:
+# NOTE: these rules depend on the rules in helm_labels.yaml to rename labels
+
+- name: "Rename /metadata/labels/memcached-queries to memcached-index-queries"
```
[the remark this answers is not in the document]
Logiraptor: `rename_object` is currently special-cased to only change `/metadata/name`, not `/metadata/labels/name`. I could see an argument to change a name label as well though (if it exists).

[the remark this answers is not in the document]
Logiraptor: Alternatively, a `set_label` shorthand could iterate through all the various places labels appear and set a new value. So:

```
/metadata/labels/xyz
/spec/selector/xyz
/spec/selector/matchLabels/xyz
/spec/template/metadata/labels/xyz
```

charts/mimir-distributed/k8s-diff/rules/services.yaml
Logiraptor: You're 100% right using an array index is flaky. Unfortunately jsonpointer doesn't have any smarter way to identify an array element. I've considered extending it, but have avoided that so far in order to hopefully keep the number of net-new concepts in this tool to a minimum.

In the case of this rule, if someone changed the grpc port to come first, then the `match` would not succeed. (`/spec/ports/1/name` != `grpc`). Then during validation `yaml-patch` will report that this rule didn't match any objects and will exit 1. It's not ideal, since the error message doesn't necessarily point out the actual problem. 

Secondly, each rule only has to effect a single object from either side of the diff in order to be considered valid. There are quite a few rules that only impact helm for example.

scripts/k8s-diff/README.md
Logiraptor: It's new! https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

scripts/k8s-diff/README.md
Logiraptor: I love this idea! Since `match` is an array, it might need to look just a little different:

```
- name: string
  match:
  - jsonpointer: any
  - op: test
    ...
```

This is slightly longer, but I think it's way clearer what's going on, and length of rule file is definitely not a goal to optimize for over understanding.

Logiraptor: Also, the multiple-value form is used a bunch in the latest version of the rules. It helps to DRY the config quite a bit in certain sections. I'll sync that to this branch in a bit.

Logiraptor:
I got interrupted by on-call duties, but I'll push some updates in line with the feedback as soon as I get a chance. 👍

charts/mimir-distributed/k8s-diff/jsonnet-workspace/lib/k.libsonnet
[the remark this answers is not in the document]
dimitarvdimitrov: from the same readme:

> k.libsonnet is a Jsonnet library for Kubernetes. Most other Jsonnet libraries for Kubernetes expect this file to exist in your JSONNET_PATH.

have you tried removing it? it shouldn't compile if it's required

charts/mimir-distributed/k8s-diff/rules/memcached.yaml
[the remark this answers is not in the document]
dimitarvdimitrov: oh I understand now. I'm more inclined to keep them verbose as they are now as opposed to making more complex shorthand operators.

I guess we'll tackle these changes one by one and not necessarily in big chunks. So having the changes documented more verbosely and more granularly makes sense to me.

Logiraptor:
Closing this PR in favor of the migrated version in the Mimir-repo: https://github.com/grafana/mimir/pull/2067

I'll take the feedback here into account shortly.

PR #1328, opened 2022-05-02
[agent-operator] Fix Grafana Agent finalizers errors on OpenShift

NiklasWagner:
Closes https://github.com/grafana/agent/issues/1660

I just added all /finalizers resources because I don't know which are required for all usecases. Feel free to reduce them but this configuration now works on OKD 4.10 / Kubernetes v1.23.3-2003+e419edff267ffa-dirty with a GrafanaAgent and MetricsInstance resource.

### DIFF
--- a/charts/agent-operator/templates/operator-clusterrole.yaml
+++ b/charts/agent-operator/templates/operator-clusterrole.yaml
@@ -20,7 +20,7 @@ rules:
   - metricsinstances/finalizers
   - logsinstances/finalizers
   - podlogs/finalizers
-  verbs: [get, list, watch, create, update, patch, delete]
+  verbs: [get, list, watch, update]
 - apiGroups: [monitoring.coreos.com]
   resources:
   - podmonitors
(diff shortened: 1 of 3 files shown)


hjet:
hey there and thanks for tracking this down! could you try reducing the verbs to `update` and see if that works for your use case?

thank you!!

charts/agent-operator/templates/operator-clusterrole.yaml
```diff
@@ -14,12 +14,25 @@ rules:
   - logsinstances
   - podlogs
   verbs: [get, list, watch]
+- apiGroups: [monitoring.grafana.com]
+  resources:
+  - grafanaagents/finalizers
+  - metricsinstances/finalizers
+  - logsinstances/finalizers
+  - podlogs/finalizers
+  verbs: [get, list, watch, create, update, patch, delete]
```
hjet, line 23:
```suggestion
  verbs: [get, list, watch, update]
```

hjet (commented):
just to reduce scope a bit, although not sure how much this matters...

hjet:
Closes https://github.com/grafana/agent/issues/1660

charts/agent-operator/templates/operator-clusterrole.yaml
NiklasWagner: Still seems to work with reduced scope, just pushed these changes.

zanhsieh (approved):

Issue #1315, opened 2022-04-28
[mimir-distributed] Consider using own memcached manifest instead of bitnami chart?


Skaronator:
Currently, the mimir-distributed helm-chart uses the bitnami memcached helm-chart instead of its own manifest files like the loki-distributed chart. The Bitnami chart doesn't work out-of-the-box on OpenShift/OKD due to SecurityContextContrains.

Installing mimir-distributed with memcached enabled results in the following error:

> FailedCreate  statefulset-controller  create Pod mimir-distributed-memcached-0 in StatefulSet mimir-distributed-memcached failed error: pods "mimir-distributed-memcached-0" is forbidden: unable to validate against any security context constraint: [provider "scc-custom-order-service": Forbidden: not usable by user or serviceaccount, provider "anyuid": Forbidden: not usable by user or serviceaccount, provider "scc-custom-default": Forbidden: not usable by user or serviceaccount, provider "uba-default": Forbidden: not usable by user or serviceaccount, provider "vault": Forbidden: not usable by user or serviceaccount, provider "uba-anti-virus-service": Forbidden: not usable by user or serviceaccount, provider "custom-postgres": Forbidden: not usable by user or serviceaccount, provider restricted: .spec.securityContext.fsGroup: Invalid value: [[]int64{1001}: 1001 is not an allowed group, spec.containers[0].securityContext.runAsUser: Invalid value: 1001: must be in the ranges: [1000660000, 1000669999], spec.containers[1].securityContext.runAsUser: Invalid value: 1001: must be in the ranges: [1000660000, 1000669999], provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "loki-distributed": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount, provider "scc-custom-storage": Forbidden: not usable by user or serviceaccount]

The mimir-distributed chart includes a PSP, which would _probably_ fix it, but that doesn't apply to sub-charts since they have their own SA/Role.
[...]

Issue #1309, opened 2022-04-27
[mimir-distributed] [Epic] collect breaking changes for 3.0.0 of the chart


krajorama:
Targeting end of June or early August 2022

- [ ] Update minimum required Kubernetes version from 1.10. Possibly to 1.21 ? Needs investigation.
- [ ] Remove redundant non headless services (ref: #1307) Break: configuration
- [ ] Use non-root image (ref: #1282) Break: existing files on PVC need to be chown-ed . 
- [ ] Move activity log somewhere from /data, as it causes warning in Alertmanager `{"caller":"multitenant.go:1186","component":"MultiTenantAlertmanager","file":"/data/metrics-activity.log","level":"warn","msg":"ignoring unexpected file while scanning local alertmanager configs","ts":"2022-04-27T22:05:37.499007346Z"}` Break: configuration
- [x] Rename `memcached` service to `memcached-chunks` or something more specific, reflecting its intended usage. Break: configuration, possibly causing cache unavailable during upgrade. 
- [x] Remove dependency on bitnami memcached and introduce our own version like loki-distributed (#1315 ). Set up proper ServiceMonitor CRDs. Break: less values supported for memcached - not full bitnami anymore

dimitarvdimitrov:
> Move activity log somewhere from /data, as it causes warning in Alertmanager

The problem comes from the fast that the activity log and the AM temporary directory are interfering, right? Changing the activity log is a breaking change, but we can change the `alertmanager.data_dir`. The `data_dir` directory does not need to be persisted during restarts ([doc](https://grafana.com/docs/enterprise-metrics/v2.0.x/config/reference/#alertmanager)), I think it's ok to change the value of it and not consider this a breaking change.

krajorama:
> > Move activity log somewhere from /data, as it causes warning in Alertmanager
> 
> The problem comes from the fast that the activity log and the AM temporary directory are interfering, right? Changing the activity log is a breaking change, but we can change the `alertmanager.data_dir`. The `data_dir` directory does not need to be persisted during restarts ([doc](https://grafana.com/docs/enterprise-metrics/v2.0.x/config/reference/#alertmanager)), I think it's ok to change the value of it and not consider this a breaking change.

Quite possibly, but I haven't had time to test it out

krajorama:
this is now tracker internally, closing. The items regarding memcached will be included in 3.0 release of the chart.

dimitarvdimitrov:
[...]

Issue #1217, opened 2022-04-11
[mimir-distributed] Make it possible to run mimir-distributed on OpenShift 


patrickabrennan:
Memcache needs a service account so that the mimir-distributed helm chart can be installed on non-vanilla k8s distributions like OpenShift. Without this service account, the memcached deployment used in the Helm chart doesn't conform to the security requirements of OpenShift. 

Blocked on https://github.com/grafana/helm-charts/issues/1315. 

------- more detail --------

The Bitnami 5.5.2 memcache chart does not provide for a service account name. This means that it will use the "default" service account in the namespace it is deployed. This creates two problems. 1) It is not a K8s best practice - all applications should have its own service account and 2) The memcache chart from Bitnami has a `runAsUser: 1001` setting. While this is normally ok for some k8s distributions like GKE which uses Pod Security Policy (PSP), this does not work for OpenShift which uses Security Context Constraints (SCC).  In OpenShift the default SCC is `restricted` and it does not allow the use of a `runAsUser: 1001`, it uses a range which is randomly generated when a namespace (what OpenShift calls a project) is created. To use `runAsUser: 1001` one has to either use an SCC such as `anyuid` which is one of the 8 built in SCC's or create ones own SCC. In order to use an SCC other than the default of `restricted` one needs a service account for the application. I would also note that the newer versions of the Bitmani memcache chart has a setting for a service account. While using the default service account is possible (connect the RBAC permission with the proper SCC) it is generally not recommended.

09jvilla:
Draft PR:
https://github.com/grafana/helm-charts/pull/1329

PR #1329, opened 2022-05-02
[mimir-distributed] OpenShift support

krajorama:
Solves #1217 Related to #1309 #1315

### DIFF
--- a/charts/mimir-distributed/templates/rolebinding.yaml
+++ b/charts/mimir-distributed/templates/rolebinding.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.pspEnabled }}
+{{- if or .Values.rbac.pspEnabled .Values.rbac.sccEnabled }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
--- a/charts/mimir-distributed/templates/rolebinding.yaml
+++ b/charts/mimir-distributed/templates/rolebinding.yaml
@@ -12,4 +12,5 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: {{ template "mimir.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace }}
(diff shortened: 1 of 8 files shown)


charts/mimir-distributed/templates/rolebinding.yaml
```diff
@@ -12,4 +12,5 @@ roleRef:
 subjects:
```
Skaronator, line 12: Can't comment on line 1 but the if condition seems to be missing the `.Values.rbac.sccEnabled`?

```
{{- if or .Values.rbac.pspEnabled .Values.rbac.sccEnabled }}
```

charts/mimir-distributed/values.yaml
```diff
@@ -32,7 +32,7 @@ global:
 
 serviceAccount:
   create: true
-  name:
+  name: &serviceAccountName "mimir-service-account"
```
Skaronator, line 35: The only thing I dislike is that the `nameOverride` and `fullnameOverride` logic doesn't work for the serviceAccount.name anymore.

Furthermore, when you set `serviceAccount.name` in your values, it doesn't update the `memcached.serviceAccount.name` YAML anchors below. Since the anchor doesn't exist in my values.yaml.

But I don't have a better solution, since this is a limitation of helm when you use sub-charts.

krajorama: Yeah, I'm not proud of the anchor thing :( , but that was the only thing I saw at the moment. 
I talked to my PM and and they are ok with having this feature in the next major version where we can break more things. If you can manage with this experimental chart in the mean time (couple of months) I'd rather defer merging it.

zanhsieh:
mimir-distributed has been [moved](https://github.com/grafana/mimir/tree/main/operations/helm/charts/mimir-distributed).

zanhsieh:
https://github.com/grafana/mimir/pull/2274

PR #1340, opened 2022-05-04
[loki-simple-scalable] add initContainers

Grounz:
Hi,
I submit this pr beacause when we want deploy loki simple-scalable helm charts we use boltdb-shipper.

When read and write loki pod start they have a var/loki/boltdb-shipper-active: permission denied errir and we cannot fix that without use an initContainer which do chown 10001:10001 /var/loki.

Loki-simple-scalable helm chart not containers initContainers in statefulset templates for read and write pods.

So i have add the possibility to add initContainers for read and for write pods.

in values.yaml:

```
write:
...
  initContainers: []
read:
...
  initContainers: []
```

Signed-off-by: lenglet-k <lenglet-k@mgdis.fr>


zanhsieh:
@Grounz 
Please bump the Chart.yaml to latest and sign license/cla. Thank you.

charts/loki-simple-scalable/templates/read/statefulset-read.yaml
```diff
@@ -38,6 +38,8 @@ spec:
       {{- include "loki.readPriorityClassName" . | nindent 6 }}
       securityContext:
         {{- toYaml .Values.loki.podSecurityContext | nindent 8 }}
+      initContainers:
```
trevorwhitney, line 41: WDYT of wrapping this in a `with` block so the key is only included when init containers are provided?

trevorwhitney (changes requested):
Would prefer a `with` block so key is excluded rather than included with empty array. Also, please sign CLA.

zanhsieh:
@trevorwhitney @joe-elliott 
Documentation CI seems having some problem.

zanhsieh:
@Grounz 
Since loki and loki-simple-scalable moved to https://github.com/grafana/loki/tree/main/production/helm/loki . Please PR over there.

PR #1346, opened 2022-05-05
[mimir-openshift-experimental] Copy of branch krajo/mimir-update-memcached to publish for testing

krajorama:
See #1329 


Logiraptor (approved):
Approving to unblock. I haven't reviewed deeply since this is just experimental

PR #1347, opened 2022-05-05
[loki-distributed] Ensure querier is mounting data directory

simonswine:
Otherwise the querier will fail with writing to /var/loki


vlad-diachenko (approved):
LGTM

kavirajk (approved):

PR #1333, opened 2022-05-03
[grafana] Add more support for templates in values

vtavernier:
This PR adds more possibilities for injecting templated values in the Grafana chart, when used as a dependency. This follows previous PRs such as #1167.

This enabled templates in the following parameters:
* extraConfigmapMounts
* affinity
* imageRenderer.affinity
* admin.existingSecret
* persistence.subPath
* persistence.existingClaim
* image.pullSecrets
* imageRenderer.image.pullSecrets
* serviceAccount.annotations

PR #1351, opened 2022-05-06
fix(grafana): fix template issue with existingClaim

emaincourt:
Hi,

https://github.com/grafana/helm-charts/pull/1333 seems to have introduced a templating issue when setting `persistence.enabled` to `true` and not passing an `existingClaim` value. `tpl` expects a string to parse but gets an empty interface. Please find below the error message:

```
Error: template: grafana/templates/deployment.yaml:49:10: executing "grafana/templates/deployment.yaml" at <include "grafana.pod" .>: error calling include: template: grafana/templates/_pod.tpl:659:20: executing "grafana.pod" at <tpl>: wrong number of args for tpl: want 2 got 1

Use --debug flag to render out invalid YAML
```

Defining an empty string as a default value fixes the issue.

Thanks in advance for reviewing

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -656,7 +656,7 @@ volumes:
 {{- if and .Values.persistence.enabled (eq .Values.persistence.type "pvc") }}
   - name: storage
     persistentVolumeClaim:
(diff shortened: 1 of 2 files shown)


emaincourt:
Hi @vtavernier, thanks for addind the tests. I can see that you already included a fix to your PR so I'm closing this one.

PR #1361, opened 2022-05-09
[loki-simple-scalable] Use policy/v1 apiVersion for PodDisruptionBudget when possible

thar:
The `PodDisruptionBudget` resources are using the deprecated `policy/v1beta1` apiVersion.

This commit replaces it by the `policy/v1` apiVersion when it is available.


parera10 (commented):
You should bump to the lastest version on Chart.yaml.

zanhsieh (approved):

zanhsieh (approved):

Issue #1358, opened 2022-05-08
[Grafana] Persistence existingClaim ignored


greg-mcnamara:
I'm deploying grafana with a custom values file. The persistence settings are:

```
persistence:
  type: pvc
  enabled: true
  existingClaim: pvc-grafana
```

The k8s pod remains in state "Pending" and when I run kubectl describe pod I see the following in events:
```
Warning  FailedScheduling  2m4s  fargate-scheduler  Pod not supported on Fargate: volumes not supported: storage not supported because: persistentvolumeclaim "grafana" not found
```

And in the pod description I see:
```
Volumes:
  config:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      grafana
    Optional:  false
  storage:
    Type:       PersistentVolumeClaim (a reference to a PersistentVolumeClaim in the same namespace)
    ClaimName:  grafana
    ReadOnly:   false
```

Does anyone know why it's using the wrong claim name? The existing claim pvc referenced is bound and doesn't appear to have any problems. I have previously deployed the same configuration on another k8s cluster with no problem.

greg-mcnamara:
Just found #1350 so it looks like a bug in the latest chart release

PR #1362, opened 2022-05-09
[Grafana] Fix existing volume claim usage issue

### LINKED ISSUE #1350: Version 6.29.0 of the grafana chart has tpl usage problems
While trying to deploy the latest chart it errors with:
```
Error: Failed to render chart: exit status 1: Error: template: grafana/templates/deployment.yaml:49:10: executing "grafana/templates/deployment.yaml" at <include "grafana.pod" .>: error calling include: template: grafana/templates/_pod.tpl:659:20: executing "grafana.pod" at <tpl>: wrong number of args for tpl: want 2 got 1
```

The line in question is:
```
      claimName: {{ tpl (.Values.persistence.existingClaim .) | default (include "grafana.fullname" .) }}
```
And it should be:
```
      claimName: {{ (tpl .Values.persistence.existingClaim .) | default (include "grafana.fullname" .) }}
```

It also looks like line 426 in the same file has a similar problem:
```
        subPath: {{ tpl (.subPath $root) | default "" }}
```
```
        subPath: {{ (tpl .subPath $root) | default "" }}
```

kallaics:
The issue is still the order inside 'default' section.

Usage:  default "<default value"> <item>
Source: https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

The wrong line: 

```
claimName: {{ tpl (default .Values.persistence.existingClaim (include "grafana.fullname" .)) . }}
```

The fixed line based on the guide:

```
claimName: {{ tpl (default (include "grafana.fullname" .) .Values.persistence.existingClaim) . }}
```

Tested the result will be as expected:

Default claim name: grafana 
Existing claim name: grafana-pvc
So look like the expected result provided after the change.

```
volumes:
        - name: config
          configMap:
            name: grafana
        - name: storage
          persistentVolumeClaim:
            claimName: grafana-pvc
```

fixes #1350 #1358

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.29.1
+version: 6.29.2
 appVersion: 8.5.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -656,7 +656,7 @@ volumes:
 {{- if and .Values.persistence.enabled (eq .Values.persistence.type "pvc") }}
   - name: storage
     persistentVolumeClaim:
-      claimName: {{ tpl (default .Values.persistence.existingClaim (include "grafana.fullname" .)) . }}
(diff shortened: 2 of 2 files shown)


kallaics:
Same as #1360

PR #1372, opened 2022-05-12
[loki] add service per replica capability

ronleb:
Add ability to create a dedicated service for each replica of loki

We have a setup with multiple instances of loki for high availability in case a problem occurs on a node running one instance of loki. In that regard we need the ability when investigating a problem to target a specific instance of loki that was not impacted by a problem


ronleb:
After further analysis it appears the loki headless service fulfills our need so I will close this pull request

PR #1374, opened 2022-05-13
[loki-stack] support configurable datasource urls

### LINKED ISSUE #1373: [loki-stack] Configurable datasource url
The current datasource template assumes how the url should be constructed, and I think this should be switched to a `default` instead.

cornfeedhobo:
This PR seeks to support configurable datasource URLs.

It'd be great if `isDefault` and `url` could be moved under a `datasource` key, but I assume that'd be too much of a breaking change.

Fixes #1373


zanhsieh:
@cornfeedhobo 
Could you bump the Chart.yaml version to latest and sign DCO please?

zanhsieh (approved):

PR #1376, opened 2022-05-16
[tempo] Upgrade to 1.4.1

joe-elliott:


adrianliechti:
could you consider to update the loki-query image too?

i ask because running this chart on arm64 does currently not work (since the 1.0.1 image is only built for amd64) - if i change the tempo-query image to 1.4.0/1.4.1 -> all good

joe-elliott:
Thanks for the heads up @adrianliechti, but I think this happens automatically. If I install the current chart (which installs Tempo 1.4.0) it uses the 1.4.0 `tempo-query` image:

```
$ kc describe po
  ...
  tempo-query:
    Image:         docker.io/grafana/tempo-query:1.4.0
  ...
```

Also, for anyone else reading this thread, tempo-query is only required for older versions of Grafana:

https://grafana.com/docs/tempo/latest/configuration/querying/

joe-elliott:
Oh dang, I'm confused. I was looking at the `tempo-distributed` helm chart. Pushing the change now! Thank you :pray: for the heads up!

mdisibio (approved):

adrianliechti:
that was extremely fast. thank you very much! that just allows me to remove a hack from today :)

PR #1384, opened 2022-05-18
[loki-stack] Promtail fails to push logs to loki after upgrade

### LINKED ISSUE #1383: [loki-stack] Promtail fails to push logs to loki after upgrade
After upgrading loki-stack from 2.4.1 to 2.6.4 the promtail daemonsets start to fail pushing logs to loki with the following error
```
level=warn ts=2022-05-18T03:37:09.360423983Z caller=client.go:349 component=client host=myloki:3100 msg="error sending batch, will retry" status=-1 error="Post \"http://myloki:3100/loki/api/v1/push\": dial tcp: lookup myloki on 10.100.0.10:53: no such host"
```

This seems to be related to the [following line](https://github.com/grafana/helm-charts/blob/9de374802e38d88c119f2d61bfa6c87856b84f51/charts/loki-stack/values.yaml#L8), which specifies
```yaml
promtail
  config:
    lokiAddress: http://{{ .Release.Name }}:3100/loki/api/v1/push
```
as the default value for `lokiAddress` whereas it should be rather (mind the `-loki`)
```yaml
promtail:
  config:
    lokiAddress: http://{{ .Release.Name }}-loki:3100/loki/api/v1/push
```
to reference the k8s service for loki that gets created by the loki helm chart, see https://github.com/grafana/helm-charts/blob/9bce229569edc2035c39249ae818f1003331b821/charts/loki/templates/_helpers.tpl#L22

Current workaround is to specify the latter config in an override.yml when installing/upgrading via helm.

hatzlj:
This PR fixes #1383, adjusting the `promtail.config.lokiAddress` default value in values.yaml of `loki-stack` to follow the convention for the loki service name created by the loki chart (which defaults to `{{ .Release.Name }}-{{ .Chart.Name}}`).


charts/loki-stack/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: loki-stack
-version: 2.6.4
+version: 2.6.5
```
zanhsieh, line 3: This one is 2.6.6

zanhsieh (changes requested):

charts/loki-stack/Chart.yaml
hatzlj: fixed in b4099bcaa1d21729f9e8e3c15ba801feb3088d6f

zanhsieh:
@hatzlj 
I think somewhere the original assumption is not correct:
https://github.com/grafana/helm-charts/runs/6967284513?check_suite_focus=true#step:9:405

level=warn ts=2022-06-20T12:58:31.20618349Z caller=client.go:349 component=client host=loki-stack-lzn564ojr1-loki:3100 msg="error sending batch, will retry" status=-1 error="Post \"http://loki-stack-lzn564ojr1-loki:3100/loki/api/v1/push\": dial tcp: lookup loki-stack-lzn564ojr1-loki on 10.96.0.10:53: server misbehaving"

zanhsieh:
@hatzlj 
This one is fixed. Please rebase and re-PR if the problem still exists.

PR #1390, opened 2022-05-19
[loki-distributed] Add compactor to the memberlist

ludydoo:
As stated in the documentation https://grafana.com/docs/loki/latest/fundamentals/architecture/rings/?src=blog&camp=timeshift_53 the compactor should be in the ring


ludydoo:
@unguiculus @Whyeasy not sure what is wrong here, but when I try to sign the CLA, it says that the item does not exist..

Issue #1110, opened 2022-03-17
[tempo-distributed] configmap should allow values for ingester's ring


AlexDHoffer:
The chart's default value for ingester shows:

````
ingester:
    lifecycler:
      ring:
        replication_factor: 1
        kvstore:
          store: memberlist
````

It would be nice to have some of these configurable, e.g.:

````
ingester:
    lifecycler:
      ring:
        replication_factor: {{ .Values.ingester.config.replication_factor }}
        kvstore:
          store: {{ .Values.ingester.config.kvstore }}
````

As it stands I'm overriding the entire configmap so I can increase the replication factor for a highly available ingester.

PR #1396, opened 2022-05-19
[tempo-distributed] Allow configurable replication factor for the ingester

AlexDHoffer:
Signed-off-by: AlexDHoffer <alexdchoffer@gmail.com>

Resolves https://github.com/grafana/helm-charts/issues/1110


joe-elliott:
Yup, this we need this update. I'm tempted to default to 3 which is the only RF I'd consider appropriate for production.

WDYT?

joe-elliott:
Yup, let's do both. Increase ingesters and RF to 3 by default. If someone is using tempo-distributed they are likely looking for a distributed, fault tolerant system and the defaults should give them that.

Thanks for these fixes!

AlexDHoffer:
@joe-elliott Hey Joe, this is ready to go but there's a hard pod anti-affinity on the Ingester that is causing issues with CI/CD because there aren't enough nodes to satisfy the requirement.

Shall I turn down that affinity to ``preferred`` rather than ``required``?

joe-elliott:
Yeah, after some thought let's try setting the pod anti-affinifty to preferred for `kubernetes.io/hostname`. If that doesn't work we can just go back to defaulting to RF1 and 1 ingester.

AlexDHoffer:
@joe-elliott all set here

joe-elliott (approved):
Great stuff. Thanks Alex!

PR #1407, opened 2022-05-24
[mimir-distributed] Feature toggle for topologySpreadConstraints

daxroc:
Enhancement to allow setting of topologySpreadConstraints along with a feature flag to enable/disable

### DIFF
--- a/charts/mimir-distributed/templates/overrides-exporter/overrides-exporter-dep.yaml
+++ b/charts/mimir-distributed/templates/overrides-exporter/overrides-exporter-dep.yaml
@@ -92,7 +92,7 @@ spec:
         {{- with .Values.overrides_exporter.extraContainers }}
         {{ toYaml . | nindent 8 }}
         {{- end }}
-      {{if .Values.overrides_exporter.topologySpreadEnabled }}
+      {{ if .Values.overrides_exporter.topologySpreadConstraints }}
       topologySpreadConstraints:
         {{- toYaml .Values.overrides.topologySpreadConstraints | nindent 8 }}
       {{- end}}
--- a/charts/mimir-distributed/templates/overrides-exporter/overrides-exporter-dep.yaml
+++ b/charts/mimir-distributed/templates/overrides-exporter/overrides-exporter-dep.yaml
@@ -92,6 +92,10 @@ spec:
         {{- with .Values.overrides_exporter.extraContainers }}
         {{ toYaml . | nindent 8 }}
         {{- end }}
+      {{if .Values.overrides_exporter.topologySpreadEnabled }}
+      topologySpreadConstraints:
+        {{- toYaml .Values.overrides.topologySpreadConstraints | nindent 8 }}
+      {{- end}}
       nodeSelector:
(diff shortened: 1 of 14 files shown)


colega:
Thank you for the contribution, this looks good to me, although I have no experience at all in maintaining/using the Helm chart so I'll ask someone else to approve/merge it.

If any, I have one question: would it make sense to render the `topologySpreadConstraints` whenever it's not empty, and remove the `topologySpreadEnabled` flag completely? As I said, I have no experience in using Helm charts and I'm not sure if this is a common pattern.

colega:
I meant doing something like:
```
      {{if .Values.alertmanager.topologySpreadConstraints }}
      topologySpreadConstraints:
        {{- toYaml .Values.alertmanager.topologySpreadConstraints | nindent 8 }}
      {{- end}}
```

(Checking the slice itself instead of the bool var, this way the key will be rendered only if there's something to render in it).

This would be backwards compatible in a way that if your k8s doesn't support it, you just leave it empty and it's not rendered.

daxroc:
@colega made those changes suggested, appreciate the feedback!

charts/mimir-distributed/templates/overrides-exporter/overrides-exporter-dep.yaml
```diff
@@ -92,6 +92,10 @@ spec:
         {{- with .Values.overrides_exporter.extraContainers }}
         {{ toYaml . | nindent 8 }}
         {{- end }}
+      {{ if .Values.overrides_exporter.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml .Values.overrides.topologySpreadConstraints | nindent 8 }}
```
colega, line 97:
```suggestion
      {{ if .Values.overrides_exporter.topologySpreadConstraints }}
      topologySpreadConstraints:
        {{- toYaml .Values.overrides_exporter.topologySpreadConstraints | nindent 8 }}
```

krajorama (commented):
LGTM after fix proposed by @colega . The one thing I wonder about is that we'll have Mimir 2.1 launch soon and currently our versioning is very similar (we'll diverge over the next couple of months). So since this change is small and backwards compatible, I'd rather we just used version 2.0.15 to avoid confusion around the versioning.

colega:
@krajorama if we're not following semver in this repo, then I'd say we should document that in the CHANGELOG header, wdyt? 

I wasn't familiar with the versioning scheme, that's why I suggested bumping the minor.

krajorama:
> @krajorama if we're not following semver in this repo, then I'd say we should document that in the CHANGELOG header, wdyt?
> 
> I wasn't familiar with the versioning scheme, that's why I suggested bumping the minor.

You're right in general, I'm just weighing the impact of keeping semver vs confusing people in this particular case. The other option to avoid confusion is to hold off merging this PR. But that could be weeks. Or we do merge and accept the possible confusion.

daxroc:
Any update on the versioning ?

zanhsieh:
Mimir-distributed repo moved. close this PR.

PR #1410, opened 2022-05-24
[grafana] Upgrade PodDisruptionBudget apiVersion

besha100:
PodDisruptionBudget apiVersion policy/v1beta1 is deprecated in v1.21+. and will be unavailable in the next Kubernetes version. Nothing else needs to be changed in the yaml specs

### DIFF
--- a/charts/grafana/templates/_helpers.tpl
+++ b/charts/grafana/templates/_helpers.tpl
@@ -142,6 +142,17 @@ Return the appropriate apiVersion for ingress.
 {{- end -}}
 
 {{/*
+Return the appropriate apiVersion for podsecuritypolicy and PodDisruptionBudget.
+*/}}
+{{- define "grafana.policy.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">= 1.21-1" .Capabilities.KubeVersion.Version) -}}
+    {{- print "policy/v1" -}}
+  {{- else -}}
+    {{- print "policy/v1beta1 -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
 Return if ingress is stable.
 */}}
 {{- define "grafana.ingress.isStable" -}}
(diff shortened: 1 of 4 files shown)


charts/grafana/templates/poddisruptionbudget.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.podDisruptionBudget }}
```
zanhsieh, line 1: Could you move this to templates/_helpers.tpl and define like this?

```
{{/*
Return the appropriate apiVersion for ingress.
*/}}
{{- define "grafana.psp.apiVersion" -}}
  {{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">= 1.21-1" .Capabilities.KubeVersion.Version) -}}
    {{- print "policy/v1" -}}
  {{- else -}}
    {{- print "policy/v1beta1 -}}
  {{- end -}}
{{- end -}}
```
and use as:
kind: {{ template "grafana.psp.apiVersion" . }}

charts/grafana/templates/podsecuritypolicy.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.rbac.pspEnabled }}
-apiVersion: policy/v1beta1
+apiVersion: policy/v1
```
zanhsieh, line 2: See above.

zanhsieh (changes requested):

besha100:
Thank you! It's done. Can you review @zanhsieh

zanhsieh:
@besha100 
I think you need to bump the chart version to latest, as the lint test failed.

charts/grafana/templates/_helpers.tpl
```diff
@@ -141,6 +141,17 @@ Return the appropriate apiVersion for ingress.
...
 {{- end -}}
 
+{{/*
+Return the appropriate apiVersion for podsecuritypolicy and PodDisruptionBudget.
+*/}}
+{{- define "grafana.policy.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">= 1.21-1" .Capabilities.KubeVersion.Version) -}}
+    {{- print "policy/v1" -}}
+  {{- else -}}
+    {{- print "policy/v1beta1 -}}
```
zanhsieh, line 151: Missing close double quote here. Should be:
```
{{- print "policy/v1beta1" -}}
```

zanhsieh (changes requested):

charts/grafana/templates/_helpers.tpl
besha100: @zanhsieh Fixed! Thank you :)

besha100:
@zanhsieh now suceeded. Thank you

zanhsieh (approved):

PR #1416, opened 2022-05-26
Make workflows in this repository reusable from other repositories

krajorama:
This PR enables storing helm charts next to their product (like we do with jsonnet), but still allows to reuse common github actions for linting and testing. Out of scope: release (will be another PR).


pracucci (approved):
LGTM!

PR #1417, opened 2022-05-26
Fixed default chunks and rules directories in loki-simple-scalable

### LINKED ISSUE #1415: chunksDirectory and rulesDirectory variable name is error!
when I use create loki-simle-scalable use loki.storage.type:local then loki ConfigMap chunks_directory and rules_directory is null!

values.yaml:
```yaml
loki:
    storage:
    bucketNames:
      chunks: chunks
      ruler: ruler
      admin: admin
    type: s3
    //......
    local:
      chunks_directory: /var/loki/chunks  
      rules_directory: /var/loki/rules
```

_helpers.tpl:
```yaml
{{- with .Values.loki.storage.local }}
filesystem:
  chunks_directory: {{ .chunksDirectory }}  //the values.yaml loki.storage.local.chunksDirectory and loki.storage.local.rulesDirectory is not exists!
  rules_directory: {{ .rulesDirectory }}
{{- end -}}
```

Artemigos:
Fixes #1415 .
I'm not sure about version changes - is that something I should do?

### DIFF
--- a/charts/loki-simple-scalable/values.yaml
+++ b/charts/loki-simple-scalable/values.yaml
@@ -142,8 +142,8 @@ loki:
       requestTimeout: "0s"
       enableHttp2: true
     local:
-      chunks_directory: /var/loki/chunks
-      rules_directory: /var/loki/rules
+      chunksDirectory: /var/loki/chunks
+      rulesDirectory: /var/loki/rules
 
   # -- Check https://grafana.com/docs/loki/latest/configuration/#schema_config for more info on how to configure schemas
   schemaConfig: {}


charts/loki-simple-scalable/values.yaml
```diff
@@ -142,8 +142,8 @@ loki:
       requestTimeout: "0s"
       enableHttp2: true
     local:
-      chunks_directory: /var/loki/chunks
-      rules_directory: /var/loki/rules
+      chunksDirectory: /var/loki/chunks
```
trevorwhitney, line 145: this is very nit-picky, so I apologize, but I'd rather keep the key here `chunks_directory` and `rules_directory` as that more closely matches what you would pass to loki itself. I would prefer fixing this by fixing the helper function instead, WDYT?

trevorwhitney (changes requested):

trevorwhitney:
WDYT of the approach in https://github.com/grafana/helm-charts/pull/1420 instead? I like whenever possible to keep the key names in the `values.yaml` as similar as possible to those in the loki config file, which are all snake case.

charts/loki-simple-scalable/values.yaml
Artemigos: I chose this way around because if you look at the `gcs` or `s3` configs right above it, they use camel case. Also in the entire `values.yaml` the only other example of a snake case key I found is `loki.commonConfig.path_prefix`. That being said, I leave the decision up to you - just tell me if I should change this :)

PR #1418, opened 2022-05-26
Make release workflow reusable from other repos

krajorama:
Enables our other repositories to use the same action and github pages to publish helm charts.


trevorwhitney (approved):
LGTM!

krajorama:
This turned out to create the release in the other repo, but we'd want to avoid that for Mimir - to not pollute the release namespace there.

PR #1419, opened 2022-05-26
[loki] supported memberlist

vlad-diachenko:
enabled memberlist by default and added support of memberlist( kubernetes service and exposed port on the pod)

### DIFF
--- a/charts/loki/templates/_helpers.tpl
+++ b/charts/loki/templates/_helpers.tpl
@@ -73,3 +73,13 @@ extensions/v1
 {{- end }}
 {{- end -}}
 
+
+{{/*
+Common labels
+*/}}
+{{- define "loki.labels" -}}
+app: {{ template "loki.name" . }}
+chart: {{ template "loki.chart" . }}
+release: {{ .Release.Name }}
+heritage: {{ .Release.Service }}
+{{- end }}
--- a/charts/loki/templates/service-memberlist.yaml
+++ b/charts/loki/templates/service-memberlist.yaml
@@ -9,6 +9,7 @@ metadata:
 spec:
   type: ClusterIP
   clusterIP: None
+  publishNotReadyAddresses: true
   ports:
     - name: http
       port: {{ .Values.config.memberlist.bind_port | default 7946 }}
--- a/charts/loki/templates/service-memberlist.yaml
+++ b/charts/loki/templates/service-memberlist.yaml
@@ -1,21 +1,20 @@
+{{- if .Values.config.memberlist -}}
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "loki.fullname" . }}-headless
+  name: {{ include "loki.fullname" . }}-memberlist
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "loki.labels" . | nindent 4 }}
-    variant: headless
 spec:
+  type: ClusterIP
   clusterIP: None
   ports:
-    - port: {{ .Values.service.port }}
+    - name: http
+      port: {{ .Values.config.memberlist.bind_port | default 7946 }}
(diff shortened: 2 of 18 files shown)


charts/loki/templates/_helpers.tpl
```diff
@@ -73,3 +73,13 @@ extensions/v1
...
 
+
+{{/*
+Common labels
+*/}}
+{{- define "loki.labels" -}}
+app: {{ template "loki.name" . }}
+chart: {{ template "loki.chart" . }}
+release: {{ .Release.Name }}
+heritage: {{ .Release.Service }}
```
trevorwhitney, line 84: there's some discrepancy here with the label names we use across charts and should probably try to be consistent. For example, in `loki-simple-scalable` and `loki-distributed` we us `helm.sh/chart` instead of just `helm`, and prefix a lot of other labels with `app.kubernetes.io`.

trevorwhitney: I do realize this was just a refactor to reduce duplication, but it made be aware of this at the same time.

charts/loki/templates/service-memberlist.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.config.memberlist -}}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "loki.fullname" . }}-memberlist
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "loki.labels" . | nindent 4 }}
+spec:
```
trevorwhitney, line 9: do we want to include `publishNotReadyAddresses` here, to fix the chicken and egg issue when the ring is first being populated? or maybe we don't care?

charts/loki/values.yaml
```diff
@@ -49,6 +49,13 @@ tracing:
 config:
   # existingSecret:
   auth_enabled: false
+
+  memberlist:
+    bind_port: 7946
```
trevorwhitney, line 54: I think this is the default, so we shouldn't need it

trevorwhitney (changes requested):
just a couple small things, and a question about what to do re: annotations?

charts/loki/templates/service-memberlist.yaml
alexandre1984rj: @trevorwhitney I had to do it manually for the query-frontend service to avoid the error below because the chart does not give the chance to add annotations or the publishNotReadyAddresses parameter.
```
level=error ts=2022-05-28T01:05:41.798975654Z caller=frontend_processor.go:69 msg="error processing requests" address=10.20.3.238:9095 err="rpc error: code = Canceled desc = context canceled"
```

vlad-diachenko: yes, it's a great idea... added

charts/loki/templates/_helpers.tpl
vlad-diachenko: in case we change it, it will require redeploying the whole cluster to apply the changes, so I left it unchanged. 
**let's do it in a separate pr in future... there are many things to be done before our charts become identical...
what do you think?**

I struggle more with the structure of values files in these 3 charts.  `loki`, `loki-distributed` and `loki-simple-scalable` use different ways and different sets of properties/values to create Loki's config ;)

charts/loki/values.yaml
vlad-diachenko: removed. thanks

charts/loki/templates/_helpers.tpl
trevorwhitney: :+1: to doing in a separate PR

trevorwhitney (approved):
LGTM

PR #1424, opened 2022-05-30
[promtail] add `instance` labels

dungdm93:
Current, we have `app` label that refer to k8s's `app.kubernetes.io/name` label or helm's `app` legacy label.

In the cases there are multiple deployments/releases in the same namespace, it's very helpful if there is `instance` label that used to distinguish among them.

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.5.0
-version: 5.0.0
+version: 5.1.0
 home: https://grafana.com/loki
 sources:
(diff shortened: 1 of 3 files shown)


dungdm93:
cc @trevorwhitney

trevorwhitney:
Hmm, I worry about getting too opinionated around things like this. What is the argument for having this in the base `values.yaml` vs requiring users to specify any different labels they might need in their own `values.yaml`. I err on the side on keeping the default configs simpler.

dungdm93:
@trevorwhitney agreed that default configs should be simple.
However, `instance` should be supported along side with `app` and `component` because all of them are in [k8s recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

trevorwhitney (approved):
That argument works for me, LGTM!

PR #1425, opened 2022-05-30
[charts/promtail] support multiple clients via `clientConfigs`

dungdm93:
See [#1214](https://github.com/grafana/helm-charts/issues/1214#issuecomment-1136147358)


trevorwhitney (approved):
Thanks, I like this approach! LGTM

damyan:
I guess https://github.com/grafana/helm-charts/blob/main/charts/promtail/README.md.gotmpl#L155-L156 should have been adjusted as well, shouldn't they?

PR #1431, opened 2022-05-31
Reusable workflow to release a chart from a different repository

krajorama:
This workflow builds a chart in a source repository and releases it
to the helm-charts repository. It tags the source repository. It creates
a PR to update the helm repo index.yaml in the helm-charts repository.

Eventually we'd want the workflow to just commit the change to the index.yaml for helm without a PR - but let's test things e2e first.

**Note** this workflow is only triggered by other workflows, not by itself.

### DIFF
--- a/.github/workflows/update-helm-repo.yaml
+++ b/.github/workflows/update-helm-repo.yaml
@@ -1,0 +1,185 @@
+name: Release a chart
+
+on:
+  workflow_call:
+    inputs:
+      charts_dir:
+        description: path to the helm charts
+        default: charts
+        required: false
+        type: string
+      cr_configfile:
+        description: location of the chart releaser (cr) config file
+        default: cr.yaml
+        required: false
+        type: string
+      ct_configfile:
+        description: location of the chart tester (ct) config file
+        default: ct.yaml
+        required: false
+        type: string
+    secrets:
+      helm_repo_token:
+        description: GitHub api token to use against the helm-charts repository
+        required: true
+      gpg_key_base64:
+        description: GPG key in base64 for signing helm package with
+        required: true
+
+env:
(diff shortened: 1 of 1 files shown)


Logiraptor (approved):

.github/workflows/update-helm-repo.yaml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Push release tag on origin
+        run: |
+          cd source
+          echo "Pushing tag ${{ steps.parse-chart.outputs.tagname }}"
+          git push origin "${{ steps.parse-chart.outputs.tagname }}"
+
+      - name: Update helm repo index.yaml
+        run: |
+          cd helm-charts
+          "${CR_TOOL_PATH}/cr" index --config "${CR_CONFIGFILE}" --token "${{ secrets.helm_repo_token }}" --index-path "${CR_INDEX_PATH}" --package-path "${CR_PACKAGE_PATH}" --pr
```
Logiraptor, line 185: Just out of curiosity, is there a reason not to use the chart-releaser-action that we use elsewhere in this repo?

https://github.com/grafana/helm-charts/blob/main/.github/workflows/release.yaml#L48

Logiraptor: @krajorama

krajorama: I've tried to run it in the mimir repo, but didn't work as it created the release in mimir repo and tried to update its gh-pages (which would put the index.yaml in the wrong place).

On the other hand, I'm realizing now that I ended up checking out the helm-charts repo anyway, so potentially we could copy the code from mimir to the new repo, stage the files (for [git diff](https://github.com/helm/chart-releaser-action/blob/a3454e46a6f5ac4811069a381e646961dda2e1bf/cr.sh#L250) to work) and run the chart releaser.

PR #1432, opened 2022-05-31
[grafana] supported the case when only plugins are provisioned

vlad-diachenko:
**Issue:**
If neither provisioning of dashboards nor provisioning of datasources is enabled, but we need to provision only a plugin, we were not able to do it because the permissions to read configMaps and secrets were granted only if datasources or dashboards provisioning is enabled. 
Previously, if we deploy Grafana with only plugin provisioning enabled, the sidecar started reporting the errors that it does have access to query configMaps and secrets.

**Fix:**
added permissions to clusterRole and role to access "configmaps" and "secrets" if any resource(plugins or datasources or dashboards) provisioning is enabled. 


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.29.5
+version: 6.29.6
```
zanhsieh, line 3: now 6.29.7

zanhsieh (changes requested):

charts/grafana/Chart.yaml
vlad-diachenko: done

zanhsieh (approved):

PR #1435, opened 2022-06-02
Fix change detection in helm repo update action

krajorama:
Change detection didn't work on main branch because target (main) and HEAD
were the same.
Add detection of latest tag to compare to.
Also remove running change detection twice and place changed chart path
on output of the first job.


dimitarvdimitrov (approved):
LGTM

PR #1438, opened 2022-06-02
Workaround for bitnami releasing mimir with old memcached

krajorama:
Bitnami has removed packages older than 6 months from their helm
repo , breaking the build. We're in the middle of a release, enable
workaround to unblock us. Eventually we want to get rid of the bitnami
chart for memcached anyway, see issue 1315

See:
https://github.com/bitnami/charts/pull/10530
https://github.com/bitnami/charts/issues/10539
https://github.com/grafana/helm-charts/issues/1315


dimitarvdimitrov (approved):
LGTM

PR #1441, opened 2022-06-03
[loki-canary] Adding flexibility to add pod labels to loki-canary daemon set

praveenjindal62:
We are using loki-canary charts and wanted to manage pods using some custom labels on pods. Loki-Canary doesn't provide fexibility yet to use custom podLabels on Daemonset loki-canary. Added .Values.podLabels that can be applied to loki-canary daemonset.

### DIFF
--- a/charts/loki-canary/values.yaml
+++ b/charts/loki-canary/values.yaml
@@ -21,6 +21,9 @@ priorityClassName: null
 # -- Common annotations for all pods
 podAnnotations: {}
 
+# -- Common labels for all pods
+podLabels: {}
+
 # -- The number of old ReplicaSets to retain to allow rollback
 revisionHistoryLimit: 10
 
--- a/charts/loki-canary/Chart.yaml
+++ b/charts/loki-canary/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-canary
 description: Helm chart for Grafana Loki Canary
 type: application
 appVersion: 2.5.0
-version: 0.8.0
+version: 0.8.1
 home: https://github.com/grafana/helm-charts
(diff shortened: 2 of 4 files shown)


charts/loki-canary/values.yaml
```diff
@@ -21,6 +21,9 @@ priorityClassName: null
 # -- Common annotations for all pods
 podAnnotations: {}
 
+# -- Common labels for all pods
+podLabels: {}
```
trevorwhitney, line 25: WDYT of making this `selectorLabels` or `podSelectorLabels` to match the convention in the helper function `loki-canary.selectorLabels` well call right before including this?

trevorwhitney, line 25: WDYT of making this `selectorLabels` or `podSelectorLabels` to match the convention in the helper function `loki-canary.selectorLabels` well call right before including this?

trevorwhitney (commented):
Looks like a good addition, thank you! One small nit-picky question.

charts/loki-canary/values.yaml
trevorwhitney: ahh, so you want to differentiate these from `selectorLabels` as in they are _only_ meant for the pod. I agree with that argument

trevorwhitney (approved):
LGTM

praveenjindal62:
Hi @trevorwhitney , Could you please approve it again, I had to update README.md via helm-docs which I didn't do earlier so one of the workflow was failing

praveenjindal62:
@trevorwhitney  @unguiculus  @Whyeasy Please approve this merge request

PR #1076, opened 2022-02-24
[tempo] add security context config

kraman:

PR #1443, opened 2022-06-03
[tempo] add securityContext & enable removing tempoQuery container 

smartbit:
This PR
- allows adding `securityContext` to
  - pod `tempo`
  - container `tempo`
  - container `tempo-query`
- allows for removing the container `tempo-query` for two reasons
  - not needed for Grafana 7.5+ as [Grafana 7.5.x and higher can query Tempo directly](https://grafana.com/docs/tempo/latest/configuration/querying/#grafana-75x-and-higher-easy-mode)
  - container `tempo-query` fails with `securityContext.readOnlyRootFilesystem: true`

Maintainers: feel free to modify, reword comments/readme.md or remove parts. 

#### Technical Requirements 
- [x] Signed-off-by: smartbit <smartbit@users.noreply.github.com>
- [x] follows [Chart Best Practices Guide](https://helm.sh/docs/chart_best_practices/)
- [x] chart linted & installed successfully using ct
- [x] bumped version to `0.15.4`

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.15.3
+version: 0.15.4
 appVersion: 1.4.1
 engine: gotpl
 home: https://grafana.net
--- a/charts/tempo/README.md
+++ b/charts/tempo/README.md
@@ -1,6 +1,6 @@
 # tempo
 
-![Version: 0.15.3](https://img.shields.io/badge/Version-0.15.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.4.1](https://img.shields.io/badge/AppVersion-1.4.1-informational?style=flat-square)
+![Version: 0.15.4](https://img.shields.io/badge/Version-0.15.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.4.1](https://img.shields.io/badge/AppVersion-1.4.1-informational?style=flat-square)
 
 Grafana Tempo Single Binary Mode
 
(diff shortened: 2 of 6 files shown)


joe-elliott:
Thanks for the contribution. I like the idea of adding the pod security context, but I'm unsure about the other change. Tempo-query is no longer really necessary unless you're running an older version of Grafana. It doesn't really have anything to do with search.

https://grafana.com/docs/tempo/latest/configuration/querying/#grafana-74x

smartbit:
> Thanks for the contribution. 

You're welcome. Happy to try to give back something.

> I like the idea of adding the pod security context, 

Thanks. And the container securityContext, those are fine with you?

> but I'm unsure about the other change. Tempo-query is no longer really necessary unless you're running an older version of Grafana. It doesn't really have anything to do with search.
> 
> https://grafana.com/docs/tempo/latest/configuration/querying/#grafana-74x

My lack of knowledge of Tempo, I was under the impression that `tempoSearch` replaced `tempo-query`. Are you saying that without both `tempoSearch` and `tempo-query` Grafana 7.5.x+ can query tempo? In that case `{{- if or (not .Values.tempo.searchEnabled) .Values.tempoQuery.enabled }}` can be replace with `{{- if .Values.tempoQuery.enabled }}`. I'd be happy to add that change to the PR.

joe-elliott:
> Thanks. And the container securityContext, those are fine with you?
Yup, let's move forward with the `securityContext` change. 

Also, I'd prefer just removing the tempo-query change. Do you need this explicitly?

smartbit:
> Yup, let's move forward with the `securityContext` change.

:+1:

> Also, I'd prefer just removing the tempo-query change. Do you need this explicitly?

I assume you are asking why I think allowing the possibility to remove the tempo-query container is a good idea? For 2 reasons:

1. Tempo-query is not needed for those running [Grafana 7.5.x and higher](https://grafana.com/docs/tempo/latest/configuration/querying/#grafana-75x-and-higher-easy-mode). Why run a container that is not needed on your cluster? It takes resources, is an additional attack vector, requires maintenance, etc. worries we don't have once it is disabled.
2. The tempo-query container has a _feature_ that disallows `readOnlyRootFilesystem: true` and therefor it doesn't pass our compliance. IOW without `tempoQuery.enabled` we'd still need to maintain our own Chart.

I'd like to turn it around ask you why the charts can not be delivered with Full security enabled _out-of-the-box_? Make it a `1.15.0` release, and change these features to default:
1. enable all `securityContext` that are now commented-out in `values.yaml`
2. set `tempoQuery.enabled: false` to work around the issue with `readOnlyRootFilesystem: true`-not-possible in tempo-query  and update the comment to:
   ```yaml
   # -- set to True for [Grafana 7.4.x and earlier](https://grafana.com/docs/tempo/latest/configuration/querying/#grafana-74x)
   enabled: false
   ```

joe-elliott:
Thanks for the explanation. We can merge the tempo-query changes as well :+1:

joe-elliott (approved):

smartbit:
Great. 

Please close #1076 as it is now redundant.

PR #1452, opened 2022-06-06
[agent-operator] Add: Kubernetes recommended labels, custom labels and extra args

rjpnps:
#### What this PR does / why we need it:

**ENHANCEMENT**
In favour of using Kubernetes [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) to `agent-operator`:

- Label `app.kubernetes.io/component` has been moved from resources to common label template;

**BUG-FIX**
Manifest `operator-deployment.yaml` contained a bug where deployment template metadata labels, were using `ga-operator.selectorLabels` instead of `ga-operator.labels`.

**FEATURE**
In addition, **optional** custom user labels can now be added to `agent-operator` resources:

```yaml
...
# -- Additional labels to add to all resources
customLabels: {}
  # app: grafana-agent-operator
...
```

**FEATURE**
Chart now supports adding extra arguments for the cli using `extraArgs`:

```yaml
...
# List of additional cli arguments to configure grafana-agent-operator
# for example: --log.level, etc.
extraArgs: []
...
```

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - none

#### Special notes for your reviewer:
Minor has been bumped, since there is some new features.
There is no breaking change (selector labels remain the same) and upgrade is performed normally (i.e. `helm upgrade --install`).

#### Checklist

- [X] [DCO](https://github.com/grafana/helm-charts/blob/main/CONTRIBUTING.md#sign-off-your-work) signed
- [X] Chart Version bumped
- [X] Title of the PR starts with chart name (e.g. `[agent-operator]`)

### DIFF
--- a/charts/agent-operator/Chart.yaml
+++ b/charts/agent-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.14
+version: 0.2.0
 appVersion: "0.24.2"
 home: https://grafana.com/docs/agent/latest/
 icon: https://raw.githubusercontent.com/grafana/agent/v0.24.2/docs/assets/logo_and_name.png
--- a/charts/agent-operator/Chart.yaml
+++ b/charts/agent-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.12
+version: 0.2.0
 appVersion: "0.24.2"
 home: https://grafana.com/docs/agent/latest/
 icon: https://raw.githubusercontent.com/grafana/agent/v0.24.2/docs/assets/logo_and_name.png
--- a/charts/agent-operator/templates/_helpers.tpl
+++ b/charts/agent-operator/templates/_helpers.tpl
@@ -36,9 +36,14 @@ Common labels
 {{- define "ga-operator.labels" -}}
 {{ include "ga-operator.selectorLabels" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: operator
+app.kubernetes.io/part-of: {{ template "ga-operator.name" . }}
 helm.sh/chart: {{ include "ga-operator.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+{{- if .Values.customLabels }}
+{{ toYaml .Values.customLabels }}
 {{- end }}
 {{- end }}
 
--- a/charts/agent-operator/templates/_helpers.tpl
+++ b/charts/agent-operator/templates/_helpers.tpl
@@ -37,7 +37,6 @@ Common labels
 {{ include "ga-operator.selectorLabels" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/component: operator
-app.kubernetes.io/part-of: {{ template "ga-operator.name" . }}
 helm.sh/chart: {{ include "ga-operator.chart" . }}
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
--- a/charts/agent-operator/templates/operator-deployment.yaml
+++ b/charts/agent-operator/templates/operator-deployment.yaml
@@ -35,21 +35,16 @@ spec:
         resources:
         {{- toYaml . | nindent 10 }}
         {{- end }}
-        {{- if and .Values.kubeletService.namespace .Values.kubeletService.serviceName }}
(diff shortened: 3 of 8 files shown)


charts/agent-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.13
+version: 0.2.0
 appVersion: "0.24.2"
```
hjet, line 6: may as well bump the agent version to `0.25.0`

charts/agent-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.1.13
+version: 0.2.0
```
hjet, line 5: just confirming that these changes are backwards-compatible? (thanks for bumping the minor version, we should have been following semver anyways...)

charts/agent-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
...
 appVersion: "0.24.2"
 home: https://grafana.com/docs/agent/latest/
 icon: https://raw.githubusercontent.com/grafana/agent/v0.24.2/docs/assets/logo_and_name.png
```
hjet, line 8: as mentioned above, may as well bump agent version at the same time (but not necessary)

charts/agent-operator/templates/_helpers.tpl
```diff
@@ -36,10 +36,15 @@ Common labels
 {{- define "ga-operator.labels" -}}
 {{ include "ga-operator.selectorLabels" . }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/component: operator
+app.kubernetes.io/part-of: {{ template "ga-operator.name" . }}
```
hjet, line 40: moving the `component` label makes sense, thanks for catching that! if possible, i would omit the `part-of` label - if users need this, they can now create this label with your other `customLabels` addition.

charts/agent-operator/templates/operator-deployment.yaml
```diff
@@ -17,7 +16,7 @@ spec:
   template:
     metadata:
       labels:
-{{ include "ga-operator.selectorLabels" . | indent 8 }}
+{{ include "ga-operator.labels" . | indent 8 }}
```
hjet, line 19: thanks for catching this!

charts/agent-operator/templates/operator-deployment.yaml
```diff
@@ -39,6 +38,18 @@ spec:
         {{- if and .Values.kubeletService.namespace .Values.kubeletService.serviceName }}
```
hjet, line 38: would something like this maybe simplify the template here a bit? i found it a bit difficult to follow at first glance:

```yaml
        {{- if or .Values.kubeletService .Values.extraArgs }}
        args:
          {{- if .Values.kubeletService }}
          - --kubelet-service={{ .Values.kubeletService.namespace }}/{{ .Values.kubeletService.serviceName }}
          {{- end }}
          {{-  if .Values.extraArgs  }}
          {{- range .Values.extraArgs  }}
          - {{ . }}
          {{- end }}
          {{- end }}
        {{- end }}
```

hjet:
cc @captncraig

hjet:
would also bump image version in `values.yaml` at the same time

hjet:
also - thanks so much for your for your contribution!!! it's super appreciated

charts/agent-operator/templates/operator-deployment.yaml
rjpnps: Totally! I've done a bad job at that 😞  Have refactored with your approach - thanks for the hint 👍

charts/agent-operator/Chart.yaml
rjpnps: I would suggest to bump the agent version in a different PR, for the following reasons:

- The agent in `0.25.0` starts complaining that CRD `integrations.monitoring.grafana.com` is not found.
- It sounds to be a good moment as well, to sync the CRDs folder from https://github.com/grafana/agent/tree/main/production/operator/crds to the CRDs folder in the chart.
- Since there is a new CRD that the operator handles, upgrade path from `0.24.2` will require some thought? Including the new CRD and performing `helm upgrade --install --force` doesn't install the new CRD, nor update the existing one's.

charts/agent-operator/Chart.yaml
rjpnps: commented above about the agent version bump

charts/agent-operator/Chart.yaml
hjet: ah yeah that makes sense — since there is a new CRD we should definitely keep it separate

hjet (approved):
lgtm! thank you

PR #1453, opened 2022-06-06
Skip chart release if tag already exists

Logiraptor:
Also added some informational logging to aid debugging in the future

This is necessary for Mimir since we are moving to a weekly release instead of releasing every commit. There are potentially many commits with "changed" charts despite the version in `Chart.yaml` being unchanged. For example, this run failed when I merged some changes to main: https://github.com/grafana/mimir/actions/runs/2450081153

Signed-off-by: Patrick Oyarzun <patrick.oyarzun@grafana.com>


krajorama (approved):
LGTM

PR #1466, opened 2022-06-08
[loki-simple-scalable] Add region to s3 storage config options

### LINKED ISSUE #1458: [loki-simple-scalable] Cannot connect loki to s3
Hello,

We are trying to setup loki simple scalable on Kubernetes. The last step we have to do is to connect loki to our buckets in amazon s3. However, the configuration options in the yaml file are not clear since the commit 354995bf213554a9cd1f2152def1ebbd5a8d5bf3 of the loki-simple-scalable/values.yaml file.

There is a new **s3: null** item and we don't know what it does. We also tried to configure the region using the **region** item but the region stays to dummy.

So we have two problems:

- What is the purpose of s3: null 
- How to setup the region of the aws bucket

We are using the latest version (Chart version 1.2.0). Commit hash : b6af1bed803011da47623798a70c111af0d5bc14, date May 24, 2022.

Thanks.

trevorwhitney:
Fixes #1458

### DIFF
--- a/charts/loki-simple-scalable/Chart.yaml
+++ b/charts/loki-simple-scalable/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-simple-scalable
 description: Helm chart for Grafana Loki in simple, scalable mode
 type: application
 appVersion: 2.5.0
-version: 1.3.0
+version: 1.4.0
 home: https://grafana.github.io/helm-charts
 sources:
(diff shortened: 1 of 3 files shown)


Tycale (approved):

owen-d (approved):

DylanGuedes (approved):
lgtm!

PR #1470, opened 2022-06-08
[agent-operator] Add missing helm hook for test resource

allir:
Grafana Agent Operator. 
Installing using the chart creates a test resource as it does not have the appropriate helm hook to mark it as a test.


allir:
Added a commit to bump the chart version to pass `lint-test`

PR #1476, opened 2022-06-09
[grafana] add non object grafana.ini configs

karasjoh000:
There are configs that are not an object within grafana.ini e.g. force_migration=<bool>
This commit adds these configs. With this change this will be possible in values.yml: 
grafana.ini:
  force_migration: true


zanhsieh:
@karasjoh000 
Could you bump the Chart.yaml version to latest? Thank you.

zanhsieh (approved):

Issue #1459, opened 2022-06-07
[loki-distributed] Add a pure Ingress option


aplovich:
Hello,

The gateway service is redundant with K8s' Ingress, which is able to handle the request routing it does.  This adds an extra hop between the Ingress and Loki's services.  It also adds an external dependency for the chart.

It'd be nice to have the option of routing requests directly from the Ingress to the services.

trevorwhitney:
Can you provide an example or a PR? I'm not familiar with this approach, and am curious how you would handle routing, like `/distributor/ring` or `/distributor/config`? Also, could auth be handled in this approach?

aplovich:
I'm working on a PR currently, I'll look at submitting it later today.

To answer your questions:  
You can expose the ring / config endpoints through the ingress, but, since they contain sensitive information, a better approach would probably be to have users directly connect to those when they need to troubleshoot.

kubectl port-forward svc/loki-loki-distributed-distributor 3100:3100 
then access http://localhost:3100/config or /ring in your browser.

authn is handled by the ingress controller, so it's implementation specific.  I'll be including an example for the nginx-ingress controller.

ruant:
@trevorwhitney @zanhsieh 
This issue can be closed? 
https://github.com/grafana/helm-charts/pull/1478 is merged and implements this.

PR #1478, opened 2022-06-10
[loki-distributed] Add a pure ingress option to use instead of the gateway service

aplovich:
Signed-off-by: Anthony Plovich <aplovich@anl.gov>

This adds the option to route requests using just a K8s ingress instead of the gateway service.  
https://github.com/grafana/helm-charts/issues/1459

Some things to note:
- Users will have to supply authn configuration specific to their ingress controller.
- If you're hosting Grafana on the same cluster, you'll need to point its Loki datasource at the ingress URL instead of one of the cluster's internal services.


charts/loki-distributed/templates/ingress.yaml
```diff
@@ -0,0 +1,55 @@
+{{- if .Values.ingress.enabled }}
```
trevorwhitney, line 1: should we also check to make sure the gateway is disabled here? I'd like to enforce the 1 or the other restriction in code if possible.

trevorwhitney (commented):
I don't have much experience with k8s ingress, but this seems reasonable to me. I would like to enforce the 1-or-the-other gateway vs. ingress in code. The mimir [helm-chart](https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/templates/validate.yaml) does some of this stuff through a `validate` tempalte, WDYT of something like that?

charts/loki-distributed/templates/ingress.yaml
aplovich: Sounds like a good idea.  I'll take a look at using the validate template for this today, I haven't used one for that yet.

trevorwhitney (approved):
lgtm

PR #1481, opened 2022-06-11
[grafana] fix cve-2021-25742

zanac1986:

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.29.7
+version: 6.29.8
 appVersion: 8.5.3
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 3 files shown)


zanhsieh (approved):

afirth:
Hello,

This configuration is only valid on the ingress-nginx configmap, but you seem to have applied it to the grafana configmap. Is there another meaning of this annotation in grafana-land? Otherwise, I'm pretty sure you can't "fix" this here. I assume you got here from [this linter](https://hub.datree.io/built-in-rules/prevent-configmap-security-vulnerability-cve-2021-25742#how-to-fix-this-failure) but this linter is wrong. It only applies to the ingress-nginx configmap - not all configmaps in the cluster (that would be a K8s vulnerability, not an ingress-nginx one). Suggest to revert. I will open an issue too.

Thanks for all the hard work maintaining!

PR #1488, opened 2022-06-13
[loki-simple-scalable] Change default volume retention policies

trevorwhitney:

### DIFF
--- a/charts/loki-simple-scalable/CHANGELOG.md
+++ b/charts/loki-simple-scalable/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 1.5.0
+
+-[CHANGE] Set `persistentVolumeClaimRetentionPolicy` for read pods to `Delete` for both `whenDeleted` and `whenScaled` since this data is easy to replace by re-fetching from object storage. This should make for easier roll-outs of the stateful sets.
+
 ## 1.3.0
 
 - [BUGFIX] Fix the storage heper functions to use the right properties (snake vs camel case) #1420
--- a/charts/loki-simple-scalable/templates/read/statefulset-read.yaml
+++ b/charts/loki-simple-scalable/templates/read/statefulset-read.yaml
@@ -13,7 +13,7 @@ spec:
       partition: 0
   serviceName: {{ include "loki.readFullname" . }}-headless
   revisionHistoryLimit: {{ .Values.loki.revisionHistoryLimit }}
-  {{- if (semverCompare ">= 1.23-0" .Capabilities.KubeVersion.Version) -}}
+  {{- if (semverCompare ">= 1.23-0" .Capabilities.KubeVersion.Version) }}
   {{/*
     Data on the read nodes is easy to replace, so we want to always delete PVCs to make
     operation easier, and will rely on re-fetching data when needed.
@@ -21,7 +21,7 @@ spec:
   persistentVolumeClaimRetentionPolicy:
     whenDeleted: Delete
     whenScaled: Delete
-  {{- end -}}
+  {{- end }}
   selector:
     matchLabels:
       {{- include "loki.readSelectorLabels" . | nindent 6 }}
--- a/charts/loki-simple-scalable/templates/read/statefulset-read.yaml
+++ b/charts/loki-simple-scalable/templates/read/statefulset-read.yaml
@@ -13,6 +13,15 @@ spec:
       partition: 0
   serviceName: {{ include "loki.readFullname" . }}-headless
   revisionHistoryLimit: {{ .Values.loki.revisionHistoryLimit }}
+  {{- if (semverCompare ">= 1.23-0" .Capabilities.KubeVersion.Version) -}}
+  {{/*
+    Data on the read nodes is easy to replace, so we want to always delete PVCs to make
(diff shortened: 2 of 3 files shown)


charts/loki-simple-scalable/templates/write/statefulset-write.yaml
```diff
@@ -14,6 +14,9 @@ spec:
       partition: 0
   serviceName: {{ include "loki.writeFullname" . }}-headless
   revisionHistoryLimit: {{ .Values.loki.revisionHistoryLimit }}
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: Retain
+    whenScaled: Delete
```
dannykopping, line 19: Can you please elaborate (or even better - add inline comments) about why we want to retain a PVC when deleting a `write` pod, but delete it when scaling down?

charts/loki-simple-scalable/templates/read/statefulset-read.yaml
```diff
@@ -13,6 +13,9 @@ spec:
       partition: 0
   serviceName: {{ include "loki.readFullname" . }}-headless
   revisionHistoryLimit: {{ .Values.loki.revisionHistoryLimit }}
+  persistentVolumeClaimRetentionPolicy:
+    whenDeleted: Delete
+    whenScaled: Delete
```
dannykopping, line 18: Forgive my ignorance - which parts of the `read` target are using statefulsets? The compactor and ruler?

charts/loki-simple-scalable/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 1.5.0
+
+-[CHANGE] Set `persistentVolumeClaimRetentionPolicy` for read pods to `Delete` for both `whenDeleted` and `whenScaled`, and only `Delete` for `whenSclaed` on write pods. This should make for easier roll-outs of the stateful sets.
```
dannykopping, line 16:
```suggestion
-[CHANGE] Set `persistentVolumeClaimRetentionPolicy` for read pods to `Delete` for both `whenDeleted` and `whenScaled`, and only `Delete` for `whenScaled` on write pods. This should make for easier roll-outs of the stateful sets.
```

charts/loki-simple-scalable/templates/read/statefulset-read.yaml
trevorwhitney: it's actually probably less important for the `compactor` and `ruler`, and most important for fetched indexes, which was previously the job of the `querier`, but is now the job of the `indexgateway` which is running on the `read` nodes as a result of a recent PR to loki.

charts/loki-simple-scalable/templates/write/statefulset-write.yaml
trevorwhitney: The reason for this is based on my understanding of how the [rollout operator](https://github.com/grafana/rollout-operator) will work. If I understand correctly, since the update strategy is set to `OnDelete`, the stateful sets will not update on their own, and instead the rollout operator will delete the old replica for a given zone and replace it with a new one. In this case, we want to `Retain` the existing PVCs so the new pods can reuse them. When scaling, we want to the `Delete` the PVCs however, as our rings will re-shard any data they were holding to other instances, and that data is no long valid. Setting retention to `Delete` for scaling events will hopefully make those scaling events easier from an operations perspective.

trevorwhitney: Also, this was the best resource I could find on the difference between the different policy types: https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/#the-new-statefulset-pvc-retention-policy

pracucci: Few comments:
- We're **not** used to set the `persistentVolumeClaimRetentionPolicy` in Mimir
- I think `persistentVolumeClaimRetentionPolicy` also requires K8S 1.23 or above ([doc](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention))
- According to the doc, `whenDeleted` "configures the volume retention behavior that applies when the StatefulSet is deleted". It's not when the pod is deleted, but when the StatefulSet is deleted. I would be suprised if a PVC would get deleted as soon as a pod is deleted, because the pod deletion is just a normal operation which occurs during rollout even when using the `RollingUpdate` update strategy (the difference is who deletes pod, either the StatefulSet controller or an external entity, like the rollout-operator).

All in all, I think that:
- You don't want to set `persistentVolumeClaimRetentionPolicy` if you want to support K8S < 1.23
- If you set it, you may want to set `Retain` for both `whenDeleted` and `whenScaled` for the **write** path, because if you scale down or delete the StatefulSet by mistake, you don't want to loose WALs (that's the reasoning we did for Mimir ingesters too)

charts/loki-simple-scalable/templates/read/statefulset-read.yaml
trevorwhitney: @dannykopping added a comment, and removed the change to write pods since, as @pracucci pointed out, better to play it safe with that data as it's not replaceable like read path data.

dannykopping (approved):
LGTM

pracucci (approved):

PR #1489, opened 2022-06-13
[tempo-distributed] Add distributor configs

AlexDHoffer:
These are missing configs that would be really helpful to have in the chart.


joe-elliott:
Another nice addition!

PR #1496, opened 2022-06-14
[grafana] Add egress to network policy

JWilson45:
Found a need for egress to be added to the Grafana network policy. In some cases, datasources outside the cluster or namespace will need an egress to be able to connect.

Adding egress to the network policy with the ability to specify ports addresses this issue.

Signed-off-by: Wilson, Jason M <jason.wilson2@marist.edu>


JWilson45:
Testing proves that this helm chart addition is backwards compatible.
Removing all new values defaults the network policy to create only an ingress on the service port.

PR #1339, opened 2022-05-04
gateway: Set the HTTP version for proxying to HTTP/1.1

primeroz:
By default nginx uses HTTP/1.0 for proxied requests : https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version

This is suboptimal since promtail uses HTTP/1.1 and can cause issues with Istio which , by default, will reject anything lower than HTTP/1.1

Slack conversation for reference: https://grafana.slack.com/archives/CEPJRLQNL/p1651595415595109

Signed-off-by: Francesco Ciocchetti <primeroznl@gmail.com>

PR #1506, opened 2022-06-17
[loki-distributed] gateway: Set the HTTP version for proxying to HTTP/1.1

primeroz:
related to: https://github.com/grafana/helm-charts/pull/1339

By default nginx uses HTTP/1.0 for proxied requests : https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version

This is suboptimal since promtail uses HTTP/1.1 and can cause issues with Istio which , by default, will reject anything lower than HTTP/1.1

Signed-off-by: Francesco Ciocchetti <primeroznl@gmail.com>


ducnm0711:
hi guys we are also looking forward this update. Would be nice if this got approved

zanhsieh (approved):

PR #1507, opened 2022-06-17
[loki-simple-scalable] By default nginx uses HTTP/1.0 for proxied requests : https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version

primeroz:
related to https://github.com/grafana/helm-charts/pull/1339

By default nginx uses HTTP/1.0 for proxied requests : https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version

This is suboptimal since promtail uses HTTP/1.1 and can cause issues with Istio which , by default, will reject anything lower than HTTP/1.1

Slack conversation for reference: https://grafana.slack.com/archives/CEPJRLQNL/p1651595415595109

Signed-off-by: Francesco Ciocchetti primeroznl@gmail.com


charts/loki-simple-scalable/values.yaml
```diff
@@ -249,6 +249,8 @@ enterprise:
         uwsgi_temp_path       /tmp/uwsgi_temp;
         scgi_temp_path        /tmp/scgi_temp;
 
+        proxy_http_version    1.1;
```
trevorwhitney, line 252: does this have any backwards compatibility issues?

trevorwhitney (changes requested):
please sign commit

charts/loki-simple-scalable/values.yaml
primeroz: this affects the connections between nginx and the loki components. 

to my knowledge no backward compatiblity issues exists , HTTP1.1 has been out for a while :)

I _could_ put it behind a flag in helm but unless is the default value it will probably keep being used as 1.0 by most people ... which indeed is what i think is happening on nginx. The default has been left to 1.0 for legacy reason and people do reverse proxy using http 1.0 without even knowing.

primeroz: I will change all the prs ( i did one for each chart ) and set a flag in the values to enable proxy 1.1 with default value false , thinking about it i think is better.

trevorwhitney: sounds good. please give me a mention when it's ready for another review :smile:

primeroz:
Replaced by #1600

PR #1511, opened 2022-06-17
[loki-distributed] Configure Loki liveness probes

QuentinBisson:
This PR adds liveness probes to all the components of the chart and also add the readiness probe to the query frontend as it was somehow missing


QuentinBisson:
@unguiculus could you approve the github action workflow and take a look at this PR ?

charts/loki-distributed/templates/query-frontend/service-query-frontend.yaml
```diff
@@ -10,6 +10,7 @@ metadata:
 spec:
   type: ClusterIP
   clusterIP: None
+  publishNotReadyAddresses: true
```
QuentinBisson, line 13: We need to add this after adding a readiness probe to the query frontend, as otherwise the dns name is not added to the service and no querier can connect to the query frontend as defined at the end of the section https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields

QuentinBisson:
@Whyeasy could you take a look ?

zanhsieh (approved):

PR #1518, opened 2022-06-20
[promtail] Change `/run/promtail` mount to configurable with `values.yaml`

### LINKED ISSUE #447: [promtail] Promtail incompatible with GKE Autopilot
Promtail Helm chart version: v3.5.1

As described in the [GKE Autopilot docs](https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-overview#host_options_restrictions):

> Using hostPath volumes in write mode is prohibited, while using hostPath volumes in read mode is allowed only for /var/log/ path prefixes.

Currently, the DaemonSet is [hard-coded](https://github.com/grafana/helm-charts/blob/76ef792c3d2910f2a4c4996a66c77cc65daa0074/charts/promtail/templates/daemonset.yaml#L123-L125) to read and write to hostPath `/run/promtail`.
This will make Promtail Helm chart installation fail with error:
```
failed to create resource: admission webhook "validation.gatekeeper.sh" denied the request: [denied by autogke-no-write-mode-hostpath] hostPath volume run in container promtail is accessed in write mode; disallowed in Autopilot.
```

Are there any plans to support Promtail for GKE Autopilot?
Our use case is shipping logs from the GKE managed K8s nodes' `/var/log/pods/` directories to Loki.

### LINKED ISSUE #644: [promtail] Allow to configure all daemonset volumes
Hi! 
As a part of running Promtail to collect logs from our Windows EKS worker nodes, I hit an issue with re-using the official chart with our custom-build image. Right now chart allows to configure mounting container log directories from host, but mount for positions file is hardcoded:
https://github.com/grafana/helm-charts/blob/2d25705d20e439a51b5d7e3363bd79056036f3ea/charts/promtail/templates/daemonset.yaml#L124-L126
If this volume is moved to `defaultVolumes` in values as well, we can re-use the chart without the need to fork it.
I can share the details on our current setup as well.

ykzts:
⚠ **BREAKING CHANGE** ⚠

By making `/run/promtail` mount configurable, platforms with limited `hostPath` specification (e.g. GKE Autopilot) will be able to use Helm Chart of Promtail.

Close #447
Close #644

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.5.0
-version: 5.1.0
+version: 6.0.0
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
--- a/charts/promtail/README.md
+++ b/charts/promtail/README.md
@@ -1,6 +1,6 @@
 # promtail
 
-![Version: 5.1.0](https://img.shields.io/badge/Version-5.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.5.0](https://img.shields.io/badge/AppVersion-2.5.0-informational?style=flat-square)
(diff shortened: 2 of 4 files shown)


zanhsieh:
@ykzts 
Could you bump the Chart.yaml version to latest and fix lint error in the bottom of this PR page please?

zanhsieh (approved):

heph2:
Hi!
Sorry if I reopen the discussion, but i cannot understand how should i manage that in a GKE Autopilot scenario..
Should i override those values with something else?
```
  - name: run
    hostPath:
      path: /run/promtail
```

@ykzts 

Thanks in advance

PR #1520, opened 2022-06-21
[tempo-distributed] add option to configure multi_tenancy from values files

unnirapido:
Currently, the `multitenancy_enabled: false` is hardcoded in [config](https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/values.yaml#L550), and to override it, we'll have to override the entire config string. The proposed change is to modify the `multitenancy_enabled ` value using values.yaml

Signed-off-by: Krishnanunni Menon krishnanunni.menon@rapido.bike


annanay25 (commented):
Nice contribution, thanks! Will run CI on this now.

PR #1524, opened 2022-06-21
Set annotations for services from values file

aravandi:
Signed-off-by: Amir <amirghotbi@gmail.com>

This PR will help us to be able to set annotation for all Kubernetes services.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.20.0
+version: 0.21.0
 appVersion: 1.4.1
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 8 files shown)


charts/tempo-distributed/README.md
```diff
@@ -19,6 +19,13 @@ helm repo add grafana https://grafana.github.io/helm-charts
 ## Upgrading
 
 A major chart version change indicates that there is an incompatible breaking change needing manual actions.
+### From Chart version < 0.21.0
```
joe-elliott, line 22: This isn't a breaking change so we don't need an upgrade guide. We should also only increment the patch version instead of the minor version: `0.20.0` -> `0.20.1`

aravandi: Thank you @joe-elliott for reviewing my PR. I've updated the chart version.

charts/tempo-distributed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.20.0
```
annanay25, line 5: This was just updated in another PR. Could you please rebase/merge main? Thanks!

aravandi: Thank you @annanay25, I've fixed the conflict.

PR #1525, opened 2022-06-22
[loki-distributed] Fix ignoring KubeVersion containing a platform version

ksauzz:
Hi,

This PR is to handle KubeVersion containing a platform version like `1.22.9-eks-1` appropriately.

### Background

We run loki-distributed on EKS, and tried to upgrade loki-distributed to 0.48.5 to use `policy/v1` API for PodDisruptionBudgets. But `policy/v1` api doesn't appeared in the manifest.
According to [the docs](https://helm.sh/docs/chart_template_guide/function_list/#working-with-prerelease-versions) , semverCompare function skips a such KubeVersion if the version in the condition doesn't contain the suffix `-0`. This works on my environment.

Thanks.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.5.0
-version: 0.48.5
+version: 0.48.6
 home: https://grafana.github.io/helm-charts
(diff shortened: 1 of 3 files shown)


zanhsieh:
@ksauzz 
Please bump Chart.yaml versions to latest. Thanks a lot.

zanhsieh (approved):

ksauzz:
Thank you for quick action!!

PR #1532, opened 2022-06-23
[loki-distributed] Allow loadBalancerSourceRanges param on gateway service

ialejandro:
Add param `loadBalancerSourceRanges` on gateway service

Signed-off-by: Iván Alejandro Marugán <hello@ialejandro.rocks>


charts/loki-distributed/templates/gateway/service-gateway.yaml
```diff
@@ -17,8 +17,16 @@ spec:
   {{- with .Values.gateway.service.clusterIP }}
   clusterIP: {{ . }}
   {{- end }}
-  {{- if and (eq "LoadBalancer" .Values.gateway.service.type) .Values.gateway.service.loadBalancerIP }}
-  loadBalancerIP: {{ .Values.gateway.service.loadBalancerIP }}
+  {{- if eq "LoadBalancer" .Values.gateway.service.type }}
+  {{- if .Values.gateway.service.loadbalancerip }}
+  loadbalancerip: {{ .Values.gateway.service.loadbalancerip }}
+  {{- end }}
+  {{- if .Values.gateway.service.loadBalancerSourceRanges  }}
```
trevorwhitney, line 24: nit: WDYT of using a with here?

trevorwhitney (approved):
lgtm, thanks for the contribution. 1 small nit.

charts/loki-distributed/templates/gateway/service-gateway.yaml
ialejandro: Hi @trevorwhitney, thx a lot.

When I develop my own Service objects always use range (like Prometheus chart: https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/prometheus/service.yaml#L29-L33), but I checked grafana chart Helm (https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/service.yaml#L27-L30) and use `toYaml .Values.service.loadBalancerSourceRanges | indent 4`. IMHO, result will be the same.

trevorwhitney: yeah, I don't think we need a range, but was suggesting a `with` instead of the `if`.

ialejandro:
> @ialejandro just a few conficts to resolve, and you may need to bump the version after the rebase on master, then we can get this merged! thanks again!

perfect! thx for you quick reply :)

ialejandro:
ping @unguiculus / @Whyeasy

zanhsieh (approved):

PR #1535, opened 2022-06-24
[loki-simple-scalable] Fix: Simple Scalable Model Helm Chart README examples

Alex3k:
**Scope**: Simple Scalable Model Helm Chart Documentation

**Description**
[This commit ](https://github.com/grafana/helm-charts/commit/cab9e61ebdda23129bd59b4a813f5dd4749fc36b) updated the simple scalable model charts' loki.config value to be a string. However the examples were broken within the README. This PR fixes those examples.

### DIFF
--- a/charts/loki-simple-scalable/README.md
+++ b/charts/loki-simple-scalable/README.md
@@ -219,7 +219,7 @@ For exmaple, to use MinIO (deployed separately) as your backend, provide the fol
 
 ```yaml
 loki:
-  config:
+  config: |
     common:
       storage:
         filesystem: null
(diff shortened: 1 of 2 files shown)


chaudum (approved):
lgtm

charts/loki-simple-scalable/README.md
```diff
@@ -219,7 +219,7 @@ For exmaple, to use MinIO (deployed separately) as your backend, provide the fol
 
 ```yaml
 loki:
-  config:
+  config: |
```
trevorwhitney, line 222: @Alex3k these changes need to be made to the gotmpl file instead of the actual README. CI missed this because we we're currently ignoring generating helm-docs for this chart, but as soon as we turn it back on this change will get lost.

PR #1540, opened 2022-06-28
[tempo-distributed] use `priorityClassName` values

ben-natan:
The `priorityClassName` values were not used anywhere in the templates. This PR uses them the way they were intended: they can be individually set for each component or overridden for all with `global.priorityClassName`.


joe-elliott (approved):
Oof, nice fix. Didn't realize these values weren't mapped.

PR #1541, opened 2022-06-28
Remove confusing mimir-distributed-beta chart from helm repo

krajorama:
This was never really in use, since we switched to using actual
beta versions instead with mimir-distributed chart.


Logiraptor:
Thanks for doing this, should we also remove the source? mimir-openshift-experimental is still in here

krajorama:
> Thanks for doing this, should we also remove the source? mimir-openshift-experimental is still in here

There's no source for the -beta.
Regarding openshift-experimental: now that we have  a PR going in mimir, yes.

PR #1546, opened 2022-06-29
Adjustment of Tempo Helm chart

gruppferi:
hello,
The Tempo Helm chart needed some adjustment regarding the plcement of S3 storage. the commented example was in wrong place and alse indentition was wrong s3 object which was set under backend: s3

### DIFF
--- a/charts/tempo/values.yaml
+++ b/charts/tempo/values.yaml
@@ -41,19 +41,18 @@ tempo:
   server:
     # -- HTTP server listen port
     http_listen_port: 3100
-  # tempo storage backend
-  # refer https://github.com/grafana/tempo/tree/master/docs/tempo/website/configuration
-  ## Use s3 for example
-  # backend: s3                                         # store traces in s3
-  #  s3:
-  #    bucket: tempo                                   # store traces in this bucket
-  #    endpoint: s3.dualstack.us-east-2.amazonaws.com  # api endpoint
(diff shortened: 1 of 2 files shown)


BitProcessor:
Hi - thanks for your contribution. 
Would you be able to update the helm docs? 
See: https://github.com/grafana/helm-charts/runs/7122990323?check_suite_focus=true

charts/tempo/values.yaml
```diff
@@ -41,19 +41,18 @@ tempo:
...
-  #    access_key: ...                                 # optional. access key when using static credentials.
-  #    secret_key: ...                                 # optional. secret key when using static credentials.
-  #    insecure: false                                 # optional. enable if endpoint is http
-  ## end
   storage:
     trace:
+      # tempo storage backend
+      # refer https://github.com/grafana/tempo/tree/master/docs/tempo/website/configuration
+      ## Use s3 for example
+      # backend: s3                                         # store traces in s3
```
BitProcessor, line 49: Could you please fix the indentation of `# store traces in s3` so that it is aligned with the lines below.

BitProcessor (changes requested):

swartz-k:
LGTM. Maybe commits can be squashed? I think it will be better.

zanhsieh (approved):

BitProcessor (approved):

Issue #1099, opened 2022-03-09
[Loki-Simple-Scalable] Memberlist port name protocol breaks with istio


pastjean:
in: `charts/loki-simple-scalable/templates/service-memberlist.yaml`

The port name is "http", since the protocol is not really tcp this conflicts when istio is enabled and the istio port selection (https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) breaks the memberlist protocol (500: empty ring). 

We're not the first ones to get this: https://community.grafana.com/t/empty-ring-error-with-memberlist/51857

Our fix was to rename the port name to `tcp` like this:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: {{ include "loki.fullname" . }}-memberlist
  labels:
    {{- include "loki.labels" . | nindent 4 }}
spec:
  type: ClusterIP
  clusterIP: None
  ports:
    # This is a fork of the https://github.com/grafana/helm-charts
    # Where we change the "version-name" & the memberlist service port name from http to tcp
    # due to istio port selection
    - name: tcp
      port: 7946
      targetPort: http-memberlist
      protocol: TCP
  selector:
    {{- include "loki.selectorLabels" . | nindent 4 }}
    app.kubernetes.io/part-of: memberlist
```

LinAnt:
Would not adding appProtocol to the ports be a more elegant solution?

https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol

YvesZelros:
We are face to the same issue with Istio but on the Grafana service that is defined as 
`  ports:
  - appProtocol: http
    name: service
    port: 3000
    protocol: TCP
    targetPort: 3000
`
Port name must be renamed to http or a field appProtocol must be added like this

`  ports:
  - appProtocol: http
    name: service
    port: 3000
    protocol: TCP
    targetPort: 3000
`

trevorwhitney:
is there any consensus around a solution here? is the addition of `appProtocol: http` enough to solve this problem? Would this also require renaming the port to `tcp`? Does this affect other systems in any way that are not using istio? Curious if the pod annotation is enough for now until we gather more data on how many people are affected by this?

LinAnt:
@rlex do exclude the port, either add `appProtocol: http`, or memberlist uses some other protocol that is not pure http add appProtocol: tcp.

Changing the port name also works, but that might break other things already running in your cluster. i.e. serviceMonitors relyijng on port names.

[...]

PR #1549, opened 2022-06-29
[loki-distributed] Add optional appProtocol to memberlist svc

jdgeisler:
This PR references the same issue as https://github.com/grafana/helm-charts/issues/1099, but for the loki-distributed helm-chart.

Using istio protocol selection, the ports break in the memberlist service and throw the following error:
```
level=warn ts=2021-08-13T10:42:42.205933Z caller=client.go:344 component=client host=api.*.appdomain.cloud msg=“error sending batch, will retry” status=500 error="server returned HTTP status 500 Internal Server Error (500): empty ring"
```

This can be fixed by adding the `appProtocol` field in the memberlist service and setting it to `tcp` for loki-distributed. If appProtocol is left empty, then it will be ignored and not included in the service.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.5.0
-version: 0.49.0
+version: 0.50.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/charts/loki-distributed/README.md
+++ b/charts/loki-distributed/README.md
@@ -1,6 +1,6 @@
 # loki-distributed
 
-![Version: 0.49.0](https://img.shields.io/badge/Version-0.49.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.5.0](https://img.shields.io/badge/AppVersion-2.5.0-informational?style=flat-square)
+![Version: 0.50.0](https://img.shields.io/badge/Version-0.50.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.5.0](https://img.shields.io/badge/AppVersion-2.5.0-informational?style=flat-square)
 
 Helm chart for Grafana Loki in microservices mode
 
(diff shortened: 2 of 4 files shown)


trevorwhitney (approved):
lgtm

CharlieC3:
@jdgeisler After specifying the `appProtocol` as `tcp`, I started seeing these warnings every few seconds for all loki microservices:
```
Was able to connect to loki-<component> over TCP but UDP probes failed, network may be misconfigured
```

Are you seeing this too? Any idea how to resolve this problem without removing the istio sidecar from Loki?

jdgeisler:
@CharlieC3 hmm we are not seeing that error in any of our loki microservices with `appProtocol` set to `tcp` in memberlist.

What is your current configuration? Maybe something else could be conflicting with it.

rlex:
memberlist protocol is specified as TCP anyway, so UDP probes will always fail with or without appProtocol

CharlieC3:
@jdgeisler You said you're using `istio` too, right? I can confirm if I disable sidecar injection for all Loki pods, I don't see this error anymore. It's only when istio injects a sidecar into the Loki pods.

I don't think I'm doing anything special. I have a pretty vanilla deployment of the loki-distributed Helm chart deployed to a cluster in GKE (with a dataplane v2 network), shipping logs to GCS. The Istio deployment is pretty simple as well, coming from [their Helm charts](https://github.com/istio/istio/tree/master/manifests/charts).

Any ideas what could be happening?

CharlieC3:
For posterity, I was able to get around this by adding the following labels to all Loki pods (not deployments/STSs) in the chart:
```
traffic.sidecar.istio.io/excludeInboundPorts: "7946"
traffic.sidecar.istio.io/excludeOutboundPorts: "7946"
```

However, am still curious as to why this is happening. Running a manual UDP test shows the ports are open and able to communicate. I had to add these annotations to my Grafana Mimir deployment as well since that uses the memberlist/gossip-ring protocol too.

jdgeisler:
@CharlieC3 yes, we are using Istio and have the loki-distributed helm chart deployed to an EKS cluster. 

I attempted to reproduce your issue but was unable to replicate that behavior. From what I've seen though, this may just be an ignorable error from memberlist. Do you see ingestion or querying impacted from these errors, or is everything working as expected?

Here is a sort of related thread for Tempo https://github.com/grafana/tempo/issues/927#issuecomment-1155103282

CharlieC3:
@jdgeisler I didn't see any performance impact while the issue was happening. I haven't been able to find anyone that's having the same problem as me either. I feel it may be specific to GKE clusters running a certain version of Cilium, but don't have an hard evidence to support that.

PR #1572, opened 2022-07-05
[loki-simple-scalable]Added examples and some associated READMEs

Alex3k:
Added some examples of override values files for GCS and GKE so folks can take that and fill in the placeholder variables to get started quicker. Also added READMEs which describe how they work. This was done for both GEL and Loki OSS.


Alex3k:
@trevorwhitney do I need to bump the chart version for this considering it's just doc changes? Also, I have used both relative and absolute paths to point to things. Do you have a preference which I use?

zanhsieh:
@Alex3k 
Please bump the Chart.yaml version to latest and then rerun helm-doc to update README.md. Thank you.

Issue #1550, opened 2022-06-30
[loki-simple-scalable] s3/IRSA authentcation issues upgrading from 0.4.0


cebidhem:
Hey community,

I've tried to upgrade `loki-simple-scalable` from `0.4.0` to `1.4.3`, and we are running into a lot of issues, some resolved but some not at all.
I've tried to look for similar issues in GitHub, found some but still, even following some of the resolutions proposed, it's still not running properly.
I've tried to look also in the documentation - readme and online docs - but we know from the beginning it's not up-to-date.

We are using a quite basic setup, s3 as data backend and IRSA for authentication. I tried a few configuration, initially the bucketname was not set properly, then without setting accesskey and secretacesskey (as in 0.4.0), it fails because of the chart values defaults, and setting them to null gives me an unmarshall error.

**0.4.0 values:**
```yaml
    fullnameOverride: loki
    read:
      persistence:
        size: 5Gi
        storageClass: ebs-sc
    write:
      persistence:
        size: 5Gi
        storageClass: ebs-sc
    loki:
      commonConfig:
        storage:
          filesystem: null
          s3:
            s3: s3://eu-west-1
            bucketnames: my-company-loki-objstore
      schemaConfig:
        configs:
          - from: "2020-10-24"
            store: boltdb-shipper
            object_store: s3
            schema: v11
            index:
              prefix: loki_index_
              period: 24h
      storageConfig:
        boltdb_shipper:
          shared_store: s3
    gateway:
      ingress:
        enabled: true
        ingressClassName: ingress-nginx
        annotations:
          kubernetes.io/tls-acme: "true"
          cert-manager.io/cluster-issuer: letsencrypt
        hosts:
          - host: loki.mydomain.com
            paths:
              - path: /
                pathType: Prefix
        tls:
          - secretName: loki-gateway-tls
            hosts:
              - loki.mydomain.com
    serviceAccount:
      annotations:
        eks.amazonaws.com/role-arn: arn:aws:iam::redacted_aws_account_id:role/loki-irsa-role
    serviceMonitor:
      enabled: true
```
This works perfectly fine.

**1.4.3 values:**
```yaml
    fullnameOverride: loki
    gateway:
      ingress:
        annotations:
          cert-manager.io/cluster-issuer: letsencrypt
          kubernetes.io/tls-acme: "true"
        enabled: true
        hosts:
        - host: loki.mydomain.com
          paths:
          - path: /
            pathType: Prefix
[...]

PR #1575, opened 2022-07-05
[loki-simple-scalable] Change s3 defaults to null

trevorwhitney:
Fixes: #1550 

* this is better as often they will not be overriden, for example if
  leveraging env vars or IRSA


cebidhem:
@trevorwhitney I've created a PR here https://github.com/trevorwhitney/helm-charts/pull/1 to fix the CI above.

I leave it to you to use it or fix directly in your branch.

Thanks!

JordanRushing (approved):

PR #1576, opened 2022-07-06
[loki-simple-scalable] Update loki simple scalable read me

Alex3k:
Updated Loki Simple scalable read me to fix broken link to examples


trevorwhitney:
@Alex3k looks to be just a version bump here? Did a previous PR fail to build because of incorrect version?

Issue #534, opened 2021-06-29
[loki-distributed] Prometheus metrics


dyvol777:
Now, no pod or service is annotate for prometheus scraping. 
May be you can add it?

davidsuba:
At least ability to add custom annotations for service would be useful. Right now there is no way how to annotate services for prometheus scraping

PR #1577, opened 2022-07-06
[loki-distributed] Add service annotations to loki and memcached

jdgeisler:
Adding optional service annotations to all loki services and memcached services. This is required if a user wants to use service annotations for prometheus scraping.

Example values.yaml:
```yaml
loki:
  serviceAnnotations:
    prometheus.io/scrape: 'true'
```

Example service:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: loki-distributed-query-frontend
  labels:
    app.kubernetes.io/name: loki-distributed
  annotations:
    prometheus.io/scrape: "true
```

This references https://github.com/grafana/helm-charts/issues/534

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.5.0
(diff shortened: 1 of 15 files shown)


trevorwhitney (approved):
lgtm, thanks for the addition!

PR #1581, opened 2022-07-07
[loki-distributed] add possibility to set affinity for compactor deployment

hervenicol:
Added the possibility to define affinity just like for other components. Except this one is optional.
This way we can avoid a mix of affinities and selectors in our custom values.


zanhsieh:
Duplicate with:
https://github.com/grafana/helm-charts/pull/647
`{{- if ... }}` is not necessary.

PR #1582, opened 2022-07-07
[loki-distributed] Add optional appProtocol to memcached services

jdgeisler:
This PR addresses the same issue that was resolved in https://github.com/grafana/helm-charts/pull/1549, but for memcached instead.

With istio protocol selection, loki components are unable to communicate with the memcached services and throw the following error: 
```
level=error ts=2022-07-07T18:25:48.762006363Z caller=memcached.go:144 org_id=fake msg="Failed to get keys from memcached" keysrequested=1 err="memcache: unexpected line in get response: \"HTTP/1.1 400 Bad Request\\r\\n\""
level=error ts=2022-07-07T18:25:48.949472733Z caller=memcached.go:220 msg="failed to put to memcached" name=chunks err=null
level=warn ts=2022-07-07T18:25:48.949510324Z caller=background.go:127 msg="backgroundCache writeBackLoop Cache.Store fail" err="server=100.80.71.54:11211: memcache: unexpected response line from \"set\": \"HTTP/1.1 400 Bad Request\\r\\n\""
```

This can be resolved by setting `appProtocol: tcp` in the memcached services.  If appProtocol is left empty, then it will be ignored and not included in the service.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.6.1
-version: 0.53.3
+version: 0.54.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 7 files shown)


MasslessParticle (approved):
This looks totally reasonable :+1:

PR #1583, opened 2022-07-07
[loki-simple-scalable] Update ssd monitoring

trevorwhitney:
This PR updates the built-in dashboards to match the latest from loki after the work to make SSD specific dashboards was complete. It also add alerting and recording rules.


trevorwhitney:
I couldn't find any better way to be able to enable the rules by default other than copy/pasting the required crds from the `kube-prometheus-stack` chart. That would be a heavy dependency to bring in, so unless there are plans to pull those CRDs out into their own chart, I'm not sure of a better solution for now?

dannykopping (approved):

Issue #838, opened 2021-11-22
tempo-distributed limitations on config entries


warrenhodg:
Using the current config system, there are several config entries that cannot easily be tuned, and that are probably quite commonly needed:
* trace_idle_period
* flush_check_period
* querier->frontend_worker->grpc_client_config->max_send_size
* query_shards
* storage->pool->queue_depth
* max_traces_per_user
* max_bytes_per_trace

bueti:
Has anybody from Grafana looked at this? Would you be open for a PR?

rlex:
i added support for structuredConfig (like in loki and mimir) to tempo-distributed: https://github.com/grafana/helm-charts/pull/1591

As soon as it will be merged it should be easy to customize any variable you want

PR #1591, opened 2022-07-08
[tempo-distributed] structuredConfig support

rlex:
Add tempo structuredconfig override

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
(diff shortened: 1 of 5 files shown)


rlex:
Looks like this chart is ignored in helm-docs, so i temporary removed it from ignores, re-generated it and manually picked some changes. Should be fine i guess?

rlex:
should close https://github.com/grafana/helm-charts/issues/838

rlex:
added some info to README as well with examples

joe-elliott (approved):
Nice improvement, thanks!

I'll dig into why we are currently in `.helmdocsignore`. I had recently noticed that it was no longer requiring regenning the readme.md but I hadn't dug into why. Thanks for the heads up.

PR #1619, opened 2022-07-19
[enterprise-logs] updated image version to use GEL 1.5

vlad-diachenko:


MichelHollands (approved):
LGTM

Issue #1375, opened 2022-05-13
[loki-stack] [feature request] scrape config sidecar


cornfeedhobo:
Similar to the grafana chart, it would be nice to have a sidecar that listens for ConfigMaps or Secrets that have a certain label instead of having to add them statically as helm chart values. This makes dependency resolutions much easier.

PR #1625, opened 2022-07-20
[loki] add sidecar to find loki rules in configmaps and secrets

Jeinhaus:
This PR adds a sidecar to the `loki` chart that queries the Kubernetes API to find `Configmaps` and/or `Secrets` that contain `loki` rules, similar to the way the `grafana` chart does it for dashboards, plugins, ....

I'm not quite sure how to handle the combination of this feature with the `useExistingAlertingGroup` and `alertingGroups` features.
All of these have to be mounted to a single folder in the loki container and then the config has to point to that folder.
Do you think documenting this would be sufficient or should a new top-level value like `rulePath` be added (which would probably be a breaking change :disappointed: )?
Any input would be appreciated.

This might also fix https://github.com/grafana/helm-charts/issues/1375.

### DIFF
--- a/charts/loki/Chart.yaml
+++ b/charts/loki/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: loki
-version: 2.13.2
+version: 2.14.0
 appVersion: v2.6.1
 kubeVersion: "^1.10.0-0"
 description: "Loki: like Prometheus, but for logs."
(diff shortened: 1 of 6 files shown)


zanhsieh:
@Jeinhaus 
Since you proposed a minor upgrade, would you mind to add a section under README.md > Upgrade to explain what would potentially breaking change be?

zanhsieh:
@Jeinhaus 
Close PR due to loki chart deprecation https://github.com/grafana/helm-charts/pull/1772 .

Jeinhaus:
Argh, sorry for being so slow. I'll update the chart in the new repository. Thanks for the heads up.

PR #1630, opened 2022-07-20
[loki-simple-scalable] Service Memberlist tcp prefix fix

rgarcia6520:
We had made this change to our helm-charts we re-host a while ago and upon upgrade to latest version we left the port name as `http` and encountered nothing but issues, think a PR is necessary since this breaks istio injected loki-simple-scalable deployments.

For issue https://github.com/grafana/helm-charts/issues/1099


zanhsieh (approved):

majid-akbari:
This `appProtocol` is missing for new `backend` target and it causing issues with service mesh.

PR #1647, opened 2022-07-28
[tempo-distributed] Include PodDisruptionBudget for the Ingester

zalegrala:
This change amends the podManagementPolicy that was introduced in #1057 to include a PodDisruptionBudget.  This will help ensure that the Ingesters won't take an outage during a deployment.

### DIFF
--- a/charts/tempo-distributed/templates/ingester/pdb.yaml
+++ b/charts/tempo-distributed/templates/ingester/pdb.yaml
@@ -1,0 +1,14 @@
+{{- if .Values.podDisruptionBudget -}}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "tempo.ingesterFullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "tempo.ingesterLabels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ include "tempo.ingesterFullname" . }}
+{{ toYaml .Values.podDisruptionBudget | indent 2 }}
+{{- end }}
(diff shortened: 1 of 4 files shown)


osg-grafana:
cc @knylander-grafana

knylander-grafana (commented):
Thank you for the pointer, Ursula.

joe-elliott:
I think we do this on purpose for quick scale ups.

Wouldn't a "PodDisruptionBudget" cover what you're concerned about?

zalegrala:
I think you are right @joe-elliott, this was pointed out by a few folks but I have't had time to test.  We probably need the PDB in our jsonnet also.

zalegrala:
I've made an update to include the PDB.  Testing locally yields the following.
```
❯ k describe pdb chart-tempo-distributed-ingester
Name:             chart-tempo-distributed-ingester
Namespace:        default
Max unavailable:  1
Selector:         app.kubernetes.io/component=ingester,app.kubernetes.io/instance=chart,app.kubernetes.io/name=tempo-distributed
Status:
    Allowed disruptions:  1
    Current:              3
    Desired:              2
    Total:                3
Events:
  Type    Reason  Age                From               Message
  ----    ------  ----               ----               -------
  Normal  NoPods  37m (x2 over 37m)  controllermanager  No matching pods found
```

charts/tempo-distributed/templates/ingester/pdb.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if .Values.podDisruptionBudget -}}
+apiVersion: policy/v1beta1
```
joe-elliott, line 2: `policy/v1` ?

joe-elliott (approved):

PR #1651, opened 2022-08-01
[grafana] bump Grafana version to 9.0.5

cedi:


charts/grafana/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
 version: 6.32.9
```
mhoyer, line 3: I guess the chart version also needs to be changed here as well.

```diff
- version: 6.32.9
+ version: 6.32.10
```

zanhsieh (approved):

Issue #1653, opened 2022-08-02
[tempo-distributed] PDB is too strict for ingester


AlexDHoffer:
When going in to upgrade the Kubernetes cluster that hosts our Tempo Ingester, I realized that the PDB is set to the very strict maxUnavailable: 1. 

In my case, I have a big set of Ingesters and each has requests and limits for memory/cpu such that they can be a bit of a pain to schedule onto nodes, making the upgrade take a very long time. I tend to believe other people would also have more advanced requirements for the Ingester since it does a lot of the heavy lifting in Tempo.

I propose we set the Ingester PDB to instead be ``minAvailable: {{ .Values.ingester.config.replication_factor }}``. Or, if this is too risky, I'm happy to switch it to something more conservative. 1 replica, however, is too extreme in my opinion.

PR #1654, opened 2022-08-02
[tempo-distributed] Allow higher count of unavailable ingester replicas in pdb

AlexDHoffer:
Resolves https://github.com/grafana/helm-charts/issues/1653

Happy to make this more conservative, just interested in increasing the count from 1


joe-elliott:
With RF3 the system can tolerate being down 1 ingester and still accept writes and return reads. Now, depending on how you do your rollouts you could do more than one at once, but it would potentially increase load on the remaining ingesters quite a bit.

AlexDHoffer:
@joe-elliott I have a lot more than 3 ingester replicas in my ring. My understanding thus is that I can afford to lose more than one and the ring can still service read/writes.

zanhsieh (approved):

PR #1656, opened 2022-08-02
[enterprise-logs] Support configuring `containerSecurityContext` for compactor and gateway targets in the enterprise-logs chart

JordanRushing:
Signed-off-by: JordanRushing <rushing.jordan@gmail.com>

### DIFF
--- a/charts/enterprise-logs/CHANGELOG.md
+++ b/charts/enterprise-logs/CHANGELOG.md
@@ -11,6 +11,10 @@ Entries should be ordered as follows:
 
 Entries should include a reference to the pull request that introduced the change.
 
+## 2.3.1
+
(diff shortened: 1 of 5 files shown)


simonswine (approved):
I haven't actually run the compactor and verified it works, but in general those changes look good to me.

I think we should consider renaming `securityContext` to `podSecurityContext` and also adding `containerSecurityContext` to `tokengen`

trevorwhitney (approved):
LGTM

+1 to @simonswine's suggestions. The current `securityContext` field is applied on the pod level, and since we now have 2 we should rename that to `podSecurityContext`, which matches the pattern in `loki-distributed`

I'm less concerned about tokengen since no one's asked for it, but I guess for consistency it's best.

Thanks Jordan!

zanhsieh (approved):

PR #1663, opened 2022-08-04
[promtail] Remove unnecessary cluster role and its binding

### LINKED ISSUE #1662: [promtail] unnecessary permissions granted to service account
This [clusterrole](https://github.com/grafana/helm-charts/blob/main/charts/promtail/templates/clusterrole.yaml) is created as part of chart.
But there is no component in promtail that talk to k8s control plane, so why does this role even exist?

rkosegi:
Promtail does not talk to k8s control plane, so there is no reason for any RBAC

Closes #1662 

Signed-off-by: Richard Kosegi <richard.kosegi@gmail.com>


zanhsieh:
@rkosegi 
Could you take a look why the readiness return 500 please?

rkosegi:
not needed

PR #1671, opened 2022-08-07
[tempo] optional query volume

### LINKED ISSUE #1670: [tempo] query conf volume is created even when tempoQuery.enabled is set to false
Hello, 

In the statefulset template, the `query-conf` volume is always created, even if the param `tempoQuery.enabled` is set to false, the relevant section is here - https://github.com/grafana/helm-charts/blob/tempo-0.15.7/charts/tempo/templates/statefulset.yaml#L139-L144

Since the `query configmap` is behind a flag (https://github.com/grafana/helm-charts/blob/tempo-0.15.7/charts/tempo/templates/configmap-tempo-query.yaml#L1), setting the flag to `false` causes the `configmap` to not be created but the `sts` still tries to use that volume

HagaiBarel:
Fixes #1670 

Signed-off-by: Hagai Barel <hagaibarel@gmail.com>

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.15.7
+version: 0.15.8
 appVersion: 1.4.1
 engine: gotpl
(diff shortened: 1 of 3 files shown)


joe-elliott (approved):
Thanks for the improvement!

Issue #1570, opened 2022-07-05
[tempo-distributed] Service port names cannot be customized


kmai:
I'm using Istio and I cannot change port names to be Istio-compliant.

I know https://github.com/grafana/helm-charts/pull/1021 fixed it for the AIO chart but not for the distributed one :)

joe-elliott:
We would gladly take a PR if you want to try your hand at it. I don't know enough about Istio to suggest a work around (if there is one).

zanac1986:
@kmai 
Could you review this PR ^^^ ?

PR #1681, opened 2022-08-11
[tempo-distributed] specify grpc appProtocol for tempo distributor se…

zanac1986:
…rvice

See:
https://github.com/grafana/helm-charts/issues/1570
https://github.com/grafana/helm-charts/pull/1021


joe-elliott (commented):
I like this change (as well as some of the other cleanup), but this has a chance of breaking other people's deployments?

We need to list this change in the readme under the "upgrading" section and should include the  rationale: we are changing this to be istio compliant.

zanac1986:
> I like this change (as well as some of the other cleanup), but this has a chance of breaking other people's deployments?
> 
> We need to list this change in the readme under the "upgrading" section and should include the rationale: we are changing this to be istio compliant.

@joe-elliott 
Updated. Please check.

joe-elliott (approved):

joe-elliott:
Thanks!

PR #1689, opened 2022-08-15
[grafana] Make grafana server domain default to ingress host

### LINKED ISSUE #1685: [grafana] Should ingress hostname also be set in grafan.ini
At the moment if you set up ingress on the grafana chart and set the hostname you have to repeat the hostname in the `grafana.ini` `server` `domain` to get things to work properly (redirects when adding new charts to dashboards).

The README.md does give a hint to this, but it's implied that it is only needed when looking to run grafana on a subpath rather than at the root of the hostname.

Would you accept a PR to default the `grafana.ini` value to the first entry in the `ingress.hosts` array or if not a README.md update to make it clear you need to set both to make things work properly?

hardillb:
Fixes #1685

At the moment if you enable ingress on the Grafana chart you also have to remember to specify the hostname again in the grafana.ini file to make sure redirects work properly. This update defaults the ini value to the first hostname in the ingress object to save repeating information.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.32.13
+version: 6.32.15
 appVersion: 9.0.5
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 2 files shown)


Xtigyro (dismissed):
Thank you for the contribution, @hardillb! 🥇 

Could you please bump the minor version of the chart too?

hardillb:
@Xtigyro just checking should I be aiming for version: 6.32.15 (current looks to be version: 6.32.14)?

hardillb:
I have no idea how to resolve the DCO requirement when I fixed the merge conflict

zanhsieh (approved):

Issue #1695, opened 2022-08-17
[loki-simple-scalable] server block in the config value should be configurable


cebidhem:
Hey there,

I'm facing an issue with Loki when some of our users try to fetch logs within Grafana. Grafana is timing out and from loki we can see those logs:
```
"rpc error: code = ResourceExhausted desc = grpc: received message larger than max (4567615 vs. 4194304)"
```
This is somthing we can configure within the server block of the config as described here https://grafana.com/docs/loki/latest/configuration/#server but unfortunately, in the loki-simple-scalable chart that not something we can do:
```
server:
      http_listen_port: 3100
      grpc_listen_port: 9095
```
The solution has been documented in the loki repo here: https://github.com/grafana/loki/issues/2271

I have a PR ready for this. I think it would be useful to allow users to change defaults.

PR #1696, opened 2022-08-17
[loki-simple-scalable] Allow loki server block to be configured

cebidhem:
Fixes https://github.com/grafana/helm-charts/issues/1695

- This will allow users of the chart to configure the server block as documented here https://grafana.com/docs/loki/latest/configuration/#server through values.

Signed-off-by: cebidhem <cebidhem@pm.me>

### DIFF
--- a/charts/loki-simple-scalable/values.yaml
+++ b/charts/loki-simple-scalable/values.yaml
@@ -69,6 +69,9 @@ loki:
     server:
       http_listen_port: 3100
       grpc_listen_port: 9095
+      {{- if .Values.loki.extraServerConfig }}
+      {{- toYaml .Values.loki.extraServerConfig | nindent 2 }}
+      {{- end }}
 
     memberlist:
       join_members:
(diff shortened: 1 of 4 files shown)


zanhsieh (approved):

cebidhem:
Hi @zanhsieh,

I see you approved the MR but it still needs a review. Would you be able to review it so we can merge this ? 
It is blocking us from solving an issue on our side and it would be much appreciated to be able to use this quite quickly.

Thanks in advance!

charts/loki-simple-scalable/values.yaml
```diff
@@ -69,6 +69,9 @@ loki:
     server:
       http_listen_port: 3100
       grpc_listen_port: 9095
+      {{- if .Values.loki.extraServerConfig }}
+      {{- toYaml .Values.loki.extraServerConfig | nindent 2 }}
+      {{- end }}
```
zanhsieh, line 74: I would say `with` statement should be better:

```suggestion
      {{- with .Values.loki.extraServerConfig }}
      {{- toYaml . | nindent 2 }}
      {{- end }}
```

The difference between `if` and `with` is while given `{}`, `if` will render as `{}`, whereas `with` will skip. You could try comparing these two with `helm template lokiss .`

cebidhem: Indeed, thanks for noting this @zanhsieh . Changes pushed!

PR #1703, opened 2022-08-18
[loki-distributed] add support for in-memory disk in ingester and indexGateway

rlex:
Allows to use emptyDir in ingester and indexGateway with medium: Memory. Useful in certain cases.

Allows setting limit as well, uses same variable as ordinary persistence block.

For ingester, supports both Deployment and StatefulSet.


rlex:
added warning similar to one requested in https://github.com/grafana/helm-charts/pull/1705

rlex:
bumped version and rebased

zanhsieh (approved):

PR #1705, opened 2022-08-18
[tempo-distributed] allow using inmemory emptyDir with tempo-ingester

rlex:
same as https://github.com/grafana/helm-charts/pull/1703 but for tempo.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.24.0
+version: 0.24.1
 appVersion: 1.5.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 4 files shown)


BitProcessor:
1. what is the actual use case ? 
2. it is a bit weird to call it persistent when it is stored in-mem ? 
3. and this is more a question for @joe-elliott ; given persistence and in-mem are quite contradictory, do we want to support this scenario ? A simple container restart and the persistence is lost. Or when using cluster-autoscaler for example, pods can get moved around automatically to other nodes. Persistence: gone.

rlex:
1. Nodes with huge amount of RAM and small/slow disks. Edge case, probably, but i'm fully aware of all downsides.
2. Yep, probably, i took it from Grafana chart to use same format: https://github.com/grafana/helm-charts/blob/cecb957eb264be8f8952dac7678d757b0e695092/charts/grafana/values.yaml#L307
3. Yep, but it's also pretty obvious that ramdisk is not persistent. I'm open to naming suggestions.

BitProcessor:
If we do go forward with it, we probably want to put a huge warning in the config file. 
Seems like a valid edge-case though, thanks for pointing that out.

btw: obviousness is not a shared thing in general ;)

rlex:
So should i just add bold warning to value description that all data will be lost at restart in case of inMemory: true?

BitProcessor:
> So should i just add bold warning to value description that all data will be lost at restart in case of inMemory: true?

Yes please: ⚠️⚠️⚠️⚠️

rlex:
That should do?

BitProcessor:
@joe-elliott went ahead with this one, as it DOES seem like a valid edge case. The big fat warning should do. 
@rlex thanks for another contribution.

BitProcessor (approved):

Issue #1574, opened 2022-07-05
[loki-distributed] Helm Chart has no option to deploy a query-scheduler for Loki distributed


sudo6si:
Hi,

A separate query scheduler service improves Loki's query performance. Currently, the Helm chart has no option to enable a separate query scheduler. There's a similar issue+PR for [mimir](https://github.com/grafana/mimir/pull/2087) which supports a separate query scheduler.

I would be happy to submit a PR to resolve this issue. But before I begin I wanted to know if this is something that is within the scope of the helm chart for loki-distributed.

Please let me know if it's a yes and I will go ahead with creating the PR and testing it.

Thanks,
-sudo

caibirdme:
Also need query scheduler feature. BTW, I didn't find any examples about how to configure it.

salvacorts:
Closing this one as it has been addressed by https://github.com/grafana/helm-charts/pull/1719

PR #1719, opened 2022-08-22
[loki-distributed] Feature: Deploy query-scheduler as an independent component

Raskyld:
### Rationale

Linked to #1574, the recommended approach for the _read-path_ is to deploy the `query-scheduler` as an independent component which allows us to scale `query-frontend` regardless of restriction imposed by `-querier.max-concurrent`.
If you want to learn more about this limitation I invite you to check the [Why Query Frontend Scalability is Limited](https://grafana.com/docs/mimir/latest/operators-guide/architecture/components/query-frontend/#why-query-frontend-scalability-is-limited) article from _Mimir_.

Among other benefits, it allows users to benefit from more metrics in the frontend as well as metrics from the scheduler components; which, in turn, allow users to follow the new [operational guide to auto-scale the queriers](https://github.com/grafana/loki/pull/6801) by @salvacorts.

#### Should it be the default?

I think the independent `query-scheduler` setup should be the default. I chose to left it as an "opt-in" (`query-scheduler.enabled` is `false` by default) because it can mess with current installations (breaking change). If you think the BC is worth it to promote the usage of the `query-scheduler` we should announce a breaking change.

#### Should we let users configure memberlist over DNS resolution?

By default, when both `scheduler_address` and `frontend_address` are undefined, components default to using memberlist to resolve schedulers. From my codebase crawling it looks like the DNS resolution is the recommended approach ATM (even though I don't know why) so my PR does not support memberlist option.
Should we support it? If yes, in this PR or in another to keep things atomic?

### Tests made

I released the chart on a lab, here is what I looked for:

- `query-schedulers` are deployed when enabled is `true`,
- The service is headless so DNS SRV records are created,
- The DNS resolution from queriers and frontend is working,
- frontend, querier and scheduler logs look fine,
- I can query my cluster.

### DIFF
--- a/charts/loki-distributed/templates/query-scheduler/deployment-query-scheduler.yaml
+++ b/charts/loki-distributed/templates/query-scheduler/deployment-query-scheduler.yaml
@@ -1,0 +1,116 @@
+{{- if .Values.queryScheduler.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "loki.querySchedulerFullname" . }}
+  labels:
+    {{- include "loki.querySchedulerLabels" . | nindent 4 }}
+  {{- with .Values.loki.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  replicas: {{ .Values.queryScheduler.replicas }}
+  strategy:
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+  revisionHistoryLimit: {{ .Values.loki.revisionHistoryLimit }}
+  selector:
+    matchLabels:
+      {{- include "loki.querySchedulerSelectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print .Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- with .Values.loki.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+        {{- with .Values.queryScheduler.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      labels:
+        {{- include "loki.querySchedulerSelectorLabels" . | nindent 8 }}
+        {{- with .Values.loki.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+        {{- with .Values.queryScheduler.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      serviceAccountName: {{ include "loki.serviceAccountName" . }}
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- include "loki.querySchedulerPriorityClassName" . | nindent 6 }}
+      securityContext:
+        {{- toYaml .Values.loki.podSecurityContext | nindent 8 }}
+      terminationGracePeriodSeconds: {{ .Values.queryScheduler.terminationGracePeriodSeconds }}
+      containers:
+        - name: query-scheduler
+          image: {{ include "loki.querySchedulerImage" . }}
+          imagePullPolicy: {{ .Values.loki.image.pullPolicy }}
+          args:
+            - -config.file=/etc/loki/config/config.yaml
+            - -target=query-scheduler
+            {{- with .Values.queryScheduler.extraArgs }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          ports:
+            - name: http
+              containerPort: 3100
+              protocol: TCP
+            - name: grpc
+              containerPort: 9095
+              protocol: TCP
+          {{- with .Values.queryScheduler.extraEnv }}
+          env:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          {{- with .Values.queryScheduler.extraEnvFrom }}
+          envFrom:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          securityContext:
+            {{- toYaml .Values.loki.containerSecurityContext | nindent 12 }}
+          readinessProbe:
+            {{- toYaml .Values.loki.readinessProbe | nindent 12 }}
+          livenessProbe:
+            {{- toYaml .Values.loki.livenessProbe | nindent 12 }}
+          volumeMounts:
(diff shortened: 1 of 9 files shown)


charts/loki-distributed/templates/query-scheduler/deployment-query-scheduler.yaml
```diff
@@ -0,0 +1,116 @@
...
+          livenessProbe:
+            {{- toYaml .Values.loki.livenessProbe | nindent 12 }}
+          volumeMounts:
+            - name: config
+              mountPath: /etc/loki/config
+            {{- with .Values.queryScheduler.extraVolumeMounts }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          resources:
+            {{- toYaml .Values.queryScheduler.resources | nindent 12 }}
```
zanhsieh, line 88: Seems better with `with` since by default `resource` is `{}` in values.yaml

```suggestion
          {{- with .Values.queryScheduler.resources }}
          resources:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

charts/loki-distributed/templates/query-scheduler/service-query-scheduler.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  type: ClusterIP
+  clusterIP: None
+  publishNotReadyAddresses: true
+  ports:
+    - name: http
+      port: 3100
+      targetPort: http
+      protocol: TCP
+    - name: grpclb
```
zanhsieh, line 24: would it be better to istio friendly?

```suggestion
    - name: lb-grpc
```

zanhsieh (commented):
overall good. just need some minor polish.

charts/loki-distributed/templates/query-scheduler/deployment-query-scheduler.yaml
Raskyld: Good catch! Thanks for the feedback!

charts/loki-distributed/templates/query-scheduler/service-query-scheduler.yaml
Raskyld: I tried to stay consistent with [ports name when using Tanka deployment](https://github.com/grafana/loki/blob/main/production/ksonnet/loki/common.libsonnet#L29). So I don't know if we should diverge from it or not, maybe someone has the rationale behind this name?

Raskyld: It was also used in the [frontend](https://github.com/grafana/helm-charts/blob/cb6f01bc3bd2ec7ce35880c021942d2c38b6c037/charts/loki-distributed/templates/query-frontend/service-query-frontend.yaml#L27), but ye I don't really know the reason behind the name, think it's related to the fact it resolves to a DNS SRV record?

Raskyld:
I will need to rebase and apply the changes myself @zanhsieh to add the "signed-off" mention to the commits

salvacorts (approved):
Thanks for putting this together!

charts/loki-distributed/values.yaml
```diff
@@ -554,6 +561,66 @@ queryFrontend:
   # -- Tolerations for query-frontend pods
   tolerations: []
 
+# Configuration for the query-scheduler
+queryScheduler:
+  # -- Specifies whether the query-scheduler should be decoupled from the query-frontend
+  enabled: false
```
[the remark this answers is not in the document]
Raskyld: True, it's highly desirable to run with the scheduler as it's the maintained approach, but if we release this without announcing a BC, one may update the chart to the new version and end up with the `query-scheduler` now being deployed without realizing it.

While it seems harmless, there is the budget and/or quota question, the user will effectively be consuming more resources.

Then, there is the question around `query-frontend` different implementations. When not using the `query-scheduler`, the `query-frontend` run a specific implementation called `v1`, this implementation exposes some metrics (see [there](https://github.com/grafana/loki/blob/main/pkg/lokifrontend/frontend/v1/frontend.go#L83)); when using the scheduler independently, the `query-frontend` implementation change under the hood for the one called `v2` which exposes different metrics. So if an user configured alerts based on the `v1` metrics, he will break those alerts as the metrics will disappear when upgrading to `v2` (which is seamless for him, hence, uncontrolled as it's a side-effect of using the scheduler). There may be other incompatibilities between `v1` and `v2` implementations that I'm unaware of.

Moreover, users must be careful in the configuration. There are some operational rules that are not enforced but should be well-known: When `-querier.worker-match-max-concurrent` is enabled, _managers_ responsible for fetching queries from the schedulers will try to fairly balance among available scheduler replicas to match the constraint `-querier.max-concurrent` defining how much workers queriers have. Let me give you an explanation: you have 2 scheduler replicas and 6 workers, we will pop 6 managers to ensure a One-To-One parity between workers and managers in order to ensure there is no back-pressure (managers don't accept more queries than worker actually can run). Then the fair balance apply : We have 6 managers and 2 replicas, so we will allocate `6/2 = 3` manager to each scheduler replica.
Alright, but what happens when the evenly distribution can't happen (e.g. 7 managers but 2 replicas) ? Well, the distribution will be unbalanced with some replicas being more requested than others. If you are unlucky and you fall into the worst case scenario which is all your queriers favorite the same replicas.. Well your workload is not balanced. Note that the default config (10 workers and 2 replicas) ensure the rules is satisfied but if it happens that the user changed the workers count to an odd number.. well, you get it!

You can even override all of it by not enabling the quoted option. The result is for each scheduler replicas, we pop `-querier.parallelism` managers (this option is ignored when the match max concurrent option is enabled).

But there is more! If you are not careful and you reach the point where the 1:1 parity between managers and worker is broken (typically when `queryScheduler.replicas > -querier.max-concurrent` as each replicas is ensured to have at least one manager allocated to it) you will start accepting queries you can't treat immediately.
If we take our last example, and turn off the fair balancing, the default `-querier.parallelism` (which is 10) apply. The result is we pop 10 managers for each replicas; in our example we said we have 2 replicas => `2*10 = 20` managers but only `10` workers resulting in `20-10 = 10` query being buffered and not treated instantly. 

**tl; dr The migration from a monolithic frontend to a setup with an independent scheduler is a non-trivial operation that may have various impact.**

Raskyld:
Thanks for reviewing my work! :rocket: :heart:

zanhsieh (approved):

Raskyld:
Fixes #1574 !
Thanks everyone :tada:

sudo6si:
Thanks a lot for this @Raskyld !

Issue #1710, opened 2022-08-19
[grafana] Add provisioning of new unified alerting available in version 9.1


mpostument:
Unified alerting provisioning was added in grafana version 9.1.

https://grafana.com/docs/grafana/latest/administration/provisioning/#alerting

Please add possibility to  provision with helm chart following resources:
 - Rules
 - Contact points
 - Notification policies

booleanbetrayal:
This has become a big headache for us as we introduce new Alerts in our provisioned dashboards. We've been generating these dashboards for some time now using IaC approaches where the dashboard templating is versioned in git. In the legacy system, introducing a new Alert was as simple as introducing a new rule in Panel code, deploying it, and allowing the provisioned dashboards sidecar to pick up and reflect those changes reliably.

After adopting Unified Alerts, we've noticed that these updates do not trigger subsequent migration passes over dashboards with introduced changes and therefore no new Alerts are being reflected after being deployed. We've found ourselves conducting force rollbacks of the Grafana DB to move temporarily back to Legacy Alerts, then re-migrate with the new Alert discovered. At that point we are back on the Unified Alert system with a rule defined in an IaC paradigm, but the journey is complex and also requires rebuilding notification templates, not to mention this migration scheme will probably be deprecated at some point.

It looks like this new provisioning support can solve this problem. Getting first-class support for this feature in Helm would be a huge quality of life improvement.

andrew-gropyus:
In addition to #1720 which adds support for adding alerts by customising the chart values.yaml, would it make sense to also include an approach similar to the configmap sidecar listener that is being used for dashboards, datasources and notifiers?

vsinha:
Is there a method for loading envFromSecret into the alerting sidecar? For things like slack tokens and whatnot

Karsten-Fyhn:
Is it correctly understood that #1720 handles rules and contact points, but that notification policies still cannot be configured via the Helm charts? If so, is that on the roadmap?

tarikdem:
> Is there a method for loading envFromSecret into the alerting sidecar? For things like slack tokens and whatnot

The sidecar approach already enables provisioning contact points from secrets, so that can be used for slack tokens.

> Is it correctly understood that #1720 handles rules and contact points, but that notification policies still cannot be configured via the Helm charts? If so, is that on the roadmap?

[...]

PR #1720, opened 2022-08-23
[grafana] Add support for alerting provisioning present in grafana 9.1

mmusenbr:
This should allow provisioning of files via the [grafana provisioning support](https://grafana.com/docs/grafana/latest/administration/provisioning/#alerting) and should address #1710.
This will only work when using the chart with Grafana 9.1

Example usage:
```yaml
alerting:
  rules.yaml:
    apiVersion: 1
    groups:
    ....
  contactpoints.yaml:
    apiVersion: 1
    contactPoints:
    ....
```

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.33.2
+version: 6.34.0
 appVersion: 9.1.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/configmap.yaml
+++ b/charts/grafana/templates/configmap.yaml
@@ -56,6 +56,13 @@ data:
   {{- end -}}
 {{- end -}}
 
+{{- if .Values.alerting }}
+  {{- range $key, $value := .Values.alerting }}
+  {{ $key }}: |
+{{ toYaml $value | indent 4 }}
+  {{- end -}}
+{{- end -}}
+
 {{- if .Values.dashboardProviders }}
   {{- range $key, $value := .Values.dashboardProviders }}
   {{ $key }}: |
--- a/charts/grafana/templates/configmap.yaml
+++ b/charts/grafana/templates/configmap.yaml
@@ -57,9 +57,10 @@ data:
 {{- end -}}
 
 {{- if .Values.alerting }}
+{{ $root := . }}
   {{- range $key, $value := .Values.alerting }}
   {{ $key }}: |
-{{ toYaml $value | indent 4 }}
+{{ tpl $value $root | indent 4 }}
   {{- end -}}
 {{- end -}}
 
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -504,6 +504,29 @@ datasources: {}
 #        authType: default
 #        defaultRegion: us-east-1
 
+## Configure grafana alerting
+## ref: http://docs.grafana.org/administration/provisioning/#alerting
+##
+alerting: {}
+#  rules.yaml:
+#    apiVersion: 1
+#    groups:
+#    - orgId: 1
+#      name: my_rule_group
+#      folder: my_first_folder
+#      interval: 60s
+#      rules:
+#      - uid: my_id_1
+#        ...
+#  contactpoints.yaml:
+#    apiVersion: 1
+#    contactPoints:
+#    - orgId: 1
+#      name: cp_1
+#      receivers:
+#      - uid: first_uid
+#        ...
+
 ## Configure notifiers
 ## ref: http://docs.grafana.org/administration/provisioning/#alert-notification-channels
 ##
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -504,28 +504,70 @@ datasources: {}
 #        authType: default
 #        defaultRegion: us-east-1
 
-## Configure grafana alerting
+## Configure grafana alerting (can be templated)
 ## ref: http://docs.grafana.org/administration/provisioning/#alerting
 ##
 alerting: {}
-#  rules.yaml:
+#  rules.yaml: |
 #    apiVersion: 1
 #    groups:
-#    - orgId: 1
-#      name: my_rule_group
-#      folder: my_first_folder
-#      interval: 60s
-#      rules:
-#      - uid: my_id_1
-#        ...
-#  contactpoints.yaml:
+#      - orgId: 1
+#        name: {{ .Chart.Name }}_my_rule_group
+#        folder: my_first_folder
+#        interval: 60s
+#        rules:
+#          - uid: my_id_1
+#            title: my_first_rule
+#            condition: A
+#            data:
+#              - refId: A
+#                datasourceUid: '-100'
+#                model:
+#                  conditions:
+#                    - evaluator:
+#                        params:
+#                          - 3
+#                        type: gt
+#                      operator:
+#                        type: and
+#                      query:
+#                        params:
+#                          - A
+#                      reducer:
+#                        type: last
+#                      type: query
+#                  datasource:
+#                    type: __expr__
+#                    uid: '-100'
+#                  expression: 1==0
+#                  intervalMs: 1000
+#                  maxDataPoints: 43200
+#                  refId: A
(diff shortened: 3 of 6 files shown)


Xtigyro (changes requested):
Thanks for the contribution, @mmusenbr! 🥇

Could you please bump the minor (not the bugfix) chart version in `Chart.yaml`?

mmusenbr:
> Could you please bump the minor (not the bugfix) chart version in `Chart.yaml`?

@Xtigyro sure, just pushed it with 6.33.0

charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.32.18
+version: 6.33.0
 appVersion: 9.0.5
```
pdambrauskas, line 4: Maybe app version should be bumped to 9.1.1, since Unified provisioning is supported only from Grafana 9.1.0?

mmusenbr: Even though I use it too with 9.1, i did not feel I am qualified to push the major version within this PR. I assume further compatibility testing is done by the maintainers for such upgrade.

For users this can simply be overwritten by defining a custom values.yaml with:
```yaml
image:
  tags: "9.1.1"
  ```

**Comment by zanhsieh** [comment]
@mmusenbr 
Do you mind to push your example into values.yaml with comment? For example,

```yaml
...
alerting: {}
#   rules.yaml: |
#      apiVersion: 1
#      groups:
#      ....
#   contactpoints.yaml: |
#     apiVersion: 1
#     contactPoints:
#     ....
```

zanhsieh (changes requested):
@mmusenbr
Do you mind to push your example into values.yaml with comment? For example,

...
alerting: {}
#   rules.yaml: |
#      apiVersion: 1
#      groups:
#      ....
#   contactpoints.yaml: |
#     apiVersion: 1
#     contactPoints:
#     ....
I don't expect ppl would know how to config simply just look into the code.

charts/grafana/templates/configmap.yaml
```diff
@@ -56,6 +56,13 @@ data:
   {{- end -}}
 {{- end -}}
 
+{{- if .Values.alerting }}
+  {{- range $key, $value := .Values.alerting }}
+  {{ $key }}: |
+{{ toYaml $value | indent 4 }}
```
zanhsieh, line 62: I don't think this would work if the alerting sub yaml files contains self-reference values, e.g. host, port, etc. 

Maybe try this:

```suggestion
{{- tpl $value  . | fromYaml | nindent 4 }}
```

zanhsieh (changes requested):

charts/grafana/templates/configmap.yaml
mmusenbr: I will have a look into that

mmusenbr:
Rebased to the latest main-branch, added example section in `values.yaml`

charts/grafana/templates/configmap.yaml
mmusenbr: @zanhsieh I had a look and tried it locally. As using `{{ ... }}` is quite common inside the deployed files (eg as `legendFormat: "{{ pod }}"` or as contact points template) a lot of 'escaping' and rewrite in form of ``legendFormat: "{{` {{ pod }} `}}"`` would be necessary.

I think this will increase config and maintenance work for the user of the alerting section, by adding this feature.
How should we proceed here?

zanhsieh: @mmusenbr 
Did you try / consider tearing down alerts yamls to separate files? See:
https://helm.sh/docs/chart_template_guide/accessing_files/

Xtigyro (approved):
Super nice - thank you, @mmusenbr!

charts/grafana/templates/configmap.yaml
zanhsieh: @mmusenbr 
I made some update to my previous suggestion as well as values.yaml and hope to shed some light. Please take a look.

zanhsieh: > 

Maybe raw string constant?

https://stackoverflow.com/questions/51326021/escape-curly-brackets-in-templates

zanhsieh: > @zanhsieh I had a look and tried it locally. As using `{{ ... }}` is quite common inside the deployed files (eg as `legendFormat: "{{ pod }}"` or as contact points template) a lot of 'escaping' and rewrite in form of `` legendFormat: "{{` {{ pod }} `}}" `` would be necessary.
> 
> I think this will increase config and maintenance work for the user of the alerting section, by adding this feature. How should we proceed here?

I think we could do raw string.
https://stackoverflow.com/a/51326166/324275

mmusenbr: @zanhsieh no worries, escaping does not make me uncomfortable, I was just thinking/assuming that a common use-case is to export existing config and provisioning it. And the whole escaping needs to be done and wanted to bring that up. Both ways are fine for me personally.

I will check your latest suggestion and then push it again.

charts/grafana/values.yaml
```diff
@@ -504,6 +504,71 @@ datasources: {}
...
 
+## Configure grafana alerting (can be templated)
+## ref: http://docs.grafana.org/administration/provisioning/#alerting
+##
+alerting: {}
+#  rules.yaml: |
+#    apiVersion: 1
+#    groups:
+#      - orgId: 1
+#        name: my_rule_group
```
zanhsieh, line 515: We should demo the user the yaml we passed in could be self-referable, otherwise the user would question why they need to raw string the ``{{  `{{ ... }}` }}``:

```suggestion
#        name: {{ .Chart.Name }}_my_rule_group
```

zanhsieh (changes requested):

charts/grafana/templates/configmap.yaml
```diff
@@ -56,6 +56,14 @@ data:
   {{- end -}}
 {{- end -}}
 
+{{- if .Values.alerting }}
+{{ $root := . }}
+  {{- range $key, $value := .Values.alerting }}
+  {{ $key }}: |
+{{ tpl (toYaml $value | indent 4) $root }}
```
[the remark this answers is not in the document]
mmusenbr: Sorry, I was not finished testing this. I now have: `{{- tpl (toYaml $value) $root | nindent 4 }}` I still need to use the 'toYaml' as tpl expects a string (at least what I saw on testing and read).
And this produces now:
```yaml
data:
  ...
  alerts.yaml: |
    apiVersion: 1
    groups:
    ...
```

charts/grafana/values.yaml
mmusenbr: ok, can add this in the examples.

mmusenbr: adapted

charts/grafana/templates/configmap.yaml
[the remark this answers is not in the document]
zanhsieh: @mmusenbr 
- `|` will be detected $value as a string (e.g. `alerts.yaml: |`).
- Using `nindent` will generate extra empty line between alert.yaml and its following content. For example:

``` yaml
  alerts.yaml: |

    apiVersion: 1
    groups:
    ...
```
- Dunno what version of your helm, my helm is `v.3.9.0`.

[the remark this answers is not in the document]
mmusenbr: @zanhsieh 
Ah thanks for the `|` treat as string info. Sheds some light.
Yes, this will generate an extra line, but I had, as at some point suggested, `{{-` in front of the template, which again consumes it from nintend.
So it is either `{{ tpl $value $root | indent 4 }}` or `{{- tpl $value $root | nindent 4 }}`, or? I will push the first version, as it matches the last suggestion I think and is closer to the way it is done in the other parts of the `configmap.yaml`.

And I hope this is the final point, should we use (on which I am, with my currently limited template knowledge not sure if it is the same or not):
```yaml
{{- if .Values.alerting }}
{{ $root := . }}
...
{{ tpl $value $root | indent 4 }}
....
```
or, as suggested at another point:
```yaml
{{- if .Values.alerting }}
...
{{ tpl $value $ | indent 4 }}
....
```

zanhsieh (approved):

PR #1732, opened 2022-08-26
[tempo-distributed] Add log_received_spans option

faustodavid:
The feature https://github.com/grafana/tempo/pull/1465 was added to Tempo v1.5.0 which made `log_received_traces` deprecated and replaced with `log_received_spans` with extra options.


joe-elliott (approved):
Nice addition. Thx again, Fausto.

PR #1747, opened 2022-08-30
[loki-stack] update loki chart dependency

jrockway:
This is mostly an attempt to get a k8s v1.25 compatible install.  I hope it's really this easy, but this PR mostly exists to see if CI passes ;)

The "contributing guidelines" suggest bumping the version number for each change.  I don't know if that's actually the procedure, but I did it.  It seems like there were no breaking changes mentioned in the Loki README since 2.13.0 (i.e. not from 2.14 to 2.15 which this PR updates), so I just did a patch number bump here.

### DIFF
--- a/charts/loki-stack/Chart.yaml
+++ b/charts/loki-stack/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: loki-stack
-version: 2.8.0
+version: 2.8.1
 appVersion: v2.6.1
 kubeVersion: "^1.10.0-0"
 description: "Loki: like Prometheus, but for logs."
--- a/charts/loki-stack/requirements.yaml
+++ b/charts/loki-stack/requirements.yaml
@@ -2,7 +2,7 @@ dependencies:
 - name: "loki"
   condition: loki.enabled
   repository: "https://grafana.github.io/helm-charts"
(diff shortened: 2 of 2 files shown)


charts/loki-stack/Chart.yaml
```diff
@@ -1,14 +1,14 @@
...
 home: https://grafana.com/loki
 icon: https://raw.githubusercontent.com/grafana/loki/master/docs/sources/logo.png
 sources:
-- https://github.com/grafana/loki
+    - https://github.com/grafana/loki
 maintainers:
-- name: Loki Maintainers
-  email: lokiproject@googlegroups.com
+    - name: Loki Maintainers
+      email: lokiproject@googlegroups.com
```
zanhsieh, line 13: Is this change relevant? If not, could you restore it?

zanhsieh (changes requested):

charts/loki-stack/Chart.yaml
jrockway: Not relevant, automation gone wrong.  It's fixed now.

jrockway:
I've repaired the indentation changes.  Please take another look.

zanhsieh (approved):

PR #1754, opened 2022-08-31
[loki-simple-scalable] Support for additional limits_configs

jcputter:
Support for additional limits_configs


zanhsieh:
@jcputter could you make another PR under https://github.com/grafana/loki/tree/main/production/helm/loki ?
loki and loki-simple-scalable moved.

PR #1757, opened 2022-08-31
[loki-stack] Allow specifying UID for Loki datasource

kenske:


charts/loki-stack/templates/datasources.yaml
```diff
@@ -21,6 +21,9 @@ data:
       url: {{ tpl .Values.loki.url . | quote }}
       version: 1
       isDefault: {{ default false .Values.loki.isDefault }}
+      {{- if .Values.loki.datasource.uid }}
+      uid: {{ .Values.loki.datasource.uid | quote }}
+      {{- end }}
```
zanhsieh, line 26: How about:

```suggestion
      {{- with .Values.loki.datasource.uid }}
      uid: {{ . | quote }}
      {{- end }}
```

If uid is empty, it will just get omitted.

zanhsieh (changes requested):

charts/loki-stack/templates/datasources.yaml
kenske: good idea!

zanhsieh (approved):

PR #1759, opened 2022-08-31
[tempo-distributed] refactor and introduce Enterprise Traces support

zalegrala:
With this change, I aim to align the helpers used in the `tempo-distributed` chart with those of the other database teams, namely the `mimir-distributed` chart for improved consistency and long term maintainability.  This is also in prep to implement `enterprise-traces` into the `tempo-distributed` chart.

As part of this alignment, the following structural changes have been implemented.

* import mimir helpers and begin alignment
* centralize image reference handling
* centralize name reference handling
* **BREAKING CHANGE** centralize selector label handling -- users who wish to keep old values should still be able to use the `nameOverride` and `fullNameOverride` top level keys in their values.
* centralize pod label handling
* centralize service monitor around `metamonitoring`
* **BREAKING CHANGE** serviceMonitor has been nested under metaMonitoring -- metamonitoring can be used scrape services as well as install the operator with the following values.  Note also that the port names have changed from `http` to `http-metrics`.
```yaml
metaMonitoring:
  serviceMonitor:
    enabled: true
  grafanaAgent:
    enabled: true
    installOperator: true
```

* minio can now be enabled as part of this chart using the following values
```yaml
minio:
  enabled: true
```

As part of the above changes, many helpers now expect that a context is passed.  This pattern is meant to be used in the following way.

First declare the dictionay for the given component, and optionally set memberlist to true based on the component needs.

```handlebars
{{ $dict := dict "ctx" . "component" "compactor" "memberlist" true }}
```

Then you can pass the `$dict` when rendering a named template.

```handlebars
name: {{ template "tempo.resourceName" $dict }}
```

This allows us to use one template and centralize the logic passing in the
variables that change.

Additionally, the configuration can now be stored in a secret.  See the documentation for `useExternalConfig` and `configStorageType` in the values file for more details.


charts/tempo-distributed/values.yaml
```diff
@@ -1086,43 +1249,226 @@ gateway:
...
+    type: RollingUpdate
+    rollingUpdate:
+      maxSurge: 0
+      maxUnavailable: 1
+
+  podLabels: {}
+  podAnnotations: {}
+
+  nodeSelector: {}
+  affinity: {}
```
krajorama, line 1356: The original services in the same file use a string here and template it with `tpl`, this is a bit inconsistent now

charts/tempo-distributed/values.yaml
```diff
@@ -1086,43 +1249,226 @@ gateway:
...
+      maxUnavailable: 1
+
+  podLabels: {}
+  podAnnotations: {}
+
+  # Pod Disruption Budget
+  podDisruptionBudget: {}
+
+  nodeSelector: {}
+  affinity: {}
```
krajorama, line 1419: The original services in the same file use a string here and template it with tpl, this is a bit inconsistent now

charts/tempo-distributed/values.yaml
```diff
@@ -577,14 +592,57 @@ traces:
...
 
-
 # -- Config file contents for Tempo distributed. Passed through the `tpl` function to allow templating
 
 # @default -- See values.yaml
 config: |
   multitenancy_enabled: {{ .Values.multitenancyEnabled }}
   search_enabled: {{ .Values.search.enabled }}
   metrics_generator_enabled: {{ .Values.metricsGenerator.enabled }}
+
```
krajorama, line 602: Optional: this is subtle stuff, but in mimir-distributed I reordered the yaml in `config` template alphabetically because the processing we do will output the result in an ordered way.

charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -120,3 +161,132 @@ Return if ingress supports pathType.
...
+
+{{/*
+Calculate the config from structured and unstructred text input
+*/}}
+{{- define "tempo.calculatedConfig" -}}
+{{ tpl (mergeOverwrite (tpl .Values.config . | fromYaml) .Values.tempo.structuredConfig | toYaml) . }}
+{{- end -}}
+
+{{/*
+Calculate the overrides config from structured and unstructred text input
```
krajorama, line 180: comment doesn't match what happens in template

charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -120,3 +161,132 @@ Return if ingress supports pathType.
...
+
+{{/*
+Resource name template
+*/}}
+{{- define "tempo.resourceName" -}}
+{{ include "tempo.fullname" .ctx }}{{- if .component -}}-{{ .component }}{{- end -}}
+{{- end -}}
+
+{{/*
+Calculate the config from structured and unstructred text input
```
krajorama, line 173:
```suggestion
Calculate the config from structured and unstructured text input
```

charts/tempo-distributed/templates/configmap-tempo.yaml
```diff
@@ -1,14 +1,27 @@
...
+{{- if eq .Values.configStorageType "Secret" }}
+data:
+  tempo-query.yaml: {{ tpl .Values.queryFrontend.query.config . | b64enc }}
+  overrides.yaml: {{ include "tempo.calculatedOverridesConfig" . | b64enc }}
+  tempo.yaml: {{ include "tempo.calculatedConfig" . | b64enc }}
+{{- else }}
 data:
   tempo-query.yaml: |
     {{- tpl .Values.queryFrontend.query.config . | nindent 4 }}
   overrides.yaml: |
```
krajorama, line 22: This is a bit weird. Don't you reload overrides.yaml automatically in tempo? If it's automatically reloaded from a file, then you shouldn't have it next to the static config, because any change to the overrides will cause the config checksum to change and restart the services for no reason.

charts/tempo-distributed/templates/gateway/deployment-gateway.yaml
```diff
@@ -1,10 +1,11 @@
 {{- if .Values.gateway.enabled }}
```
krajorama, line 1: shouldn't this be
```suggestion
{{- if and .Values.gateway.enabled (not .Values.enterprise.enabled) }}
```
Because now it seems like you run both gateways when enterprise is enabled

krajorama (changes requested):

charts/tempo-distributed/templates/gateway/deployment-gateway.yaml
zalegrala: I don't suppose a user would want to run both, but since both gateways default to false, I thought just switching in the specific component for each one reduced the logic.  Users who want to run only one gateway will need to enable only one gateway in the config.

Alternatively, this could be the kind of thing we put in a validation template, but is not implemented in this pr.

charts/tempo-distributed/values.yaml
zalegrala: A very interesting.

charts/tempo-distributed/templates/configmap-tempo.yaml
zalegrala: I don't believe we reload on config change, but I could be wrong.

Aside from Tempo, are you saying a configmap updates the running pod without restart?  I'd no idea if that's true.  Interesting.  I'd been thinking that the annotations were in place specifically to restart the pod on change, but now I wonder.

charts/tempo-distributed/values.yaml
zalegrala: I've made an additional breaking change to align.  How's that look?

charts/tempo-distributed/values.yaml
zalegrala: I've made an additional breaking change to align.  How's that look?

charts/tempo-distributed/templates/configmap-tempo.yaml
zalegrala: Marty confirmed that we will reload the overrides config.  I'll come back to tidy that up.  Thanks for the pointer.

zalegrala: Perhaps we could follow up in another PR though, what do you think @krajorama ?

charts/tempo-distributed/values.yaml
zalegrala: After some discussion in slack, I expect to roll this back the other direction to template strings for the affinity.

zalegrala: This has been rolled back.  Now we use a template for all affinity.

charts/tempo-distributed/templates/admin-api/admin-api-dep.yaml
```diff
@@ -0,0 +1,126 @@
+{{- if .Values.enterprise.enabled -}}
+{{ $dict := dict "ctx" . "component" "admin-api" "memberlist" true }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  annotations:
+    {{- toYaml .Values.adminApi.annotations | nindent 4 }}
+  labels:
+    {{- include "tempo.labels" (dict "ctx" . "component" "admin-api" "memberlist" true) | nindent 4 }}
```
trevorwhitney, line 9: why not use the same `$dict` here (as you're using below)?

charts/tempo-distributed/templates/compactor/poddisruptionbudget-compactor.yaml
```diff
@@ -2,12 +2,12 @@
 apiVersion: {{ include "tempo.pdb.apiVersion" . }}
 kind: PodDisruptionBudget
 metadata:
-  name: {{ include "tempo.compactorFullname" . }}
+  name: {{ include "tempo.resourceName" (dict "ctx" . "component" "compactor") }}
```
trevorwhitney, line 5: what's the rule of thumb for when to extract a `$dict` variable vs not?

charts/tempo-distributed/templates/metamonitoring/kube-state-metrics-servmon.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if and ((.Values.metaMonitoring).grafanaAgent).enabled ((((.Values.metaMonitoring).grafanaAgent).metrics).scrapeK8s).enabled }}
```
trevorwhitney, line 1: whats the motivation for all the extra `()` here? Compared to:
```suggestion
{{- if and ((.Values.metaMonitoring).grafanaAgent).enabled .Values.metaMonitoring.grafanaAgent.metrics.scrapeK8s.enabled }}
```

charts/tempo-distributed/templates/metamonitoring/kubelet-cadvisor-servmon.yaml
```diff
@@ -0,0 +1,83 @@
+{{- if and ((.Values.metaMonitoring).grafanaAgent).enabled ((((.Values.metaMonitoring).grafanaAgent).metrics).scrapeK8s).enabled }}
```
trevorwhitney, line 1: same as above, why not:
```suggestion
{{- if and ((.Values.metaMonitoring).grafanaAgent).enabled .Values.metaMonitoring.grafanaAgent.metrics.scrapeK8s.enabled }}
```

charts/tempo-distributed/templates/metamonitoring/logs-instance-usernames-secret.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if ((.Values.metaMonitoring).grafanaAgent).enabled }}
+{{- with ((.Values.metaMonitoring).grafanaAgent).logs }}
```
trevorwhitney, line 2: another one:
```suggestion
{{- with .Values.metaMonitoring.grafanaAgent.logs }}
```

trevorwhitney (commented):
This looks great! Just a few small nits, and I'll leave the :heavy_check_mark: for @krajorama as I see he requested some changes.

charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -42,23 +49,72 @@ Create chart name and version as used by the chart label.
...
+
+{{/*
+Simple resource labels
 */}}
 {{- define "tempo.labels" -}}
-helm.sh/chart: {{ include "tempo.chart" . }}
-{{ include "tempo.selectorLabels" . }}
-{{- if or .Chart.AppVersion .Values.tempo.tag }}
-app.kubernetes.io/version: {{ .Values.tempo.tag | default .Chart.AppVersion | quote }}
+{{- if .ctx.Values.enterprise.legacyLabels }}
```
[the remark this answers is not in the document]
zalegrala: I think you're right, we probably don't need it.  I think there was another helm chart for enterprise traces, but that didn't make it back to this repo.  I'll plan to drop the legacyLabels which should clean up a bit.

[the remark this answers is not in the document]
zalegrala: `legacyLabels` has been dropped.

charts/tempo-distributed/templates/admin-api/admin-api-dep.yaml
zalegrala: Good catch.

charts/tempo-distributed/templates/compactor/poddisruptionbudget-compactor.yaml
zalegrala: Good question.  I tend to think it gets easier to read the more we do it.  I think this was brought in by a rebase and I just missed it.  I'll update to use a `$dict`.

charts/tempo-distributed/templates/configmap-tempo.yaml
krajorama: Sounds ok to me as long as it's not forgotten

charts/tempo-distributed/templates/enterprise-gateway/_helpers.tpl
```diff
@@ -0,0 +1,33 @@
+{{/*
+Return the appropriate apiVersion for ingress.
+*/}}
+{{- define "tempo.ingress.apiVersion" -}}
```
krajorama, line 4: FYI We've in Mimir recently made the breaking change to only support kube >= 1.20. Since we are not actually testing on older versions - so we could get rid of this check altogether.

charts/tempo-distributed/values.yaml
```diff
@@ -1110,43 +1275,253 @@ gateway:
...
+  tolerations: []
+  extraContainers: []
+  extraVolumes: []
+  extraVolumeMounts: []
+  env: []
+  extraEnvFrom: []
+
+# Settings for the gateway service providing authentication and authorization via the admin_api.
+# Can only be enabled if enterprise.enabled is true - requires license.
+enterpriseGateway:
```
krajorama, line 1431: Recently we got a feature request from a customer to make migrating from Mimir OSS to Enterprise Metrics easier, so now we're working on merging our "nginx" and "gateway" configuration under the "gateway" section.

So in case you want to add "nginx" for OSS users later, the "enterpriseGateway" naming will bite you. I'd consider calling it just "gateway".

See for ref: https://github.com/grafana/mimir/pull/3181

krajorama (approved):
Looks good in general. I'd consider the enterpriseGateway naming . And for sure test test test installs and migrations - including naming the helm release "tempo" and something random as well.

charts/tempo-distributed/templates/enterprise-gateway/_helpers.tpl
zalegrala: This sounds good to me.  Perhaps a follow up breaking change to upgrade the required version for the chart as well as drop the api version comparisons.

cajual:
https://github.com/grafana/helm-charts/commit/35342179f59f6aa3d9462172c4bb5283fe001286#r88097126

joshuasimon-taulia:
should minio really be enabled by default?

zalegrala:
Fair question.  After spending a bit more time with it, I think no.  Would you like to put up a PR for it @joshuasimon-taulia ?

PR #1761, opened 2022-09-01
[loki] namespaceOverride

yellowmegaman:
Adds namespaceOverride for Loki helm chart

### DIFF
--- a/charts/loki/templates/_helpers.tpl
+++ b/charts/loki/templates/_helpers.tpl
@@ -32,6 +32,17 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Allow the release namespace to be overridden for multi-namespace deployments in combined charts
+*/}}
+{{- define "loki.namespace" -}}
+  {{- if .Values.namespaceOverride -}}
+    {{- .Values.namespaceOverride -}}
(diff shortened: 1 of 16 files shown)


charts/loki/templates/_helpers.tpl
```diff
@@ -31,6 +31,17 @@ Create chart name and version as used by the chart label.
...
+{{/*
+Allow the release namespace to be overridden for multi-namespace deployments in combined charts
+*/}}
+{{- define "loki.namespace" -}}
+  {{- if .Values.namespaceOverride -}}
+    {{- .Values.namespaceOverride -}}
+  {{- else -}}
+    {{- .Release.Namespace -}}
+  {{- end -}}
+{{- end -}}
```
zanhsieh, line 43: Why don't we simplify with this?

```suggestion
{{- define "loki.namespace" -}}
{{- default .Release.Namespace .Values.namespaceOverride | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```

zanhsieh (changes requested):

zanhsieh:
@yellowmegaman 
Close due to loki deprecation https://github.com/grafana/helm-charts/pull/1772

PR #1764, opened 2022-09-01
[tempo-distributed] Add annotations to ingester's pvc

automatedops:
Allow annotations in ingestor's PVC. So that project like [pvc-autoresizer](https://github.com/topolvm/pvc-autoresizer) can be used to auto resize ingestor's PVC size.

For testing, enable persistence and update annotations
```
   persistence:
     # -- Enable creating PVCs which is required when using boltdb-shipper
-    enabled: false
+    enabled: true
     # -- use emptyDir with ramdisk instead of PVC. **Please note that all data in ingester will be lost on pod restart**
     inMemory: false
     # -- Size of persistent or memory disk
@@ -148,7 +148,8 @@ ingester:
     # set, choosing the default provisioner (gp2 on AWS, standard on GKE, AWS, and OpenStack).
     storageClass: null
     # -- Annotations for ingester's persist volume claim
-    annotations: {}
+    annotations:
+      test.com/size: 10Gi
```

Run helm template command helm template test . -f values.yaml to render templates.

Inspect ingester StatefulSet to ensure annotations are added.
```
  volumeClaimTemplates:
    - metadata:
        annotations:
          test.com/size: 10Gi
        name: data
      spec:
        accessModes:
          - ReadWriteOnce
        resources:
          requests:
            storage: "10Gi"
```

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.26.0
+version: 0.26.1
 appVersion: 1.5.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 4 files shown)


zanhsieh:
@automatedops 
Could you bump Chart.yaml version please?

zanhsieh (approved):

PR #1772, opened 2022-09-02
[loki] Deprecate old charts

trevorwhitney:
We have merged `grafana/loki` and `grafana/loki-simple-scalable` into a single chart, published as `grafana/loki`, that can be run in either configuration. We have moved this new chart into the [Loki repository](https://github.com/grafana/loki/tree/main/production/helm/loki). We believe this will allow the Loki team to have better visibility in helm issues and better support the helm community.

As a result of the move, we are deprecating the charts in this repository and pointing people towards the replacement.


zanhsieh (approved):

flokli:
I got struck by this a bit, and documented my strugglings in https://github.com/grafana/loki/issues/9427. Would be nice if the upgrade experience could be improved.

PR #1778, opened 2022-09-06
[grafana] PET-1760 - Synchronise grafana sidecar envvars, add IGNORE_ALREADY_PROCESSED

### LINKED ISSUE #1777: [grafana] sidecar envvars inconsistent between dashboards, datasource and plugins
The grafana helm chart `_pod.tpl` template contains different sidecar environment variables for each sidecar e.g., the dashboard sidecar has a `LOG_LEVEL` variable, datasources and plugins do not.
Even worse none of them support the IGNORE_ALREADY_PROCESSED parameter, though the dashboard sidecar at least allows specifying it as `.Values.sidecar.dashboards.env`. This is not supported by the other two sidecars and their initContainer pendants do not support this either.

This prevents configuring proper logging for datasource and plugin sidecars, for their initContainer pendants. In addition it prevents accomodating further envvars for the sidecar image without a subsequent helm chart code change.

---

I guess at least consistently providing configuration for those envvars that can be used for each of the sidecars would help. Explicitly adding `IGNORE_ALREADY_PROCESSED` would be a bonus.

git commit ID:  1256a0193f8da3234334212982c11618 (current main branch)
Latest grafana chart release: 6.36.2

Arabus:
Fixes #1777

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.37.3
+version: 6.38.0
 appVersion: 9.1.4
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -93,6 +93,10 @@ initContainers:
     {{- end }}
     imagePullPolicy: {{ .Values.sidecar.imagePullPolicy }}
     env:
+      {{- range $key, $value := .Values.sidecar.datasources.env }}
+      - name: "{{ $key }}"
+        value: "{{ $value }}"
+      {{- end }}
       - name: METHOD
         value: "LIST"
       - name: LABEL
@@ -142,6 +146,10 @@ initContainers:
     {{- end }}
     imagePullPolicy: {{ .Values.sidecar.imagePullPolicy }}
     env:
+      {{- range $key, $value := .Values.sidecar.notifiers.env }}
+      - name: "{{ $key }}"
+        value: "{{ $value }}"
+      {{- end }}
       - name: METHOD
         value: LIST
       - name: LABEL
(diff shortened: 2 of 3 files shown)


zanhsieh (changes requested):
@Arabus 
Could you restore non-related changes please? It's hard to keep tracking which field already pre-existed.

Arabus:
err sorry got a bit carried away there and started sorting vars too. I'll reduce the MR to a minimum.

Arabus:
Sorry about that - somehow part of my changes got lost during rebasing. The change request now contains only the addition of missing envvars to initcontainers and sidecars. If you like I can squash them before merging.

I also noticed that some values e.g., `.Values.sidecars.loglevel` can only be configured globally for all sidecars. If you like I can add an override on the individual sidecar level.
Furthermore there seems to be no safeguard against setting the WATCH_TIMEOUT values when using LIST or SLEEP methods. While not detrimental to the sidecar (they are just not used i those cases) it might help to throw a warning instead of just silently setting them. If you like I can add that too.

zanhsieh:
> Sorry about that - somehow part of my changes got lost during rebasing. The change request now contains only the addition of missing envvars to initcontainers and sidecars. If you like I can squash them before merging.
> 
> I also noticed that some values e.g., `.Values.sidecars.loglevel` can only be configured globally for all sidecars. If you like I can add an override on the individual sidecar level. Furthermore there seems to be no safeguard against setting the WATCH_TIMEOUT values when using LIST or SLEEP methods. While not detrimental to the sidecar (they are just not used i those cases) it might help to throw a warning instead of just silently setting them. If you like I can add that too.

Sure. Please do.

zanhsieh (approved):

zanhsieh:
Close and reopen to make CI run again.

zanhsieh (changes requested):
@Arabus 
Could you rebase and resolve the conflict please?

Arabus:
of course - this should do it. Do you want me to squash the commits ?

zanhsieh (approved):

Issue #1782, opened 2022-09-07
[tempo-distributed] PodDisruptionBudget API Version should be `policy/v1` for Kubernetes version 1.21 and above


andy-mcgrath:
Updating the tempo-distributed helm chart on Kubernetes 1.25 and deployment is failing due to API version `policy/v1beta1 PodDisruptionBudget` being deprecated in v1.21+ and unavailable in v1.25+

### chart tempo-distributed suggested improvements
- [x] Add helper `tempo.pdb.apiVersion`
- [x] Update all `PodDisruptionBudget` resources `apiVersion` to use `tempo.pdb.apiVersion`

PR #1783, opened 2022-09-07
[tempo-distributed] Add helper for PodDisruptionBudget API Version

andy-mcgrath:
- Add helper tempo.pdb.apiVersion
- Update all PodDisruptionBudgetresourcesapiVersionto usetempo.pdb.apiVersion`

Fixes issue #1782
Signed-off-by: Andy McGrath <upon-pave0d@icloud.com>

### DIFF
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -120,3 +120,14 @@ Return if ingress supports pathType.
 {{- define "tempo.ingress.supportsPathType" -}}
   {{- or (eq (include "tempo.ingress.isStable" .) "true") (and (eq (include "tempo.ingress.apiVersion" .) "networking.k8s.io/v1beta1") (semverCompare ">= 1.18-0" .Capabilities.KubeVersion.Version)) -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for PodDisruptionBudget.
+*/}}
+{{- define "tempo.pdb.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version) -}}
+    {{- print "policy/v1" -}}
+  {{- else -}}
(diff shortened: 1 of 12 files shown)


BitProcessor:
Hi @andy-mcgrath - thanks for your contribution. Seems like the docs need some fixing. 
See: https://github.com/grafana/helm-charts/runs/8227053520?check_suite_focus=true

charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -120,3 +120,14 @@ Return if ingress supports pathType.
 {{- define "tempo.ingress.supportsPathType" -}}
   {{- or (eq (include "tempo.ingress.isStable" .) "true") (and (eq (include "tempo.ingress.apiVersion" .) "networking.k8s.io/v1beta1") (semverCompare ">= 1.18-0" .Capabilities.KubeVersion.Version)) -}}
 {{- end -}}
+
+{{/*
+Return the appropriate apiVersion for PodDisruptionBudget.
+*/}}
+{{- define "tempo.pdb.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version) -}}
```
zanhsieh, line 128: Actually there are two common fixes for this:

1. Stick on what Kubernetes version, which is how loki / loki-simple-distributed stick on.
```suggestion
  {{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version -}}
```

2. Stick on which API available, which is how grafana and prometheus related charts stick on.
```suggestion
  {{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" -}}
```

zanhsieh (commented):
See my comment.

charts/tempo-distributed/templates/_helpers.tpl
andy-mcgrath: What's the preference? Personally I would have just check the API version but happy to update either way.

zanhsieh (approved):
See my comment.

PR #1786, opened 2022-09-07
Fix update-helm-repo.yaml

krajorama:
Better separation of package name from git tag name. I think this was the intention of PR 1779

I don't know if we want to keep the git tag in the source repository and the helm-charts repo in sync or not. This PR does not keep them in sync.

### DIFF
--- a/.github/workflows/update-helm-repo.yaml
+++ b/.github/workflows/update-helm-repo.yaml
@@ -157,7 +157,12 @@ jobs:
           version=$(yq ".version" < ${changed}/Chart.yaml)
           echo "::set-output name=chartpath::${changed}"
           echo "::set-output name=desc::${description}"
-          echo "::set-output name=tagname::${name}-${version}"
+          if [[ -n "${HELM_TAG_PREFIX}" ]]; then
+            echo "::set-output name=tagname::${HELM_TAG_PREFIX}-${name}-${version}"
+          else
+            echo "::set-output name=tagname::${name}-${version}"
+          fi
+          echo "::set-output name=packagename::${name}-${version}"
 
       - name: Install CR tool
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-helm-repo.yaml
```diff
@@ -195,10 +196,10 @@ jobs:
 
             Tag on source: https://github.com/${{ github.repository }}/releases/tag/${{ steps.parse-chart.outputs.tagname }}
           files: |
-            ${{ env.CR_PACKAGE_PATH }}/${{ steps.parse-chart.outputs.tagname }}.tgz
-            ${{ env.CR_PACKAGE_PATH }}/${{ steps.parse-chart.outputs.tagname }}.tgz.prov
+            ${{ env.CR_PACKAGE_PATH }}/${{ steps.parse-chart.outputs.packagename }}.tgz
+            ${{ env.CR_PACKAGE_PATH }}/${{ steps.parse-chart.outputs.packagename }}.tgz.prov
           repository: grafana/helm-charts
-          tag_name: ${{ steps.parse-chart.outputs.tagname }}
+          tag_name: ${{ steps.parse-chart.outputs.packagename }}
```
dimitarvdimitrov, line 202: shouldn't that remain as `tagname`?

krajorama: that's exactly my question to @trevorwhitney  :) I don't know what the intention was (for mimir the two are the same)

dimitarvdimitrov: i asked because the git tag uses tagname
```
git tag "${{ steps.parse-chart.outputs.tagname }}"
```
if we use a different value on the release than the tag name, will that work?

trevorwhitney: so the intention is to not create a tag `loki-v3.0.0`, as it confused people, making them think that was a new version of `loki` being released. We want the tag to be prefixed with `helm` to not confuse people. Mimir doesn't have this problem because your chart is called `mimir-distributed`.

I think we need to keep this as `tagname` as @dimitarvdimitrov pointed out.

trevorwhitney (approved):
I think we need to revert the `tagname` -> `packagename` for the `tag_name` part of the "Make github release" step. However, changing to `packagename` for the `.tgz` file is probably correct. Other than that LGTM

trevorwhitney (approved):
LGTM!

PR #1788, opened 2022-09-07
Fix tagname at original set-output

trevorwhitney:
The addition of the `$HELM_TAG_PREFIX` was not being picked up everywhere, so this PR moves that logic to where the output is first set.


jeschkies (approved):

trevorwhitney:
looks like this is a duplicate of https://github.com/grafana/helm-charts/pull/1786, closing

PR #1790, opened 2022-09-07
[tempo-distributed] switch poddisruptionbudget to v1 policy

rlex:
pdb v1beta1 has been deprecated since k8s 1.21, which is EOL already. 
Migrate it to v1, making it compatible with k8s 1.25, but breaking compatibility with k8s <1.21


BitProcessor:
Thanks for your contribution, this seems however double with: https://github.com/grafana/helm-charts/pull/1783

PR #1793, opened 2022-09-08
[loki-distributed] fix: pdbs should take the right value

aslafy-z:
Almost all Loki distributed PDBs takes them values from the `distributor` configuration. 
Also, PDBs are deployed for services that are not enabled.

This PR fixes these two issues.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.6.1
-version: 0.56.6
+version: 0.56.7
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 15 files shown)


zanhsieh (approved):

JeffreyVdb:
@aslafy-z This fix only works when you specify the `enabled` key. Some components like querier do not use the `enabled` key and are always enabled. The code does rely on the enabled keyword being true:

```
{{- if and .Values.querier.enabled (gt (int .Values.querier.replicas) 1) }}
```

It should actually simply be this:

```
{{- if gt (int .Values.querier.replicas) 1 }}
```

Completely unrelated to the scope of your pull request: this won't work either when autoscaling is used. I noticed that the nginx-ingress chart also takes that into account: https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/templates/controller-poddisruptionbudget.yaml#L1

aslafy-z:
@JeffreyVdb Good catch, thank you. Here's the fix: https://github.com/grafana/helm-charts/pull/1801

PR #1798, opened 2022-09-09
[tempo-distributed] fix: remove extraneous ports from query-frontend-discovery service

oliverfrye:
This PR ensures service ports related to the optional `tempo-query` component are only configured when `tempo-query` is enabled. This is consistent with the existing `query-frontend` service configuration.

Additionally, the `grpclb` port has been moved immediately below `grpc`, so that all the ports are listed in order.


BitProcessor (approved):

BitProcessor:
Thanks for your contribution 🚀

PR #1800, opened 2022-09-09
[tempo] make tempo automountServiceAccountToken configurable

gerhard-strauss-jambit:
just a simple change to make tempo automountServiceAccountToken configurable

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.16.1
+version: 0.16.2
 appVersion: 1.5.0
 engine: gotpl
 home: https://grafana.net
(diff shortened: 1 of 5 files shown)


BitProcessor:
Thanks for your contribution, a few questions/remarks:

* What would be the use case? 
* Why are you setting it both on serviceaccount and pod level? See: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server
* please change the title of the PR and add the correct prefix

gerhard-strauss-jambit:
> * What would be the use case?
> * Why are you setting it both on serviceaccount and pod level? See: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server
I know that its kinda redundant, but it also doesnt hurt. We have a Policy in our Kubernetes Clusters active, that requires the automountServiceAccountToken on Pod Level and we cant influence this. To Deploy this Chart we need this Setting.

> * please change the title of the PR and add the correct prefix
Done

BitProcessor:
Thanks for your feedback. Please also update the helm-docs.

gerhard-strauss-jambit:
> Thanks for your feedback. Please also update the helm-docs.

done

BitProcessor:
It seems your helm-docs ran all over the place ;) 
Can you remove the changes in this file?

Screenshot 2022-09-09 at 13 22 22

gerhard-strauss-jambit:
> It seems your helm-docs ran all over the place ;) Can you remove the changes in this file?
> 
> Screenshot 2022-09-09 at 13 22 22

yes, sorry for that.
done

zanhsieh (approved):

gerhard-strauss-jambit:
@zanhsieh can you merge?

PR #1801, opened 2022-09-09
[loki-distributed] fix: pdb conditional deployment 

aslafy-z:
Addressing https://github.com/grafana/helm-charts/pull/1793#issuecomment-1241866919


JeffreyVdb:
You should do the same for:

- distributor
- ingester
- queryFrontend

These are always enabled and do not rely on a flag. Doubled checked with my configuration and I have them all deployed without having `enabled: true` in my `values.yaml` file.

aslafy-z:
@JeffreyVdb please have another look. Thank you

JeffreyVdb:
@aslafy-z Looks good to me!

zanhsieh (approved):

PR #1802, opened 2022-09-09
[tempo-distributed] Add value for changing metrics-generator wal volume to be able to mount it in memory

uberspot:
The default of using local disk for scratch space is still kept in values so this should be a no-op change for people upgrading.

### DIFF
--- a/charts/tempo-distributed/templates/metrics-generator/deployment-metrics-generator.yaml
+++ b/charts/tempo-distributed/templates/metrics-generator/deployment-metrics-generator.yaml
@@ -101,7 +101,7 @@ spec:
             name: {{ template "tempo.fullname" . }}
           name: tempo-conf
         - name: wal
-          emptyDir: {}
+          emptyDir: {{- toYaml .Values.metricsGenerator.walEmptyDir | nindent 12 }}
         {{- with .Values.metricsGenerator.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -220,7 +220,11 @@ metricsGenerator:
   # -- Tolerations for metrics-generator pods
   tolerations: []
   # -- The EmptyDir location where the /var/tempo will be mounted on. Defaults to local disk, can be set to memory.
-  walEmptyDir:  {}
+  walEmptyDir: {}
+    ## Here show how to configure 1Gi memory as emptyDir.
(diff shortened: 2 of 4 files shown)


charts/tempo-distributed/values.yaml
```diff
@@ -219,6 +219,8 @@ metricsGenerator:
   nodeSelector: {}
   # -- Tolerations for metrics-generator pods
   tolerations: []
+  # -- The EmptyDir location where the /var/tempo will be mounted on. Defaults to local disk, can be set to memory.
+  walEmptyDir:  {}
```
zanhsieh, line 223: Please provide some examples under comment, e.g.
```suggestion
  walEmptyDir: {}
    ## Here show how to configure 1Gi memory as emptyDir.
    ## Ref: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#emptydirvolumesource-v1-core
    # medium: "Memory"
    # sizeLimit: 1Gi
```

zanhsieh (changes requested):

zanhsieh (approved):

zanhsieh (approved):

charts/tempo-distributed/templates/metrics-generator/deployment-metrics-generator.yaml
```diff
@@ -101,7 +101,7 @@ spec:
             name: {{ template "tempo.fullname" . }}
           name: tempo-conf
         - name: wal
-          emptyDir: {}
+          emptyDir: {{- toYaml .Values.metricsGenerator.walEmptyDir | nindent 12 }}
```
BitProcessor, line 104: Should this be an emptyDir to begin with? This is the place where the WAL is stored. 🤔 
@annanay25

uberspot: Does this need persisting? Or does it just regenerate it on startup?

BitProcessor: @mdisibio Martin, can you shine a light on this?

kvrhdn: Hi, sorry for the delayed response here.

The metrics-generator can generate the WAL directory when it starts up, so if it doesn't exist it will be created anyway, if it already exists it just checks if it's writeable.
So either is fine for this PR 🙂

BitProcessor: @kvrhdn Does that mean that persistence is not needed? And if so, should this be called a WAL ?

Issue #1727, opened 2022-08-25
add extra containers in charts/promtail/templates/_pod.tpl


xiaobin-qmcx:
I  deployed the promtail helm chart to upload the logs to loki. However, I can't find any config in charts/promtail/values.yaml which can deploy additional containers. For example, I'd deploy another container which can clean the old logs files.  Can you write some code to support this function in charts/promtail/templates/_pod.tpl  ? Thanks.
Like the code in charts/promtail/templates/_pod.tpl   :
`{{- with .Values.extrContainer }}
   {{- toYaml . | nindent 6 }}
{{- end }}`

conorevans:
This can be closed @xiaobin-qmcx (I forgot to reference properly in #1807)

PR #1806, opened 2022-09-10
[promtail] feat: support extraContainers for Promtail Deployments

### LINKED ISSUE #1805: [promtail] Support extra containers for Deployment
**Why?**

I need to expose Promtail Push HTTP API to get logs into Loki from Vercel (see https://gihub.com/grafana/loki/issues/4316). I want to put auth in front of it, but unfortunately Vercel only supports a SHA1 identification system (see [docs](https://vercel.com/docs/log-drains#secure-log-drains).

This isn't able to be handled by my standard ingress auth and I don't want to have to implement massive changes to support multiple auth mechanisms. Instead I have a simple SingleHostReverseProxy that will be routed to by my ingress, authenticate, then forward the request to Promtail HTTP Push API.

I want to ensure the proxy is deployed alongside Promtail as part of the same Deployment, but that's not currently possible.

It's common enough to have sidecar containers in these charts, e.g. https://airflow.apache.org/docs/helm-chart/stable/using-additional-containers.html

conorevans:
Closes #1805 (see there for motivation) #1727 (didn't realise this existed before)

I am not a Helm templating expert. Typically I see this sort of `extra` being done as a List, but I couldn't get it to render with the necessary `-` at the beginning of the extra containers in the overall list, so I went with the `name: values` hash approach. Feel free to improve that.

Signed-off-by: Conor Evans <coevans@tcd.ie>

### DIFF
--- a/charts/promtail/README.md
+++ b/charts/promtail/README.md
@@ -95,7 +95,7 @@ The new release which will pick up again from the existing `positions.yaml`.
 | deployment.enabled | bool | `false` | Deploys Promtail as a Deployment |
 | deployment.replicaCount | int | `1` |  |
 | extraArgs | list | `[]` |  |
-| extraContainers | object | `{}` |  |
+| extraContainers | object | `{}` | Extra containers to deploy alongside a Promtail Deployment (only), e.g. an auth proxy |
 | extraEnv | list | `[]` | Extra environment variables |
 | extraEnvFrom | list | `[]` | Extra environment variables from secrets or configmaps |
 | extraObjects | list | `[]` | Extra K8s manifests to deploy |
--- a/charts/promtail/README.md
+++ b/charts/promtail/README.md
@@ -270,8 +270,8 @@ config:
       log_level: {{ .Values.config.logLevel }}
       http_listen_port: {{ .Values.config.serverPort }}
 
-    client:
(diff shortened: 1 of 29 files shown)


charts/promtail/templates/_pod.tpl
```diff
@@ -81,6 +81,12 @@ Pod template used in Daemonset and Deployment
           resources:
             {{- toYaml . | nindent 12 }}
           {{- end }}
+      {{- if .Values.deployment.enabled }}
```
conorevans, line 84: I decided to add this only for deployments as running a sidecar for a DaemonSet seems very wrong 🙃

charts/promtail/README.md
```diff
@@ -91,6 +91,7 @@ The new release which will pick up again from the existing `positions.yaml`.
 | deployment.enabled | bool | `false` | Deploys Promtail as a Deployment |
 | deployment.replicaCount | int | `1` |  |
 | extraArgs | list | `[]` |  |
+| extraContainers | object | `{}` |  |
```
conorevans, line 94:
```suggestion
| extraContainers | object | `{}` | Extra containers to deploy alongside a Promtail Deployment (only), e.g. an auth proxy |
```

conorevans:
git merge was _not_ my friend on this occasion... working to fix

PR #1823, opened 2022-09-18
[grafana] Fix whitespace issue due to empty value

glindstedt:
If not using the ingress feature the domain value would be empty. This resulted in the template producing an ini file with a line like this:

`domain = \n`

This would cause issues when passing the resource through a tool like kustomize, which would turn the multiline block into an encoded single line string in order to preserve trailing whitespace, resulting in kubectl diffs that are very hard to review.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.38.4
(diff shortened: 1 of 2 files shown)


glindstedt:
@zanhsieh done, rebased and bumped the version

zanhsieh (approved):

PR #1828, opened 2022-09-21
[loki-distributed] Change component name to actual.

PRIHLOP:


zanhsieh (changes requested):
@PRIHLOP 
Thank you for contribution. Could you bump the version in Chart.yaml please?

patsevanton:
@zanhsieh Please close PR.
values.yaml has updated code
https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/values.yaml#L499

PR #1830, opened 2022-09-21
[agent-operator] Allow using PodSecurityPolicy

dimitarvdimitrov:
This is useful when the agent-operator is being used as a subchart and
the parent chart is using a PodSecurityPolicy. This new value will allow
the user of the parent chart to configure the agent to use the same
PSP as the parent chart.

Signed-off-by: Dimitar Dimitrov <dimitar.dimitrov@grafana.com>

### DIFF
--- a/charts/agent-operator/templates/operator-clusterrole.yaml
+++ b/charts/agent-operator/templates/operator-clusterrole.yaml
@@ -52,4 +52,11 @@ rules:
   - daemonsets
   - deployments
   verbs: [get, list, watch, create, update, patch, delete]
+{{- with .Values.rbac.podSecurityPolicyName }}
+- apiGroups: [policy]
+  resources:
+  - podsecuritypolicies
+  verbs:         [use]
+  resourceNames: [ {{ . }} ]
+{{- end -}}
 {{- end -}}
--- a/charts/agent-operator/Chart.yaml
+++ b/charts/agent-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: grafana-agent-operator
 description: A Helm chart for Grafana Agent Operator
 type: application
-version: 0.2.5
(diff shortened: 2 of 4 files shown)


charts/agent-operator/templates/operator-clusterrole.yaml
```diff
@@ -52,4 +52,11 @@ rules:
   - daemonsets
   - deployments
   verbs: [get, list, watch, create, update, patch, delete]
+{{- with .Values.rbac.podSecurityPolicyName }}
+- apiGroups: [policy]
+  resources:
+  - podsecuritypolicies
+  verbs:         [use]
+  resourceNames: [ {{ . }} ]
+{{- end -}}
```
zanhsieh, line 61: Do we need to take care k8s 1.25 psp deprecation?

```suggestion
{{- with .Values.rbac.podSecurityPolicyName }}
{{- if .Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy" }}
- apiGroups: [policy]
  resources:
  - podsecuritypolicies
  verbs:         [use]
  resourceNames: [ {{ . }} ]
{{- end }}
{{- end -}}
```

zanhsieh (changes requested):

charts/agent-operator/templates/operator-clusterrole.yaml
dimitarvdimitrov: My opinion so far has been the I would leave that to the end user. If they are still configuring a PSP maybe they rely on it; then the chart will fail to install. They can be explicit to not use a PSP and the chart will install ok.

dimitarvdimitrov: I'm not feeling super strongly about this though

zanhsieh (approved):

PR #1838, opened 2022-09-23
[grafana] Fix notifier sidecar only running as initContainer

### LINKED ISSUE #327: Notifiers sidecar doesn't run continuously
In contrast to the dashboard sidecar, the sidecar for picking up notifier definitions doesn't run throughout the lifetime of the pod. This means that a manual restart of Grafana is required once notifier changes are deployed to the cluster. An automatic mechanism would be what I would have expected after the experience with the dashboard sidecar.

MiggL:
Currently, the notifier "sidecar" only runs as an init container. Previously reported in issue #327.
This breaks the sidecar behaviour since a restart of the pod is necessary for notifier configmap updates to be loaded by Grafana.

This pull request fixes #327.

I've also added more options to configure the notifier sidecar, which makes its configuration more similar to other sidecars, such as the datasource sidecar. Therefore, I've bumped the minor version of the chart.


MiggL:
Right, documentation... I can update the README next week.

zanhsieh (approved):

PR #1839, opened 2022-09-23
[grafana] fix image renderer network policy definition

romcheck:
looks like a typo. see https://kubernetes.io/docs/concepts/services-networking/network-policies/#networkpolicy-resource


zanhsieh:
@romcheck 
Could you sign DCO and bump the version in Chart.yaml? Thank you.

zanhsieh (approved):

PR #1844, opened 2022-09-26
[grafana] Add Red Hat OpenShift support, change image to grafana-enterprise, allow…

patrickabrennan:
fields for the Grafana Enterprise license, turn PVC on so if the Grafana Pod restarts data will not be lost. Also by using the grafana-enterprise image one has all the features of community Grafana image, but easily all customers to enable the enterprise version.


charts/grafana/values.yaml
```diff
@@ -435,11 +436,11 @@ enableServiceLinks: true
 
 ## Additional grafana server secret mounts
 # Defines additional mounts with secrets. Secrets must be manually created in the namespace.
-extraSecretMounts: []
-  # - name: secret-files
-  #   mountPath: /etc/secrets
-  #   secretName: grafana-secret-files
-  #   readOnly: true
+extraSecretMounts: 
+    - name: ge-license
```
vlad-diachenko, line 440: this secret must be manually created outside the helm chart. according to this:
```
## Additional grafana server secret mounts
# Defines additional mounts with secrets. Secrets must be manually created in the namespace.
```
So, not enterprise users won't have it and it will prevent the pod from starting up.  As long as I remember.

charts/grafana/values.yaml
```diff
@@ -73,7 +74,7 @@ livenessProbe:
 # schedulerName: "default-scheduler"
 
 image:
-  repository: grafana/grafana
+  repository: grafana/grafana-enterprise
```
vlad-diachenko, line 77: it might be breaking change for somebody. I know that some companies do not use docker hub directly, instead of this they use company's docker repository proxy and they use only some list of allowed docker images. 
so, if they do not have this image in the white list, their k8s cluster might be not able to download the image. 
I think we need to mention it in upgrade guide: https://github.com/grafana/helm-charts/blob/main/charts/grafana/README.md#upgrading-an-existing-release-to-a-new-major-version

charts/grafana/values.yaml
```diff
@@ -294,7 +295,7 @@ extraContainerVolumes: []
 ##
 persistence:
   type: pvc
-  enabled: false
+  enabled: enabled
```
vlad-diachenko, line 298: it's better to mention it in upgrade guide

charts/grafana/templates/_helpers.tpl
```diff
@@ -172,3 +172,24 @@ Return if ingress supports pathType.
...
+{{/* Allow KubeVersion to be overridden. */}}
+{{- define "grafana.kubeVersion" -}}
+  {{- default .Capabilities.KubeVersion.Version .Values.kubeVersionOverride -}}
+{{- end -}}
+
+{{/*
+Return if we should create a PodSecurityPoliPodSecurityPolicycy. Takes into account user values and supported kubernetes versions.
+*/}}
+{{- define "grafana.rbac.usePodSecurityPolicy" -}}
+{{- and (semverCompare "< 1.25-0" (include "grafana.kubeVersion" .)) (and .Values.rbac.create (eq .Values.rbac.type "psp")) -}}
```
vlad-diachenko, line 187: for k8s version less than 1.25 we won't create pod security policy? 
user might be confused if he enabled psp in values file : `.Values.rbac.create and (eq .Values.rbac.type "psp")`
but we have not created it. 
can we add some WARNING or something like this?

charts/grafana/templates/_helpers.tpl
```diff
@@ -172,3 +172,24 @@ Return if ingress supports pathType.
...
+
+
+
+{{/* Allow KubeVersion to be overridden. */}}
+{{- define "grafana.kubeVersion" -}}
+  {{- default .Capabilities.KubeVersion.Version .Values.kubeVersionOverride -}}
+{{- end -}}
+
+{{/*
+Return if we should create a PodSecurityPoliPodSecurityPolicycy. Takes into account user values and supported kubernetes versions.
```
vlad-diachenko, line 184:
```suggestion
Return if we should create a PodSecurityPolicy. Takes into account user values and supported kubernetes versions.
```

vlad-diachenko (commented):
PSP and SCC config looks great. just a small nit. 
I have some doubts if we need to enable persistence and use enterprise image by default.
It's better to wait for Grafana team for the review as long as they are maintainers of this helm chart.
Also, we need to bump version of the chart   in https://github.com/grafana/helm-charts/blob/main/charts/grafana/Chart.yaml .

charts/grafana/values.yaml
patrickabrennan: Good point. So maybe we take that out.

charts/grafana/values.yaml
patrickabrennan: Also a very valid point.

charts/grafana/values.yaml
patrickabrennan: While one could do that I think this would avoid some people missing to turn PVC on and thus losing data when the Pod restarts.

charts/grafana/templates/_helpers.tpl
patrickabrennan: Makes total sense as psp goes away with K8s 1.25

charts/grafana/values.yaml
```diff
@@ -435,10 +436,10 @@ enableServiceLinks: true
 
 ## Additional grafana server secret mounts
 # Defines additional mounts with secrets. Secrets must be manually created in the namespace.
-extraSecretMounts: []
-  # - name: secret-files
-  #   mountPath: /etc/secrets
-  #   secretName: grafana-secret-files
+extraSecretMounts:
```
vlad-diachenko, line 439: I think it's better to set empty array by default to not bring possible issues 
```suggestion
extraSecretMounts: []
```

charts/grafana/values.yaml
```diff
@@ -679,6 +680,9 @@ grafana.ini:
     url: https://grafana.net
   server:
     domain: "{{ if (and .Values.ingress.enabled .Values.ingress.hosts) }}{{ .Values.ingress.hosts | first }}{{ else }}''{{ end }}"
+    root_url:
```
vlad-diachenko, line 683: just a question: do we need to set `null` to root_url?

vlad-diachenko (commented):
do we need to commit these binaries? 
* charts/grafana.original.tar
* charts/grafana.working.09232022.tar

Supra89kren (approved):
looks great. just bump a version of the chart and an entry to `CHANGELOG.md`

zanhsieh:
@vlad-diachenko @patrickabrennan 
Actually I favour the approach of loki / tempo charts for documentation: use helm-doc, but it doesn't seem widely accepting by contributors. We could help the user to resolve ambiguous type of variables / stanzas with helm-docs.

PR #1756, opened 2022-08-31
[grafana] Load unified alerts through sidecar

jleloup:
This PR adds the capacity to use Grafana 9.1 file provisioning for unified alerting through the same sidecar mechanism already in place for dashboards, datasources or notifiers.

This should help for [#1710](https://github.com/grafana/helm-charts/issues/1710)

PR #1847, opened 2022-09-28
[grafana] load alerts through sidecar

jleloup:
This is a continuation of [PR 1756](https://github.com/grafana/helm-charts/pull/1756) after a wrong manipulation there.

Previous description:

This PR adds the capacity to use Grafana 9.1 file provisioning for unified alerting through the same sidecar mechanism already in place for dashboards, datasources or notifiers.

This should help for https://github.com/grafana/helm-charts/issues/1710

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.39.0
+version: 6.40.0
 appVersion: 9.1.6
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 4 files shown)


zanhsieh (approved):

Mattie112:
We are (still) not sure on how to use this. I posted the question at https://github.com/grafana/helm-charts/issues/1992 but perhaps here at the MR would have been a better place? If there are any tips on how to use this it would be greatly appreciated!

Mattie112:
This is still an issue for us. Is anyone actually using this that can explain how it works?

Issue #1683, opened 2022-08-12
Loki distributed: Option to configure ingester pvc annotations


sbr82:
Currently there is no option to configure annotations for the ingester pvc, only storageClass and size:

ingester.persistence.storageClass
ingester.persistence.size

This is a blocker for my project as we are running Loki on multiple clusters with shared nfs server and we rely on annotations to create unique names for the volume paths.

PR #1850, opened 2022-09-28
[loki-distributed] Introduce appProtocol field for gateway service

verejoel:
This fixes issue #1849

PR #1864, opened 2022-10-01
[loki-distributed] Add annotations for Persistent Volume Claims

verejoel:
This solves #1683, and extends the options to include annotations on all Persistent Volume Claims (not just the ingesters).


charts/loki-distributed/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.6.1
-version: 0.58.0
```
verejoel, line 6: In anticipation of #1850 being merged first

zanhsieh (approved):

anthonyalayo:
@verejoel @zanhsieh I encountered the same issue, and I noticed that the annotations are only available on loki-distributed. Is it possible to port these to loki as well? There is currently no way in the loki chart to set annotations on the PVC for claim `monitoring/storage-loki-0`.

anthonyalayo:
Friendly bump @verejoel @zanhsieh

PR #1884, opened 2022-10-12
[tempo-distributed] add appProtocol field to the memberlist, distributor, ingester, metricsGenerator, querier and the queryFrontend

AlphaStyle:
When Istio is configured with `PeerAuthentication`-[policy](https://istio.io/latest/docs/concepts/security/#authentication-policies) mode set to STRICT, mutual TLS is enforced by default. Istio is relying on [protocol selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) to proxy traffic. This pull request will introduce an optional `appProtocol`- and `appProtocol.grpc`-option for the memberlist-, ingester-, distributor-, querier-, metrics-generator- and query-frontend-service.

Feedback on the field names, `tempo.memberlist.appProtocol`, `<service>.appProtocol.grpc`, and documentation are welcome.

Istio doesn't work well when the gossip-ring / memberlist is of type `http`, I have to use `tcp`. That is why the optional `tempo.memberlist.appProtocol`-option is also part of this pull request.

This pull request is similar to the [loki-distributed 1697 PR](https://github.com/grafana/helm-charts/pull/1697).

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.27.3
+version: 0.27.4
 appVersion: 1.5.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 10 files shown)


AlphaStyle:
Did a version bump and README.md update.

Also did a rebase with upstream where the `service-gossip-ring.yaml`-template got moved into it's own folder.

zalegrala (approved):
Nice, this looks good to me.  Thanks for the contribution.

PR #1053, opened 2022-02-17
[tempo] Deploy tempo ServiceAccount in release namespace

sbueringer:
The tempo Helm Chart can currently not be deployed in a namespace other than default, as the ServiceAccount is not deployed to the release namespace.

PR #1885, opened 2022-10-12
[tempo-distributed] Deploy tempo ServiceAccount in release namespace

paolopiaggio:
similar fix as #1053 by @sbueringer  but for the  tempo-distributed chart:

The tempo Helm Chart can currently not be deployed in a namespace other than default, as the ServiceAccount is not deployed to the release namespace.


paolopiaggio:
sure thing, just pushed... thanks @zanhsieh !

zanhsieh (approved):

Issue #157, opened 2021-01-04
Loki-distributed doesn't work out of the box


tim-sendible:
I'm running on EKS 1.18

If I take the loki-distributed helm chart and apply it with the values.yml as it is written, I end up with the distributor, ingester and querier in a crashloopbackoff state complaining `failed to create memberlist: Failed to get final advertise address: no private IP address found, and explicit IP not provided`.

This seems to be a reasonably common error related to the memberlist. I understand that I should provide a private IP address. However it's unclear what address I should be adding.

If I add:
```
    bind_addr:
        - 127.0.0.1
```
to the memberlist config, things seem to get a little further. All the containers at least go ready, but they eventually fail, and the ring fails to get any members added to it (shown by navigating to the /ring url of the distributor service)

127.0.0.1 is a complete guess based on trial and error as I can't find any documentation explaining what IP address I should be applying here.

I have also tried 172.120.0.0/16 which is the cidr range of IP addresses available to my pods.  This time, I see see ingester being added to the ring. It is even temporarily healthy, before the state goes to 'unhealthy' and everything grinds to a halt again.

Here are some logs from the ingester that may or may not be useful? By this point, the state of the instance in the ring is 'unhealthy', even though it seems to be uploading the tables somewhere?

```
│ level=info ts=2021-01-04T15:35:16.752555087Z caller=lifecycler.go:547 msg="instance not found in ring, adding with no tokens" ring=ingester                                                                                  │
│ level=info ts=2021-01-04T15:35:16.752736192Z caller=lifecycler.go:394 msg="auto-joining cluster after timeout" ring=ingester                                                                                                 │
│ level=info ts=2021-01-04T15:35:16.756522831Z caller=memberlist_client.go:461 msg="joined memberlist cluster" reached_nodes=2                                                                                                 │
│ ts=2021-01-04T15:35:17.753163266Z caller=memberlist_logger.go:74 level=warn msg="Got ping for unexpected node 'loki-test-loki-distributed-querier-0-104bc685' from=[::]:7946"                                                │
│ ts=2021-01-04T15:35:18.254113836Z caller=memberlist_logger.go:74 level=warn msg="Got ping for unexpected node loki-test-loki-distributed-querier-0-104bc685 from=127.0.0.1:54602"                                            │
[...]

PR #1891, opened 2022-10-14
[loki-distributed] Added support for IPv6 only clusters

stevehipwell:
This PR enables the modification of the Loki Gateway Nginx listener without needing to duplicate and maintain the whole `gateway.nginxConfig.file` value; this functionality is required to use Loki in an EKS IPv6 cluster. Unless you change `gateway.nginxConfig.serverListens` this change should be a no-op.

### DIFF
--- a/charts/loki-distributed/README.md
+++ b/charts/loki-distributed/README.md
@@ -1,6 +1,6 @@
 # loki-distributed
 
-![Version: 0.69.16](https://img.shields.io/badge/Version-0.69.16-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.8.2](https://img.shields.io/badge/AppVersion-2.8.2-informational?style=flat-square)
+![Version: 0.69.17](https://img.shields.io/badge/Version-0.69.17-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.8.2](https://img.shields.io/badge/AppVersion-2.8.2-informational?style=flat-square)
 
 Helm chart for Grafana Loki in microservices mode
 
@@ -173,6 +173,7 @@ kubectl delete statefulset RELEASE_NAME-loki-distributed-querier -n LOKI_NAMESPA
 | gateway.nginxConfig.httpSnippet | string | `""` | Allows appending custom configuration to the http block |
 | gateway.nginxConfig.logFormat | string | See values.yaml | NGINX log format |
 | gateway.nginxConfig.resolver | string | `""` | Allows overriding the DNS resolver address nginx will use. |
+| gateway.nginxConfig.serverListens | string | `"listen             8080;"` | Configure the server listener(s), this needs to be customised for IPv6 only clusters |
 | gateway.nginxConfig.serverSnippet | string | `""` | Allows appending custom configuration to the server block |
 | gateway.nodeSelector | object | `{}` | Node selector for gateway pods |
 | gateway.podAnnotations | object | `{}` | Annotations for gateway pods |
--- a/charts/loki-distributed/values.yaml
+++ b/charts/loki-distributed/values.yaml
@@ -1030,6 +1030,9 @@ gateway:
       main '$remote_addr - $remote_user [$time_local]  $status '
               '"$request" $body_bytes_sent "$http_referer" '
               '"$http_user_agent" "$http_x_forwarded_for"';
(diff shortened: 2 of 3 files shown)


zanhsieh:
@stevehipwell 
Just curious, how did you install and test this in your local machine? I plan for rework on the CI build for this.

stevehipwell:
@zanhsieh I tested it from my local machine.

zanhsieh:
> @zanhsieh I tested it from my local machine.

I mean, how did you set up your local stack as IPv6 only?

stevehipwell:
> I mean, how did you set up your local stack as IPv6 only?

I didn't, I'm not sure it's possible, I hit an IPv6 EKS cluster.

stevehipwell:
@unguiculus @Whyeasy could one of you give this a review?

charts/loki-distributed/values.yaml
```diff
@@ -943,6 +943,9 @@ gateway:
       main '$remote_addr - $remote_user [$time_local]  $status '
               '"$request" $body_bytes_sent "$http_referer" '
               '"$http_user_agent" "$http_x_forwarded_for"';
+    # -- Configure the server listener(s), this needs to be customised for IPv6 only clusters
+    serverListens: |-
+      listen             8080;
```
zanhsieh, line 948: I think we should make it clear that this stanza is for IPv6, and IP6Enabled to enable / disable it (default is false).

```suggestion
    IP6serverListens: |-
      listen             8080;

    # -- Enable IPv6 only clusters configure
    IP6Enabled: false
```

charts/loki-distributed/README.md
```diff
@@ -153,6 +153,7 @@ kubectl delete statefulset RELEASE_NAME-loki-distributed-querier -n LOKI_NAMESPA
 | gateway.nginxConfig.httpSnippet | string | `""` | Allows appending custom configuration to the http block |
 | gateway.nginxConfig.logFormat | string | See values.yaml | NGINX log format |
 | gateway.nginxConfig.resolver | string | `""` | Allows overriding the DNS resolver address nginx will use. |
+| gateway.nginxConfig.serverListens | string | `"listen             8080;"` | Configure the server listener(s), this needs to be customised for IPv6 only clusters |
```
zanhsieh, line 156:
```suggestion
| gateway.nginxConfig.IP6serverListens | string | `"listen             8080;"` | Configure the server listener(s), this needs to be customised for IPv6 only clusters |
| gateway.nginxConfig.IP6Enabled | bool | `false` | Enable IPv6 only clusters configure |
```

zanhsieh (changes requested):

charts/loki-distributed/values.yaml
stevehipwell: @zanhsieh the current use of this block is for IP family so I could get behind using `IpFamilyServerListens` but I really don't see the benefit of that form over `serverListens`. Could you explain in a bit more detail why you think we need to explicitly tie the block to the current implementation while the listens could be modified for reasons other than just switching IP family?

stevehipwell: @zanhsieh why don't we add more comments and leave it as it is, for testing purposes you can just prove that if the user leaves it alone it's a no-op? AFAIK the rest of the Nginx config is like this.

I'd have loved to be able to provide a higher level IP family argument to hide the complexity from the end users but this isn't possible as the Loki Distributed chart [won't start on IPv6 cluster](https://github.com/grafana/helm-charts/issues/157) without a hack (see https://github.com/grafana/dskit/pull/185 for the actual fix) and even when it's running it doesn't [actually work](https://github.com/grafana/loki/issues/5578).

mossad-zika:
so?

zanhsieh:
> so?

@mossad-zika 
If the original author can't provide any way to test this PR and take care backward compatibility, I won't let it pass. Other maintainer(s) might let it passed.

stevehipwell:
@zanhsieh I did test the chart and the PR is a no-op so I'm confused about your attitude to the change.

zalegrala:
It might be a good idea to wait for the dskit pr to be merged upstream, and then integrated into loki, just in case we decide to rename a config variable or some such.  Love the effort to test this out using the chart though, pretty nice.

stevehipwell:
@zalegrala this PR is intended to allow the gateway to support IPv6 but is actually a generic change which results in a no-op unless utilised so it could me merged immediately which would allow easier testing of the dskit changes.

zalegrala:
Oh nice @stevehipwell, makes sense to support ahead of the dskit pr.

stevehipwell:
Could someone give this a review please?

PR #1900, opened 2022-10-18
[tempo] Add pullSecrets to tempo and tempo-query

cles:
Signed-off-by: Oriol Francès <me@oriolfranc.es>


cles:
This adds an optional `imagePullSecrets` key the tempo chart (as suggested by the comments on the values.yml file itself)

PR #1910, opened 2022-10-20
[grafana] typo in the comment for initChownData in Grafana helm values

akshaykrjain:
This chart is for Grafana but the comment says prometheus.

Looks like a copy-paste issue. Fixing to avoid confusion.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.42.2
+version: 6.42.3
 appVersion: 9.2.0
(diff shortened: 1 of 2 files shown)


torstenwalter (commented):
You also need to bump the chart version. Increasing the patch version by one should do.

akshaykrjain:
> You also need to bump the chart version. Increasing the patch version by one should do.

Bumped to 6.42.3 with this [commit](https://github.com/grafana/helm-charts/pull/1910/commits/95b327c51de65a65841f1ae6682f8780c760d3d7)

zanhsieh (approved):

Issue #1916, opened 2022-10-21
[promtail] Option to deploy config as a ConfigMap


verejoel:
Working with a secret for the config is quite clunky, especially when we can utilize the `-config.expand-env` flag to inject secrets through environment variables. I would propose to add an option to deploy the configuration as a ConfigMap instead of a secret.

verejoel:
Solved with #1918

Issue #1915, opened 2022-10-21
[promtail] ServiceMonitor does not allow many essential configuration options


verejoel:
ServiceMonitor does not currently allow TLS, target labels, or scheme to be modified by the user of the chart.

verejoel:
Solved with #1918

PR #1918, opened 2022-10-21
[promtail] Update ServiceMonitor template and introduce config as ConfigMap option

verejoel:
This pull request solves issues #1915 and #1916.

 - The Promtail config can now optionally be generated as a ConfigMap instead of a secret. This makes hacking with `kubectl` much less of a chore. If you still need secrets in your config, consider injecting them through environment variables using `extraEnvFrom` chart value and setting `extraArgs: ["-config.expand-env=true"]`.
 - ServiceMonitor was missing some crucial settings, in particular `tlsConfig`. The ServiceMonitor template has now been updated to match that used for the `loki-distributed` chart.

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.6.1
-version: 6.5.1
+version: 6.6.0
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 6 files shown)


verejoel:
Hi @jeschkies, thanks for taking a look!

I made the default value for the new `configmap.enabled` value `false`, which means this will be backwards compatible - i.e. if a user upgrades chart version and doesn't set a value for `configmap.enabled`, the credentials will still be stored in a secret. Only by explicitly setting `configmap.enabled` to `true`, could the credentials be exposed - at which point we have to trust the user knows what they are doing.

verejoel:
Seems to be failing as I also added some new info to the docs

verejoel:
I did, but I also added some new lines which are not generated. this is the
reason why the pipeline is failing, the documentation for the chart bump
and API change is there

verejoel:
I can remove the lines but I think it’s useful to see in the readme how to
use the new ConfigMap option

verejoel:
@jeschkies pipeline should be fixed. I think it would be beneficial to include the update to the docs in the merge commit, but I leave that decision up to you.

zanhsieh (approved):

PR #1921, opened 2022-10-21
[grafana] feat: allow configuring custom labels for PVCs

### LINKED ISSUE #1920: [grafana] Allow setting custom PVC labels
Hi, it would be nice to be able to specify custom labels for the templated Grafana PVC, in order to be able to select them for other purposes (e.g. for backups).

It's theoretically possible to do this using `extraLabels`, however these extra labels are then applied to a dozen resources, which is not ideal.

FelixKampfer:
This MR fixes #1920

### DIFF
--- a/charts/grafana/templates/pvc.yaml
+++ b/charts/grafana/templates/pvc.yaml
@@ -6,6 +6,9 @@ metadata:
   namespace: {{ template "grafana.namespace" . }}
   labels:
     {{- include "grafana.labels" . | nindent 4 }}
+    {{- if .Values.persistence.extraPvcLabels }}
+    {{- toYaml .Values.persistence.extraPvcLabels | nindent 4 }}
+    {{- end }}
   {{- with .Values.persistence.annotations  }}
   annotations:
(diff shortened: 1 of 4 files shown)


charts/grafana/templates/pvc.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   namespace: {{ template "grafana.namespace" . }}
   labels:
     {{- include "grafana.labels" . | nindent 4 }}
+    {{- if .Values.persistence.extraPvcLabels }}
+    {{- toYaml .Values.persistence.extraPvcLabels | nindent 4 }}
+    {{- end }}
```
zanhsieh, line 11: Since by default `.Values.persistence.extraPvcLabels` is `{}`, we should avoid it to show up.

```suggestion
    {{- with .Values.persistence.extraPvcLabels }}
    {{- toYaml . | nindent 4 }}
    {{- end }}
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #1923, opened 2022-10-22
[loki-distributed] Do not use `replicas` if ingester statefulset autoscaling enabled

khfayzullaev:
Even if the ingester statefulset autoscaling is enabled, the statefulset uses the `replicas` value. This change ensures that the `replicas` is not used if the autoscaling is enabled.


khfayzullaev:
fixed in https://github.com/grafana/helm-charts/commit/34a6df47104c5882d745bb6149a32888b53b00b4

Issue #1936, opened 2022-10-30
[loki-distributed] memcached statefulsets should include persistence option


verejoel:
The memcached stateful sets should also allow persistence to be enabled. Doing so would configure a `volumeClaimTemplate` in the stateful set template and set the `--memory-file` argument, allowing the cache to be restored if a pod fails.

Issue #1386, opened 2022-05-18
[loki-distributed] Memcached is not used even if enabled in helm


kovaxur:
Hi,
If I enabled the memcached caches in the Loki-distributed helm's values.yml, it only creates the memcached pods, however Loki will not start to use it.

If I manually set the host in the configuration of the memcached_client, it works as expected, for example:

```        
storage_config:
  index_queries_cache_config:
    memcached_client:
      consistent_hash: true
      host: loki-loki-distributed-memcached-index-queries.XXX.svc.cluster.local
      service: http
```

verejoel:
This is solved with #1939.

PR #1939, opened 2022-10-31
[loki-distributed] wire-in memcached automatically when enabled, introduce memcached persistence

verejoel:
## Memcached configuration
Currently if memcached is enabled through the `memcachedXXX` values, the infrastructure will be deployed, but Loki is not configured to use the created resources.

This PR adds new entries to the default `loki.config` value, which minimally configures each memcached component when enabled (solving #1386).

Additionally cache configuration can still be injected and overwritten using the `structuredConfig` value.

## Memcached persistence
This PR also adds the ability to configure optional persistence for the memcached stateful sets, meaning cache data will not be lost upon pod restart. This solves issue #1936.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.6.1
-version: 0.63.2
+version: 0.64.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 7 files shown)


verejoel:
@jeschkies @zanhsieh @trevorwhitney this one is ready to go, tested in our setup. It should significantly improve the user experience when using memcached with this chart. Sorry for directly tagging you but the codeowners don't seem to check these PRs, so noone else will be requested for review.

zanhsieh (approved):

verejoel:
Thanks for the review @zanhsieh, can we merge it now?

PR #1947, opened 2022-11-03
[tempo-distributed] add headless services for ingester and metrics-generator

AlphaStyle:
This pull request will add [headless services](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) for the `ingester` and `metrics-generator`. The `query-frontend` already has a headless service. Tried to follow the same name convention as the `query-frontend` and named the new headless services with the suffix *-discovery*.

Headless services for the `ingester` and the `metrics-generator` are required when used with service-mesh solutions such as [Istio](https://istio.io/). The gRPC connections are established with internal Kubernetes endpoint IP's which Istio are not aware of and will therefore block those connections. Adding headless services will resolve this Istio issue.

**This pull request include the optional `appProtocol` option and should therefore be merged after pull request** https://github.com/grafana/helm-charts/pull/1884 **!**

```
{{- if .Values.ingester.appProtocol.grpc }}
appProtocol: {{ .Values.ingester.appProtocol.grpc }}
{{- end }}
 ```

### DIFF
--- a/charts/tempo-distributed/templates/ingester/service-ingester-discovery.yaml
+++ b/charts/tempo-distributed/templates/ingester/service-ingester-discovery.yaml
@@ -1,10 +1,11 @@
+{{- $dict := dict "ctx" . "component" "ingester" true }}
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "tempo.resourceName" (dict "ctx" . "component" "ingester") }}-discovery
+  name: {{ template "tempo.resourceName" $dict }}-discovery
   namespace: {{ .Release.Namespace }}
   labels:
-    {{- include "tempo.labels" (dict "ctx" . "component" "ingester") | nindent 4 }}
+    {{- include "tempo.labels" $dict | nindent 4 }}
     prometheus.io/service-monitor: "false"
   {{- with .Values.ingester.service.annotations }}
   annotations:
@@ -26,4 +27,4 @@ spec:
       appProtocol: {{ .Values.ingester.appProtocol.grpc }}
       {{- end }}
   selector:
-    {{- include "tempo.selectorLabels" (dict "ctx" . "component" "ingester") | nindent 4 }}
+    {{- include "tempo.selectorLabels" $dict | nindent 4 }}
--- a/charts/tempo-distributed/templates/ingester/service-ingester-discovery.yaml
+++ b/charts/tempo-distributed/templates/ingester/service-ingester-discovery.yaml
@@ -1,0 +1,29 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "tempo.resourceName" (dict "ctx" . "component" "ingester") }}-discovery
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "tempo.labels" (dict "ctx" . "component" "ingester") | nindent 4 }}
+    prometheus.io/service-monitor: "false"
+  {{- with .Values.ingester.service.annotations }}
+  annotations:
+  {{- toYaml . | nindent 4 }}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by zalegrala** [comment]
Thanks for the PR.  How will this be used?  In the case of the query-frontend, the queriers use this address so that they can make a connection to each of the frontends, so as not to make a single connection to a single frontend, but I'm not sure the use case here.  What did you have in mind?

**Comment by AlphaStyle** [comment]
From the Kubernetes [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) documentation:
> You can use a headless Service to interface with other service discovery mechanisms, without being tied to Kubernetes' implementation.

When pods are using an Istio sidecar ([Envoy proxy](https://www.envoyproxy.io/)) the traffic will be managed by Istio. I noticed the connection between Tempo pods failed because Istio did not know where to route them. Adding a headless service resolved the connection issue as Istio will allow traffic to be routed directly to the [endpoints](https://kubernetes.io/docs/concepts/services-networking/service/#endpoints) IPs.

Loki-distributed also have a headless service for the ingester ([link](https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/templates/ingester/service-ingester-headless.yaml)).

I appreciate any feedback or suggestions :slightly_smiling_face:

File: charts/tempo-distributed/templates/ingester/service-ingester-discovery.yaml [hunk 1]
```diff
@@ -0,0 +1,29 @@
...
+      targetPort: 3100
+    - name: grpc
+      port: 9095
+      protocol: TCP
+      targetPort: 9095
+      {{- if .Values.ingester.appProtocol.grpc }}
+      appProtocol: {{ .Values.ingester.appProtocol.grpc }}
+      {{- end }}
+  selector:
+    {{- include "tempo.selectorLabels" (dict "ctx" . "component" "ingester") | nindent 4 }}
```
zalegrala, line 29: You could setup a `$dict` here as in the other file as well.

charts/tempo-distributed/templates/metrics-generator/service-metrics-generator-discovery.yaml
```diff
@@ -0,0 +1,32 @@
...
+      port: {{ .port }}
+      protocol: TCP
+      targetPort: {{ .port }}
+      {{- if and (hasPrefix .name "grpc") ($.Values.metricsGenerator.appProtocol.grpc) }}
+      appProtocol: {{ $.Values.metricsGenerator.appProtocol.grpc }}
+      {{- end }}
+    {{- end }}
+    {{- end }}
+  selector:
+    {{- include "tempo.selectorLabels" (dict "ctx" . "component" "metrics-generator") | nindent 4 }}
```
zalegrala, line 31: I think passing `$dict` here would work for the selector also.

zalegrala (commented):
Oh interesting.  I'll have to read up on how that works with DNS.

This looks pretty good to me.  Left a couple notes.  A rebase might make CI happy.

charts/tempo-distributed/templates/metrics-generator/service-metrics-generator-discovery.yaml
AlphaStyle: I didn't even notice I could `$dict` here :astonished:

AlphaStyle:
I took the liberty to use the `$dict` setup in the two other non-headless `ingester` and `metrics-generator` services to keep it consistent.

zalegrala:
Nice, looks good.  I appreciate the liberties taken :)

zalegrala (approved):

PR #1956, opened 2022-11-08
[loki-distributed] Add HPA support for k8s >=1.25-0

MioOgbeni:
Hi, we migraded from loki-stack to loki-distributed helm chart, but our k8s cluster is running on 1.25.x version already. 

I ran into an issue with an unsupported apiVersion for HorizontalPodAutoscaler in loki-distributed helm chart, so I decided to add support for the autoscaling/v2 apiVersion.

Signed-off-by: Tomáš Novák <tomas.novak@bcas.cz>

### DIFF
--- a/charts/loki-distributed/templates/_helpers.tpl
+++ b/charts/loki-distributed/templates/_helpers.tpl
@@ -141,6 +141,17 @@ Return the appropriate apiVersion for PodDisruptionBudget.
   {{- end -}}
 {{- end -}}
 
+{{/*
+Return the appropriate apiVersion for HorizontalPodAutoscaler.
+*/}}
+{{- define "loki.hpa.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "autoscaling/v2") (semverCompare ">=1.25-0" .Capabilities.KubeVersion.Version) -}}
+    {{- print "autoscaling/v2" -}}
+  {{- else -}}
+    {{- print "autoscaling/v2beta1" -}}
+  {{- end -}}
+{{- end -}}
+
 {{- define "loki.ingester.readinessProbe" -}}
 {{- with .Values.ingester.readinessProbe }}  
 readinessProbe:
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.6.1
-version: 0.66.5
+version: 0.66.6
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.6.1
-version: 0.65.0
+version: 0.65.1
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
--- a/charts/loki-distributed/README.md
+++ b/charts/loki-distributed/README.md
@@ -1,6 +1,6 @@
 # loki-distributed
 
-![Version: 0.66.5](https://img.shields.io/badge/Version-0.66.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.6.1](https://img.shields.io/badge/AppVersion-2.6.1-informational?style=flat-square)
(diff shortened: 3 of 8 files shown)


charts/loki-distributed/templates/_helpers.tpl
```diff
@@ -141,6 +141,17 @@ Return the appropriate apiVersion for PodDisruptionBudget.
   {{- end -}}
 {{- end -}}
 
+{{/*
+Return the appropriate apiVersion for HorizontalPodAutoscaler.
+*/}}
+{{- define "loki.hpa.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "autoscaling/v2") (semverCompare ">=1.25-0" .Capabilities.KubeVersion.Version) -}}
```
zanhsieh, line 148: Why don't we just query api versions has `autoscaling/v2/HorizontalPodAutoscaler` ?

```suggestion
  {{- if .Capabilities.APIVersions.Has "autoscaling/v2/HorizontalPodAutoscaler -}}
```

Proof:

```
$ kind create cluster
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.25.0) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/
$ k api-resources | grep autoscaling/v2
horizontalpodautoscalers          hpa          autoscaling/v2                         true         HorizontalPodAutoscaler
```

zanhsieh (changes requested):
Also please bump the version in Chart.yaml and sign DCO. Thank you.

zanhsieh:
@MioOgbeni 
BTW, I think there're something wrong with which api version uses `HorizontalPodAutoscaler`(HPA). HPA is not only available since k8s v1.25.x, but 1.24.x. The previous k8s version supported not necessary is `v2beta1` but `v1`.

```
$ kind create cluster --image kindest/node:v1.24.6
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.24.6) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Not sure what to do next? 😅  Check out https://kind.sigs.k8s.io/docs/user/quick-start/
$ k api-resources | grep autoscaling/v2
horizontalpodautoscalers          hpa          autoscaling/v2                         true         HorizontalPodAutoscaler

$ kind delete cluster
Deleting cluster "kind" ...

$ kind create cluster --image kindest/node:v1.22.15
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.22.15) 🖼
 ✓ Preparing nodes 📦
 ✓ Writing configuration 📜
 ✓ Starting control-plane 🕹️
 ✓ Installing CNI 🔌
 ✓ Installing StorageClass 💾
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Have a question, bug, or feature request? Let us know! https://kind.sigs.k8s.io/#community 🙂

$ k api-resources | grep HorizontalPodAutoscaler
horizontalpodautoscalers          hpa          autoscaling/v1                         true         HorizontalPodAutoscaler
```

charts/loki-distributed/templates/_helpers.tpl
MioOgbeni: Yes, it can be done this way too, but I wanted to use the same syntax as in the already existing define apiVersion code blocks. I can rework it to the desired state, but is this really the wanted change @zanhsieh? Or I can refactor each occurrence to the desired state to preserve consistency.

charts/loki-distributed/templates/_helpers.tpl
```diff
@@ -141,6 +141,17 @@ Return the appropriate apiVersion for PodDisruptionBudget.
   {{- end -}}
 {{- end -}}
 
+{{/*
```
diranged, line 144: Pretty please, add in support to customize the `behavior` on the autoscalers too. We find that we cannot slow down the scale up/down behavior and that causes problems for us.

1. We can't create our own HPAs because as soon as you set `ingester.autoscaling.enabled: false`, then `replicas: ...` is set explicitly.
2. We can't pass in the `behavior` settings we want to slow down the autoscaling.

charts/loki-distributed/templates/_helpers.tpl
MioOgbeni: I really only want to make possible to use HPA on the k8s >=1.25 without huge changes and with same functionality. Please can you create enhancement issue for your request? I will try to focused on it later and add also another new features as custom and multiple metrics.

zanhsieh (approved):

z0rc:
I would strongly advice to relax kubernetes version comparison to 1.23. See https://helm.sh/docs/topics/kubernetes_apis/

> Recommendation: The best practice is to upgrade releases using deprecated API versions to supported API versions, prior to upgrading to a kubernetes cluster that removes those API versions.

PR #1964, opened 2022-11-10
[tempo-distributed] move dictionary handling to avoid nil pointer

zalegrala:
Signed-off-by: Zach Leslie <zach.leslie@grafana.com>


swartz-k:
LGTM. In old version, when gateway and gateway.ingress enabled, `nil pointer evaluating interface {}.fullnameOverride` will happen.

faustodavid (approved):

PR #1967, opened 2022-11-11
[tempo-distributed] fix wrong context on ingress-gateway

daper:
I was having the following error during the chart render.

```
template: tempo-distributed/templates/gateway/ingress-gateway.yaml:45:25: executing "tempo-distributed/templates/gateway/ingress-gateway.yaml" at <include "tempo.resourceName" (dict "ctx" . "component" "gateway")>: error calling include: template: tempo-distributed/templates/_helpers.tpl:153:3: executing "tempo.resourceName" at <include "tempo.fullname" .ctx>: error calling include: template: tempo-distributed/templates/_helpers.tpl:15:14: executing "tempo.fullname" at <.Values.fullnameOverride>: nil pointer evaluating interface {}.fullnameOverride
```

It is caused by a missing context due to the `range` in [ingress-gateway.yaml:37](https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/templates/gateway/ingress-gateway.yaml#L37). I've followed the same pattern seen on other templates here and defined a `$dict` on top which is then passed as context.


daper:
Duplicated of [pull request 1964](https://github.com/grafana/helm-charts/pull/1964)

PR #1972, opened 2022-11-14
[tempo] block_retention instead of compacted_block_retention

### LINKED ISSUE #1970: Tempo retention settable via values file
Tempo values file allows to set `retention` which translates to `compactor.compaction.compacted_block_retention` in the tempo config map.
But this value is not the value that sets how long the traces are kept, it is the length of time a compactor will leave a block in object storage after it has marked it for deletion. It is a low level setting the probably few people needs to configure via values file.
Instead, there is another setting  `compactor.compaction.block_retention` that defines the actual duration to keep blocks. And this is the setting needed to prevent disk from filling up.

I propose to add a setting in values file that controls it, as the existing `retention` needs probably to stay for backward compatibility.
See also this comment: https://github.com/grafana/tempo/issues/1884#issuecomment-1310295822

faustodavid:
Tempo chart uses `compacted_block_retention` for the retention and tempo-distributed uses `block_retention`. I am updating the tempo chart to match the tempo-distributed.

We probably shouldn't update the `compacted_block_retention` as @joe-elliott mentioned here
https://github.com/grafana/tempo/issues/1884#issuecomment-1310295822

Fixes #1970

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 0.16.3
+version: 0.16.4
 appVersion: 1.5.0
 engine: gotpl
 home: https://grafana.net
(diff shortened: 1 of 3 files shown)


joe-elliott (approved):

faustodavid:
Thanks @zanhsieh for updating the branch! :raised_hands:

PR #1971, opened 2022-11-14
[grafana] refactor

zanac1986:
Signed-off-by: zanac1986 <zanhsieh@protonmail.com>

PR #2016, opened 2022-11-28
Fix a regression introduced in 6.44.0 / #1971

elliottmb:
Previously both the statefulset and this file checked if it equals "statefulset". Now the statefulset checks against this list, meaning they don't line up together.


zanhsieh:
@elliottmb 
Thank you for the contribution. Please bump the version in Chart.yaml to latest and sign DCO.

elliottmb:
@zanhsieh Done and done.

Issue #1946, opened 2022-11-02
[tempo-distributed] centralize port handling


zalegrala:
Currently, the various network ports that are used in this chart are scattered over the templates.  I think it would be a good idea to centralize the values here, and then reference those values from the various locations.  This should allow environments to tune their ports a little more easily, and also pave the way for changing the default if we so choose.

faustodavid:
Will this list also include the trace receiver ports?

Example:
```yaml
ports:
  tempo:
    httpMetrics:
      containerPort: 3100
      servicePort: 3100
      protocol: TCP
    httpMemberlist:
      containerPort: 7946
      protocol: TCP
    grpc:
      containerPort: 9095
      servicePort: 9095
      protocol: TCP
    ...
  traces:
    otlp:
      enabled: true
      containerPort: 4317
      servicePort: 4317
      hostPort: 4317
      protocol: TCP
    otlp-http:
      enabled: true
      containerPort: 4318
      servicePort: 4318
      hostPort: 4318
      protocol: TCP
    ...
```

zalegrala:
I think it makes sense to include any duplicate port specs, so I would lump in the receiver ports also.  Good point.

If any community members want to pick up any of the issues I've filed, please feel free.

faustodavid:
I am going to pick this one :)

zalegrala:
Hey @faustodavid, are you still thinking of pushing this one forward?  No pressure, but I remembered it today after I had to modify a port and thought it might be good to have.

faustodavid:
hey @zalegrala yes, I will push this one forward :)

PR #2043, opened 2022-12-05
[tempo-distributed] Centralize port handling

faustodavid:
This is a draft for #1946.


faustodavid:
I still doubt centralizing the ports using this approach because the values file becomes verbose. Maybe I can create a definition of common ports or have the ports definition in each service, using a similar approach like in the metrics generator.  

What do you think [zalegrala](https://github.com/zalegrala) ?

PR #2046, opened 2022-12-06
[tempo-distributed] Add capability to define additional ports on gateway service

eviscares:
This creates a way to use the gateway for more complex scenarios in combination with .Values.gateway.nginxConfig.httpSnippet.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.27.10
+version: 0.27.14
 appVersion: 1.5.0
 engine: gotpl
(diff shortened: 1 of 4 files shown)


BitProcessor:
If this is a useful scenario, it would be great to have it documented a bit more?
```
    # -- Additional ports that might be required on the gateway service
     additionalPorts: {}
```
Is a bit vague imho.
"ports that might be required" is as saying:
"Here is a car, and here are some screws that might be required"
Euh, yeah, thanks I guess 🤷🏻‍♂️

eviscares:
Hey, I updated the comment string a bit, but I am unsure on how to make it clearer, as it is a pretty flexible thing.

mapno:
It needs updating the docs, please run helm-docs and commit changes!

eviscares:
I'm trying. XD Should be good now

mapno (approved):
LGTM, thanks for the patience with the docs.

PR #2049, opened 2022-12-06
[loki-distributed] Update HPAs to support autoscaling/v2 schema

z0rc:
Followup to https://github.com/grafana/helm-charts/pull/2038

I forgot that there were schema changes between `autoscaling/v2beta1` and `autoscaling/v2`. Updated HPA templates to support old and new API version. Otherwise rendering of chart with any `...autoscaling.enabled=true` will fail with
```
Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(HorizontalPodAutoscaler.spec.metrics[0].resource): unknown field "targetAverageUtilization" in io.k8s.api.autoscaling.v2.ResourceMetricSource, ValidationError(HorizontalPodAutoscaler.spec.metrics[0].resource): missing required field "target" in io.k8s.api.autoscaling.v2.ResourceMetricSource]
```

Signed-off-by: Ihor Urazov <iurazov@healthjoy.com>

### DIFF
--- a/charts/loki-distributed/templates/ingester/hpa.yaml
+++ b/charts/loki-distributed/templates/ingester/hpa.yaml
@@ -1,5 +1,6 @@
 {{- if .Values.ingester.autoscaling.enabled }}
-apiVersion: {{ include "loki.hpa.apiVersion" . }}
+{{- $apiVersion := include "loki.hpa.apiVersion" . -}}
+apiVersion: {{ $apiVersion }}
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ include "loki.ingesterFullname" . }}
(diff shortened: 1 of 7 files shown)


z0rc:
@zanhsieh PTAL.

zanhsieh:
@z0rc 
Just curious, how did you stress test HPA? Continuously call which API?

z0rc:
@zanhsieh I'm not sure I understand your question. Are you referring to updating HPA resources via API or frequency of changes made by HPA?

I believe first is no different from updating any other kubernetes resources. For second I don't have a good answer. Maybe frequently update average Utilization value targets and watch HPA status? Also `behavior` part of HPA could be taken into account, albeit it isn't set here.

z0rc:
As of HPA behaviour, AFAIK ultimately it just updates deployment replicas number, so HPA behavior can be replicated by manually updating this field in deployment spec.

zanhsieh:
@z0rc 
The reason why i asked was, if I approve HPA, then later the users might raise issues why the Deployment does not scale out and / or in what case the Deployment will scale out. Presenting some test case would be better idea.

z0rc:
@zanhsieh I only changed existing support of HPA to work with up to date API version. I tested localy, if that matters. But I'm not sure how to address your question in this PR, as HPA was added prior to this MR and I don't know what kind of stress test you expect to have.

z0rc:
@zanhsieh sure. See my initial PR message, the error is the same as in linked ticket, just formatted a bit different.

charts/loki-distributed/templates/ingester/hpa.yaml
```diff
@@ -17,12 +18,24 @@ spec:
     - type: Resource
       resource:
         name: memory
+        {{- if (eq $apiVersion "autoscaling/v2") }}
+        target:
+          type: Utilization
+          averageValue: {{ . }}
```
zaldnoay, line 24: Is this a mistake? This field in other files is `averageUtilization`.

z0rc: Thanks, fixed.

PR #2062, opened 2022-12-09
[grafana] Add ability to reload dashboard provisioning

Sumoa:
I noticed that when using the dashboards sidecar to provision dashboards no automatic reload is configured. Meanwhile the alerts and datasources sidecars are configured with the automatic reload.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.47.0
+version: 6.47.1
 appVersion: 9.3.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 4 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.47.0
+version: 6.47.1
```
zanhsieh, line 3: New feature should bump to minor version.

```suggestion
version: 6.48.0
```

zanhsieh (changes requested):

charts/grafana/Chart.yaml
Sumoa: @zanhsieh It seems that when you merged main into this branch you already bumped to version 6.48.0.

Should I bump to 6.49.0 or can I resolve this?

zanhsieh: 6.48.0 should be fine.

PR #2064, opened 2022-12-12
[promtail] Allow to Configure Promtail Namespace

Kirchen99:
This pull request makes it possible to configure the namespace the promtail to deploy outsides `.Release.Namespace`. It is useful if the user wrap all logging helm charts as an umbrella chart and want to deploy promtail into a different namespace.

Signed-off-by: Kirchen99 <latias_latios@126.com>

### DIFF
--- a/charts/promtail/values.yaml
+++ b/charts/promtail/values.yaml
@@ -118,10 +118,9 @@ rbac:
   # -- Specifies whether a PodSecurityPolicy is to be created
   pspEnabled: false
 
-namespace:
-  # -- The name of the Namespace to deploy
-  # If not set, `.Release.Namespace` is used
-  name: null
+# -- The name of the Namespace to deploy
+# If not set, `.Release.Namespace` is used
+namespace: null
 
 serviceAccount:
   # -- Specifies whether a ServiceAccount should be created
--- a/charts/promtail/values.yaml
+++ b/charts/promtail/values.yaml
@@ -118,6 +118,11 @@ rbac:
   # -- Specifies whether a PodSecurityPolicy is to be created
   pspEnabled: false
 
+namespace:
+  # -- The name of the Namespace to deploy
+  # If not set, `.Release.Namespace` is used
+  name: null
+
 serviceAccount:
   # -- Specifies whether a ServiceAccount should be created
   create: true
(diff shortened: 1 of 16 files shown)


charts/promtail/values.yaml
```diff
@@ -118,6 +118,11 @@ rbac:
   # -- Specifies whether a PodSecurityPolicy is to be created
   pspEnabled: false
 
+namespace:
+  # -- The name of the Namespace to deploy
+  # If not set, `.Release.Namespace` is used
+  name: null
```
zanhsieh, line 124: 1. I would prefer flat value. Just simple `namespace` should be fine since I don't see how the user could customize the namespace, e.g. set label / annotation in the namespace. See: 
https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values

2. Did this PR pass the following test?
```console
helm install promtail grafana/promtail -n promtail --create-namespace
```

zanhsieh (changes requested):

charts/promtail/values.yaml
Kirchen99: Thank you for your feedback.

1. OK, it has been changed to flat value.
2. Yes, this PR passed the following test:

```
➜  promtail git:(configure-promtail-namespace) helm install promtail grafana/promtail -n promtail --create-namespace
NAME: promtail
LAST DEPLOYED: Mon Jan  2 11:39:23 2023
NAMESPACE: promtail
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
***********************************************************************
 Welcome to Grafana Promtail
 Chart version: 6.8.0
 Promtail version: 2.7.0
***********************************************************************

Verify the application is working by running these commands:
* kubectl --namespace promtail port-forward daemonset/promtail 3101
* curl http://127.0.0.1:3101/metrics
➜  promtail git:(configure-promtail-namespace) helm ls -n promtail
NAME    	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART         	APP VERSION
promtail	promtail 	1       	2023-01-02 11:39:23.786706 +0100 CET	deployed	promtail-6.8.0	2.7.0      
```

zanhsieh (approved):

PR #2069, opened 2022-12-15
[tempo] Allow adding additional ServiceAccount labels

LambArchie:
I need to add additional labels to the service account so I can use Azure Workload Identity

Signed-off-by: Archie Lamb <git@archielamb.uk>


joe-elliott (approved):
Nice! Thanks for the improvement.

PR #2077, opened 2022-12-22
[loki-stack] Fix url in promtail config to point to loki service.

ne-msft:
.Release.Name is an arbitrary name provided by whoever is installing the chart. it is used to build the final name for the service, which is {{ . Release.Name }}-loki.

Currently promtail fails pushing data to loki.

Signed-off-by: Nico Erfurth <nico.erfurth@microsoft.com>


zanhsieh (changes requested):
@ne-msft 
Thank you for contribution. Could you bump the version in Chart.yaml and sign DCO please?

ne-msft:
Looking at the checks, they now seem to fail BECAUSE they can reach the actual endpoint, instead of something none existing.

PR #2078, opened 2022-12-22
[promtail] Change in promtail config to make it work for GKE Autopilot

ashmitakapoor:
GKE Autopilot only allows readOnly access to hostpath /var/log as a security consideration. It also only supports containerd as a container runtime. Therefore, the log volumes need to be modified to mount the correct directory in readOnly mode.
Refer - https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-security#built-in-security

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.27.12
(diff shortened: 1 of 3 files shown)


zanhsieh:
@ashmitakapoor 
Thank you for the contribution. Could you bump version in Chart.yaml and sign DCO please?

zanhsieh (changes requested):
@ashmitakapoor
Thank you for the contribution. Could you bump version in Chart.yaml and sign DCO please?

PR #2082, opened 2022-12-28
[tempo-distributed] feat: Enable usage of private docker registry

jhlubek:
There are many reasons to not use docker.io public registry, e.g.:
- docker.io pull rate limit is really low and this release usually pulls 10 or more images. Ref: https://www.docker.com/increase-rate-limits/
- to use modified images from private registry

This can be enabled by simply adding imagePullSecrets

ref: https://kubernetes.io/docs/concepts/containers/images/#specifying-imagepullsecrets-on-a-pod

### DIFF
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -35,6 +35,19 @@ Docker image selector for Tempo. Hierachy based on global, component, and tempo 
 {{- end -}}
 
 {{/*
+Optional list of imagePullSecrets for Tempo docker images
+*/}}
+{{- define "tempo.imagePullSecrets" -}}
+{{- $imagePullSecrets := coalesce .global.pullSecrets .component.pullSecrets .tempo.pullSecrets -}}
+{{- if $imagePullSecrets  -}}
+imagePullSecrets:
+{{- range $imagePullSecrets }}
+  - name: {{ . }}
+{{ end }}
+{{- end }}
+{{- end -}}
+
+{{/*
 Create chart name and version as used by the chart label.
 */}}
 {{- define "tempo.chart" -}}
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 0.27.15
+version: 0.27.16
 appVersion: 1.5.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
--- a/charts/tempo-distributed/README.md
+++ b/charts/tempo-distributed/README.md
@@ -1,6 +1,6 @@
 # tempo-distributed
 
-![Version: 0.27.15](https://img.shields.io/badge/Version-0.27.15-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 1.5.0](https://img.shields.io/badge/AppVersion-1.5.0-informational?style=flat-square)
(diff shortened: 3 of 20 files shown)


charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -34,6 +34,19 @@ Docker image selector for Tempo. Hierachy based on global, component, and tempo
 {{- printf "%s/%s:%s" $registry $repository $tag -}}
 {{- end -}}
 
+{{/*
+Optional list of imagePullSecrets for Tempo docker images
+*/}}
+{{- define "tempo.imagePullSecrets" -}}
+{{- $imagePullSecrets := coalesce .global.pullSecrets .component.pullSecrets .tempo.pullSecrets -}}
```
joe-elliott, line 41: looking for clarification on "global" vs "tempo" pullsecrets. is there a reason to have both?

joe-elliott (commented):
Awesome work! Thank you.

I had one question, otherwise it looks good.

charts/tempo-distributed/templates/_helpers.tpl
jhlubek: I actually started writing it for global/component only, but then I noticed you have Tempo images grouped in `tempo` key, e.g. here https://github.com/grafana/helm-charts/blob/c4fdafbeae595f51ff9f33dc3ebb8a34ec6deb39/charts/tempo-distributed/templates/_helpers.tpl#L31

so I wanted to preserve the same logic here, because I thought variable precedence is `component > tempo > global` but now I see it's actually `global > component > tempo` :thinking: and I don't really understand why.

I guess you do, so should I use this instead?
```suggestion
{{- $imagePullSecrets := coalesce .global.pullSecrets .component.pullSecrets -}}
```
and edit it in relevant places?

joe-elliott:
Thanks for clarification on the question I had. It looks like we have a few CI things to clean up and then I'd be glad to merge. Thanks!

jhlubek:
@joe-elliott Thank you. I just rebased, retested, and bumped version, so CI should pass now.

joe-elliott (approved):
Looks good. Thanks for the contribution!

z0rc:
This breaks chart rendering for me. `0.27.15` works okay with my values, while `0.27.16` results in when running `helm diff upgrade`:
```
Error: Failed to render chart: exit status 1: Error: template: tempo-distributed/templates/gateway/deployment-gateway.yaml:47:10: executing "tempo-distributed/templates/gateway/deployment-gateway.yaml" at <include "tempo.gatewayImagePullSecrets" .>: error calling include: template: tempo-distributed/templates/gateway/_helpers-gateway.tpl:22:4: executing "tempo.gatewayImagePullSecrets" at <include "tempo.imagePullSecrets" $dict>: error calling include: template: tempo-distributed/templates/_helpers.tpl:41:64: executing "tempo.imagePullSecrets" at <.component.pullSecrets>: nil pointer evaluating interface {}.pullSecrets
```

z0rc:
See https://github.com/grafana/helm-charts/pull/2115

jhlubek:
> This breaks chart rendering for me. `0.27.15` works okay with my values, while `0.27.16` results in when running `helm diff upgrade`:
> 
> ```
> Error: Failed to render chart: exit status 1:
> Error: 
>   template: tempo-distributed/templates/gateway/deployment-gateway.yaml:47:10: executing "tempo-distributed/templates/gateway/deployment-gateway.yaml" at <include "tempo.gatewayImagePullSecrets" .>: error calling include: template: tempo-distributed/templates/gateway/_helpers-gateway.tpl:22:4: executing "tempo.gatewayImagePullSecrets" at <include "tempo.imagePullSecrets" $dict>: error calling include: template: tempo-distributed/templates/_helpers.tpl:41:64: executing "tempo.imagePullSecrets" at <.component.pullSecrets>: nil pointer evaluating interface {}.pullSecrets
> ```

I'm really sorry :disappointed: I must have missed this one during rebase and tested without gateway. Thanks for the quick fix!

PR #2084, opened 2022-12-29
[tempo-distributed] add hpa for tempo querier

jdgeisler:
Autoscaling support in the tempo-distributed helm chart was added in the following PR https://github.com/grafana/helm-charts/pull/1057, however it did not include autoscaling for the querier.

The tempo querier should be auto-scalable (https://github.com/grafana/helm-charts/pull/1057#issuecomment-1048858650), so this PR is adding the HPA for the querier, similar to the other tempo components.


joe-elliott (approved):
Thanks for the contribution! Not sure how that was missed :)

Issue #2089, opened 2023-01-03
[tempo-distributed] Add option to override nginx resolver


r3nic1e:
Hi,

I have a Kubernetes cluster with external DNS (CloudDNS for GKE). And tempo-distributed chart forces me to override the whole nginx config just to change `resolver` line.
Maybe loki-distributed approach will be useful here - https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/values.yaml#L1018

```
        {{- if .Values.gateway.nginxConfig.resolver }}
        resolver {{ .Values.gateway.nginxConfig.resolver }};
        {{- else }}
        resolver {{ .Values.global.dnsService }}.{{ .Values.global.dnsNamespace }}.svc.{{ .Values.global.clusterDomain }};
        {{- end }}
```

PR #2090, opened 2023-01-03
[tempo-distributed] Allow to override nginx resolver

r3nic1e:
Implements #2089


mapno (approved):
Nice addition, thanks!

PR #2106, opened 2023-01-11
[tempo] Introduce appProtocol, HTTPS for ServiceMonitor, and RemoteWrite headers

verejoel:
This PR introduces changes that enable proper integration of the ServiceMonitor with the Istio service mesh.

* Introduce the `appProtocol` field on the Tempo metrics port. This should be set to `http` to allow Istio to automatically select the protocol for traffic.
* Introduce `scheme` and `tlsConfig` configuration options on the ServiceMonitor. These need to be set to HTTPS and provided with a valid certificate when Istio mTLS is in use.

Additionally, the default `tempo.config` value has been updated:
* The indentation has been fixed.
* Overrides block is now only shown if needed
* Remote Write HTTP headers may be set (this was required for our use-case where we remote write to Thanos and use an HTTP header to identify the tenancy).

This PR should be fully backwards compatible if none of the new values are enabled (beyond formatting of the configmap).

### DIFF
--- a/charts/tempo/templates/service.yaml
+++ b/charts/tempo/templates/service.yaml
@@ -35,6 +35,9 @@ spec:
   - name: tempo-prom-metrics
     port: 3100
     targetPort: 3100
+    {{- with .Values.serviceMonitor.appProtocol }}
+    appProtocol: {{ . }}
+    {{- end }}
   {{- if .Values.tempoQuery.enabled }}
   - name: jaeger-metrics
     port: 16687
--- a/charts/tempo/templates/service.yaml
+++ b/charts/tempo/templates/service.yaml
@@ -42,6 +42,9 @@ spec:
   - name: jaeger-metrics
     port: 16687
     targetPort: 16687
+    {{- with .Values.serviceMonitor.appProtocol }}
+    appProtocol: {{ . }}
+    {{- end }}
   - name: tempo-query-jaeger-ui
     port: 16686
     targetPort: 16686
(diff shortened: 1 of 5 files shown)


charts/tempo/templates/service.yaml
```diff
@@ -35,10 +35,16 @@ spec:
   - name: tempo-prom-metrics
     port: 3100
     targetPort: 3100
+    {{- with .Values.serviceMonitor.appProtocol }}
+    appProtocol: {{ . }}
```
joe-elliott, line 39: afaik there's no way in Tempo to configure this to be anything other than http. does it make sense to just hardcode these to "http"?

this may be a mild breaking change. does it change a service or port name in k8s?

charts/tempo/templates/servicemonitor.yaml
```diff
@@ -1,40 +1,62 @@
-{{- if .Values.serviceMonitor.enabled }}
+{{- with .Values.serviceMonitor }}
+{{- if .enabled }}
```
joe-elliott, line 2: thanks for updating this file. much cleaner now +1

charts/tempo/values.yaml
```diff
@@ -118,39 +120,57 @@ tempo:
...
+  {{- if or .Values.tempo.global_overrides .Values.tempo.metricsGenerator.enabled }}
+  overrides:
+    {{- with .Values.tempo.global_overrides }}
+    {{- toYaml . | nindent 2 }}
+    {{- end }}
+    {{- if .Values.tempo.metricsGenerator.enabled }}
+    metrics_generator_processors:
+      - 'service-graphs'
+      - 'span-metrics'
+  metrics_generator:
```
joe-elliott, line 163: it's weird that this is inside of the is statement on line 154. maybe put this in it's own "if metrics generator enabled" condition?

joe-elliott (commented):
nice cleanup and feature additions. added a few small thoughts/questions. let me know what you think!

PR #2107, opened 2023-01-11
Disable helm chart signing and provenance

slim-bean:
There are a few reasons for this:

* The GPG key being used was potentially compromised related to a recent security incident in another CI tool which had this GPG private key as a secret
* We consider these "public" helm charts, we don't want the existence of a signature to indicate a level of scrutiny or security that doesn't exist.
* The public key was never published to a key server as far as we can tell and as such we doubt anyone is actually using this as it it's extremely difficult right now to find the public key
* You should always verify changes made when updating a chart as both good operational and security practice.

In short: at best this signing provides little value, at worst a false sense of security.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -24,18 +24,6 @@ jobs:
         with:
           version: v3.5.2
 
-      - name: Prepare GPG key
-        run: |
-          gpg_dir=.cr-gpg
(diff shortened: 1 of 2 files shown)


simonswine (approved):
LGTM

jeschkies (approved):

PR #2108, opened 2023-01-11
Remove gpg-key property from update-helm-repo flow

trevorwhitney:
key was remove from the job, need to remove it from the inputs too.


trevorwhitney:
closed in favor of #2109

PR #2110, opened 2023-01-11
[loki-distributed] Update query-fronted service to set clusterIP

philipschwartz-okta:
Documentation suggests using clusterIP=None to fix issues with startup race conditions between query-frontend and queriers.

This change sets this value by default.

Signed-off-by: Philip Schwartz <philip.schwartz@okta.com>


zanhsieh:
@philipschwartz-okta 
Thanks for the contribution. Could you bump the version in Chart.yaml please?

zanhsieh (changes requested):
@philipschwartz-okta
Thanks for the contribution. Could you bump the version in Chart.yaml please?

zanhsieh (approved):

Siebjee:
Can we revert this change? Helm now fails for existing chart deployments:
https://github.com/grafana/helm-charts/issues/2127

PR #2116, opened 2023-01-12
[grafana] Explicit volumeClaimTemplate kind/apiVersion

sathieu:
Otherwise, I'm having a diff in ArgoCD:

```diff
     type: RollingUpdate
   volumeClaimTemplates:
-    - apiVersion: v1
-      kind: PersistentVolumeClaim
-      metadata:
+    - metadata:
        creationTimestamp: null
        name: storage
```

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.50.0
+version: 6.50.1
 appVersion: 9.3.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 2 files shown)


zanhsieh:
@sathieu 
Could you provide more strong support that `VolumeClaimTemplates` could take explicit PVC syntax? I only could find this the official k8s documentation:

https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#components

It does not support explicit PVC syntax.

sathieu:
@zanhsieh You are right. This is probably a bug in ArgoCD. Will try again with the latest ArgoCD version.

sathieu:
This is https://github.com/argoproj/argo-cd/issues/11143.

Closing.

PR #2119, opened 2023-01-13
[grafana] fix: also include extra ports for headless svc

clementnuss:
particularly useful when one want to set up unified alerting in HA mode, as the various grafana instances then use the gossip protocol to exchange alerting status/silences.
as this protocol runs on port 9094, we need a way to enable this port for the headless service

https://grafana.com/docs/grafana/latest/alerting/set-up/configure-high-availability/

### DIFF
--- a/charts/grafana/templates/headless-service.yaml
+++ b/charts/grafana/templates/headless-service.yaml
@@ -1,5 +1,6 @@
 {{- $sts := list "sts" "StatefulSet" "statefulset" -}}
 {{- if or .Values.headlessService (and .Values.persistence.enabled (not .Values.persistence.existingClaim) (has .Values.persistence.type $sts)) }}
+{{- $root := . }}
 apiVersion: v1
 kind: Service
 metadata:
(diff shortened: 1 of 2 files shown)


charts/grafana/templates/headless-service.yaml
```diff
@@ -20,4 +21,7 @@ spec:
   - protocol: TCP
     port: 3000
     targetPort: {{ .Values.service.targetPort }}
+    {{- with .Values.extraExposePorts }}
+    {{- tpl (toYaml . | nindent 2) $root }}
+    {{- end }}
```
zanhsieh, line 26: I don't think this will work. See:

https://github.com/clementnuss/helm-charts/blob/c79a5e44f21cc90832911c0ad7bf8105ed7b3623/charts/grafana/values.yaml#L198

Do you mean:
```yaml
  - protocol: TCP
    port: 3000
    targetPort: {{ .Values.service.targetPort }}
  {{- with .Values.extraExposePorts }}
  {{- tpl (toYaml . | nindent 2) $root }}
  {{- end }}
```

zanhsieh (changes requested):

clementnuss:
closing, as it was solved in another way in the meantime :/

PR #2121, opened 2023-01-13
[loki-distributed] Fix compactor address without protocol schema

zaldnoay:
The previous version of loki [included](https://github.com/grafana/helm-charts/blob/5ec00d6923c3045e3c387f8367ed90193998a70c/charts/loki-distributed/values.yaml#L86) a wrong compactor address, which caused the following errors from other components:
```
level=error ts=2023-01-13T09:47:29.06565832Z caller=http.go:131 msg="error getting cache gen numbers from the store" err="Get \"loki-distributed-qa-compactor:3100\": unsupported protocol scheme \"loki-distributed-qa-compactor\""
```

This changes fix compactor address by adding `http://` protocol schema.

Related comment: https://github.com/grafana/loki/pull/3109#discussion_r551354841

Signed-off-by: junwei.liang <junwei.liang@mintegral.com>

PR #2124, opened 2023-01-13
Add protocol to Compactor URL |  #2123

### LINKED ISSUE #2123: Common Compactor URL is set without protocol
Hi,

I have just upgraded `loki-distributed` helm chart to `0.69.0`.
Right after the deployment settled, ruler hit with the following error:
```logfmt
level=error ts=2023-01-13T16:39:07.632129253Z caller=http.go:97 msg="error getting delete requests from the store" err="Get \"loki-loki-distributed-compactor:3100\": unsupported protocol scheme \"loki-loki-distributed-compactor\""
```
It is right, because the protocol is not set in the `0.69.0` [values](https://github.com/grafana/helm-charts/blob/loki-distributed-0.69.0/charts/loki-distributed/values.yaml#L86). After changing it to 
```yaml
compactor_address: http://{{ include "loki.compactorFullname" . }}:3100
```
ruler works as expected. I will prepare a PR with a fix.

Kind regards,
Szymon

sgrzemski:
Add protocol to Compactor URL.

Fixes #2123

### DIFF
--- a/charts/loki-distributed/values.yaml
+++ b/charts/loki-distributed/values.yaml
@@ -83,7 +83,7 @@ loki:
       http_listen_port: 3100
 
     common:
-      compactor_address: {{ include "loki.compactorFullname" . }}:3100
+      compactor_address: http://{{ include "loki.compactorFullname" . }}:3100
 
     distributor:
       ring:


sgrzemski:
Duplicates #2121

PR #2130, opened 2023-01-16
[grafana] Add targetLabels to ServiceMonitors

jkroepke:


charts/grafana/values.yaml
```diff
@@ -194,6 +194,7 @@ serviceMonitor:
   tlsConfig: {}
   scrapeTimeout: 30s
   relabelings: []
+  targetLabels: []
```
zanhsieh, line 197: This should be empty map `{}`:
```suggestion
  targetLabels: {}
```

zanhsieh (changes requested):

charts/grafana/values.yaml
jkroepke: Why? `targetLabels` is a list.

zanhsieh: Then, did you provide an example in values.yaml? How does an engineer / developer know?

zanhsieh (changes requested):

charts/grafana/values.yaml
jkroepke: In the same way, how he knows set `relabelings`. I can provide list of string as example. Does it help?

```
% k explain servicemonitor.spec.targetLabels
KIND:     ServiceMonitor
VERSION:  monitoring.coreos.com/v1

FIELD:    targetLabels <[]string>

DESCRIPTION:
     TargetLabels transfers labels from the Kubernetes `Service` onto the
     created metrics.
```

zanhsieh: Yes.

charts/grafana/values.yaml
```diff
@@ -1068,6 +1069,7 @@ imageRenderer:
     tlsConfig: {}
     scrapeTimeout: 30s
     relabelings: []
+    targetLabels: []
```
zanhsieh, line 1072:
```suggestion
    # See: https://doc.crds.dev/github.com/prometheus-operator/kube-prometheus/monitoring.coreos.com/ServiceMonitor/v1@v0.11.0#spec-targetLabels
    targetLabels: []
      # - targetLabel1
      # - targetLabel2
```

zanhsieh (changes requested):

zanhsieh (approved):

charts/grafana/templates/image-renderer-servicemonitor.yaml
```diff
@@ -41,4 +41,8 @@ spec:
   namespaceSelector:
     matchNames:
       - {{ include "grafana.namespace" . }}
+  {{- with .Values.serviceMonitor.targetLabels }}
```
peikk0, line 44: Should have been:
```
{{- with .Values.imageRenderer.serviceMonitor.targetLabels }}
```

peikk0: Fix in https://github.com/grafana/helm-charts/pull/2137

PR #2137, opened 2023-01-20
[grafana] Fix grafana-image-renderer network policies and service monitor

peikk0:
* Fix the grafana-image-renderer network policies
  * The namespace and pods selectors need to be in the same rule to really match only the Grafana pods and nothing else from the same namespace
  * The egress policy needs to target the target port of the Grafana service
* Use the proper target labels value for the `grafana-image-renderer` ServiceMonitor (see https://github.com/grafana/helm-charts/pull/2130#pullrequestreview-1263011569)

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.50.1
+version: 6.50.2
 appVersion: 9.3.1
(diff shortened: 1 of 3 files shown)


peikk0:
The network policies issues were found while deploying to our staging environment today, and the fix was tested on that same environment successfully.

zanhsieh (approved):

Issue #1944, opened 2022-11-02
[tempo-distributed] upgrade kuberentes api and remove API checks against `apiVersion`.


zalegrala:
Currently, the chart checks the API version available in a few places to determine which objects to create.  We may be able to clean up many of these checks if we bump the required version of the kubernetes API.

I'm seeing this in a few places.
* tempo.ingress.apiVersion
* tempo.pdb.apiVersion
* tempo.podDisruptionBudget.apiVersion

These seem like good places to start for some cleanup.

PR #2157, opened 2023-01-27
[tempo-distributed] use appropriate HorizontalPodAutoscaling apiVersion

AlphaStyle:
> autoscaling/v2beta1 HorizontalPodAutoscaler is deprecated in v1.22+, unavailable in v1.25+; use autoscaling/v2 HorizontalPodAutoscaler

Taken [inspiration](https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/templates/_helpers.tpl#L144-L153) from [loki-distributed Helm chart](https://github.com/grafana/helm-charts/blob/main/charts/loki-distributed/).

This Pull Request will make the Horizontal Pod Autoscaling resources toggle between `autoscaling/v2beta1` and `autoscaling/v2` depending on whether the the cluster can support the `apiVersion` or not.

I noticed the issue https://github.com/grafana/helm-charts/issues/1944 to avoid `apiVersion` toggling, but perhaps this can be merged and then adjusted later when / if a cleanup is happening?

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.2.0
+version: 1.2.1
 appVersion: 2.0.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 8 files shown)


z0rc:
Any chance to get this reviewed and merged soon? Tempo is blocking me with EKS 1.25 upgrade due to usage of outdated autoscaling/v2beta1, which is removed with new kubernetes version https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v125.

zalegrala:
This looks reasonable to me.  Does it work?  I don't have a good way to test.  Also I was wondering about the use of the `print` function.  Does that make it easier to read given the conditions that wrap?

z0rc:
I did similar change for loki-distoributed at https://github.com/grafana/helm-charts/pull/2049. This should work as expected.

zalegrala:
Thanks for the confirmation @z0rc.  If we can get this rebased to fixup the Chart.yaml and the README, I'm good with this change.

AlphaStyle:
Did a rebase and bumped the Chart version.

PR #2158, opened 2023-01-27
[fluent-bit] Allow specify custom parsers to input tail

moha-saaid:
To Allow users to configure a default parser in the configuration rather than the docker parser specified by default within the configuration. we have added an attribute parser to be configured from the values.yaml file.

this will help users add additional parsers and also configure them to be used by default.


zanhsieh (changes requested):
@moha-saaid 
Please bump the version in Chart.yaml otherwise the CI won't let this PR pass. Also please sign DCO.

PR #2165, opened 2023-01-30
[fluent-bit] Add support for securityContext

Garnaalkroket:
In order to comply with certain audit policies I'd like to set a securityContext on the  fluent-bit pod.
It seems in the past a podSecurityPolicy was supplied, but that mechanism is deprecated now.
Would it be possible to add this support?

Kind regards,
Robin

### DIFF
--- a/charts/fluent-bit/templates/daemonset.yaml
+++ b/charts/fluent-bit/templates/daemonset.yaml
@@ -34,6 +34,10 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
+      {{- if .Values.podSecurityContext }}
+      securityContext:
+      {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- end}}
       serviceAccountName: {{ template "fluent-bit-loki.serviceAccountName" . }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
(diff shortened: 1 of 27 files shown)


charts/fluent-bit/templates/daemonset.yaml
```diff
@@ -34,6 +34,10 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
+      {{- if .Values.podSecurityContext }}
+      securityContext:
+      {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      {{- end}}
```
zanhsieh, line 40:
```suggestion
      {{- with .Values.podSecurityContext }}
      securityContext:
        {{- toYaml .| nindent 8 }}
      {{- end}}
```

charts/fluent-bit/templates/daemonset.yaml
```diff
@@ -48,6 +52,10 @@ spec:
         - name: fluent-bit-loki
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.securityContext }}
+          securityContext:
+          {{- toYaml .Values.securityContext | nindent 12 }}
+          {{- end }}
```
zanhsieh, line 58:
```suggestion
          {{- with .Values.securityContext }}
          securityContext:
            {{- toYaml . | nindent 12 }}
          {{- end }}
```

zanhsieh (changes requested):
Please bump the version in Chart.yaml. Thank you.

charts/fluent-bit/templates/daemonset.yaml
Garnaalkroket: Thanks! Should be fixed.

zanhsieh (changes requested):
Please sign DCO. Thanks. For instruction please see the bottom of this PR page `Details` next to `DCO Action required after 1s — DCO`.

PR #2167, opened 2023-01-30
[grafana] Alerting: expose UDP port

JohnnyQQQQ:
This change will expose the UDP port needed by the gossip protocol. It also changes the name of the gossip port, as the old one was too long (15 chars limit on port names in Kubernetes).


alexmobo (approved):
LGTM

PR #2171, opened 2023-01-31
[tempo-distributed] Release 2.0

joe-elliott:
Creates the 2.0 release.

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -473,9 +473,9 @@ compactor:
       # -- Blocks in this time window will be compacted together
       compaction_window: 1h
       # -- Amount of data to buffer from input blocks
-      chunk_size_bytes: 5242880
+      v2_in_buffer_bytes: 5242880
       # -- Flush data to backend when buffer is this large
-      flush_size_bytes: 20971520
+      v2_out_buffer_bytes: 20971520
       # -- Maximum number of traces in a compacted block. WARNING: Deprecated. Use max_block_bytes instead.
       max_compaction_objects: 6000000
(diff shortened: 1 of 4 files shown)


charts/tempo-distributed/values.yaml
```diff
@@ -473,17 +473,17 @@ compactor:
       # -- Blocks in this time window will be compacted together
       compaction_window: 1h
       # -- Amount of data to buffer from input blocks
-      chunk_size_bytes: 5242880
+      v2_in_buffer_bytes: 5242880
       # -- Flush data to backend when buffer is this large
-      flush_size_bytes: 20971520
+      v2_out_buffer_bytes: 20971520
```
modulitos, line 478: nit: Looks like the default value of this `v2_out_buffer_bytes` field is now 30 MB, as documented here: https://grafana.com/docs/tempo/latest/configuration/#compactor

So I think we'll want to make the following change:
```suggestion
      v2_out_buffer_bytes: 31457280
```

where `31457280 = (2**20) * 30`

faustodavid (approved):
Thank you! 🎉

charts/tempo-distributed/values.yaml
joe-elliott: thank you for pointing this out! it turns out the doc is wrong. The default is being set here:

https://github.com/grafana/tempo/blob/main/modules/compactor/config.go#L26

and the constant is defined here:

https://github.com/grafana/tempo/blob/main/tempodb/compactor.go#L28

Thanks for catching this! I'm creating an issue in the Tempo repo.

modulitos: Ah I see. Thanks for pointing out the source and following up!

PR #2172, opened 2023-01-31
[promtail] add ingress

rasta-rocket:
Hey folks 👋 ,

Little PR here to add that capability to get an ingress object linked to the deployment of Promtail

It is especially useful when we deploy a promtail agent in "Push API" mode as a kubernetes deployment object

Don't hesitate to ask me if you have any questions 

Cheers 😉

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.7.3
-version: 6.9.0
+version: 6.9.1
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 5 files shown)


charts/promtail/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.7.2
-version: 6.8.2
+version: 6.8.3
```
zanhsieh, line 6: New feature should bump to minor version.
```suggestion
version: 6.9.0
```

zanhsieh (changes requested):

charts/promtail/Chart.yaml
rasta-rocket: Should be good now @zanhsieh

zanhsieh (approved):

PR #2178, opened 2023-02-02
[fluent-bit] Add securityContext support

Garnaalkroket:
In order to comply with certain audit policies I'd like to set a securityContext on the fluent-bit pod.
It seems in the past a podSecurityPolicy was supplied, but that mechanism is deprecated now.
Would it be possible to add this support?
Sorry for opening the PR twice,  I'm new to contributing and I got lost in the whole signing procedure.

Kind regards,
Robin

### DIFF
--- a/charts/fluent-bit/Chart.yaml
+++ b/charts/fluent-bit/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: fluent-bit
-version: 2.3.2
+version: 2.3.3
 appVersion: v2.1.0
 kubeVersion: "^1.10.0-0"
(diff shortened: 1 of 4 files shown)


charts/fluent-bit/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: fluent-bit
-version: 2.3.2
+version: 2.3.3
```
zanhsieh, line 3: New feature should bump to minor version.
```suggestion
version: 2.4.0
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #2190, opened 2023-02-08
[agent-operator] Update operator to version 0.31.2

ptodev:
A new version of the operator Helm charts following the release of Agent Operator [0.31.2](https://github.com/grafana/agent/releases/tag/v0.31.2)


captncraig (approved):

ptodev:
@captncraig @marctc @jcreixell there is no version of the operator helm charts which uses 0.31.1. I guess we can safely assume that it won't be needed? I am asking because if we decide to make one, it'd have to be a larger chart version than this one and it'd look weird. Maybe in the future we could make the helm charts versioning more in line with the operator versioning.

PR #2195, opened 2023-02-10
[tempo-distributed] feat: add configs for querier.search.* to values.yaml

modulitos:
This PR enables a user to configure the options under `query.search.*`: https://grafana.com/docs/tempo/latest/configuration/#querier

I came across this need while trying to set up a serverless backend for the tempo-querier deployment, and described in this guide: https://grafana.com/docs/tempo/latest/operations/serverless_gcp/

Checklist

 - [x] DCO signed
 - [x] Chart Version bumped
 - [x] Title of the PR starts with chart name (e.g. [tempo-distributed])

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.0.1
+version: 1.1.0
 appVersion: 2.0.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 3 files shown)


charts/tempo-distributed/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.0.1
+version: 1.0.2
```
zanhsieh, line 5: feature should bump to minor version.
```suggestion
version: 1.1.0
```

zanhsieh (changes requested):

charts/tempo-distributed/Chart.yaml
modulitos: Thanks for the catch! Change has been made :+1:

zanhsieh:
@modulitos 
Does the user have to know anything for upgrading to minor version? If yes, then does that need to be addressed in README.md?

modulitos:
@zanhsieh I don't think the user has to know anything for this upgrade. The `querier.config.search.*` values are being set to their defaults, so if the user doesn't set them, they'll be set to the same values that Tempo sets them to.

Open to feedback of course, in case I may have missed something.

zanhsieh (approved):

PR #2204, opened 2023-02-14
[tempo-distributed] call toYaml on list data type

modulitos:
I added a config option for `querier.config.search.external_endpoints` in a previous PR: https://github.com/grafana/helm-charts/pull/2195 but I forgot to call `toYaml` on it. I believe it needs the `toYaml` call because it is a `list` type.

While this doesn't cause an issue for regular strings:
```
querier:
  config:
    search:
      external_endpoints:
        - https://my-serverless-endpoint-asdf-uw.a.run.app
```

when we pass in an env var:

```
querier:
  config:
    search:
      external_endpoints:
        - ${MY_SERVERLESS_ENDPOINT}
```
we get the following error:

```
Error: 'error converting YAML to JSON: yaml: line 87: did not find expected '','' or '']'''
```

This PR fixes the bug.
I tested locally and can confirm the fix using both a single item and multi-item list.

Checklist:

 - [x] DCO signed
 - [x] Chart Version bumped
 - [x] Title of the PR starts with chart name (e.g. [tempo-distributed])

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.1.0
+version: 1.1.1
 appVersion: 2.0.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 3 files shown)


joe-elliott (approved):
Thanks for the quick fix!

PR #2207, opened 2023-02-14
[tempo-distributed] fix: correct lookup of policy apiVersion for pbd

DpoBoceka:
`v1.25.5-gke.2000` supports only `policy/v1`, but the chart defaults to `policy/v1beta1`
Probably `kind` shouldn't append `apiVersion`

ArgoCD:
image

### DIFF
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -152,7 +152,7 @@ Return if ingress supports pathType.
 Return the appropriate apiVersion for PodDisruptionBudget.
 */}}
 {{- define "tempo.pdb.apiVersion" -}}
-  {{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" -}}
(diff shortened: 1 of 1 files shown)


zanhsieh:
@DpoBoceka 
I really don't support this unless this PR provides backward compatible solution that works for existing vanilla and other vendor specific k8s.

DpoBoceka:
what would be better? In mimir they [compare](https://github.com/grafana/mimir/blob/main/operations/helm/charts/mimir-distributed/templates/_helpers.tpl#L369-L376) it with k8s version, but that seems to be even worse cause some distributions may deprecate the API sooner 🤔 
I\m actually quite surprise my cluster got `v1beta1` resolved instead of `v1`

carlosjgp:
Do you have any updates on this?
The change LGTM.

This is the most common way of doing the check, not sure if it breaks the previous releases since the object is different and would delete the `1beta1` and add `v1` PSP

@DpoBoceka looks like you need to bump up the Chart version to get the linter happy

Issue #2224, opened 2023-02-24
[charts/grafana] dashboardsConfigMaps optional: true on default


rufreakde:
# Problem:
Allow to deploy configmaps for dashboards in a seperate deployment by default.

```
        dashboardsConfigMaps:
          # KEY should match dashboardProviders.dashboardproviders.yaml.providers.name
          solutionOperatorCaching: "solution-operator-caching"
```

Helm chart is consuming the configmap as `optional : false` ( k8s default )

https://github.com/grafana/helm-charts/blob/541d97051de87a309362e02d08741ffc868cfcd6/charts/grafana/templates/_pod.tpl#L1034

Set the default for the volume as "optional : true" to allow empty directories that can be populated later on by another deployment.

PR #2225, opened 2023-02-24
dashboardsConfigMaps  Optional - Update _pod.tpl

rufreakde:
Making Configmaps from .Values.dashboardsConfigMaps optional

This allows to deploy the configmaps seperately from the original deployment.

https://github.com/grafana/helm-charts/issues/2224


rufreakde:
Signing should be fixed added the correct commit message

rufreakde:
Okay... I will just cancel PR and use Kustomize since I can currently only use UI and editing the commit messages via GitHub UI is just to not possible.

PR #2235, opened 2023-02-28
[tempo-distributed] fix: disable minio by default

joshuasimon-taulia:
after https://github.com/grafana/helm-charts/pull/1759 was merged, the tempo-distributed chart unexpectedly installed minio in our k8s cluster which resulted in a bunch of new pods crash looping and setting off prom alerts. as a side note, something about the way the minio chart generates resources causes a bunch of merge conflicts in our gitops CI commitback logic. because of this, the helm principal that charts should always be installable using only default values, and the below linked commit, it seems best to disable minio by default
ref: https://github.com/grafana/helm-charts/pull/1759#issuecomment-1448460904

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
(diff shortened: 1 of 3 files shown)


faustodavid (approved):
LGTM, thanks for your contribution.

PR #2237, opened 2023-03-01
[grafana] Allow setting persistentVolumeReclaimPolicy for persistent volume

kosta:
In the default configuration (using deployment instead of statefulset), a pod eviction can delete the persistent volume. Using StatefulSet helps somewhat, but I can sleep better at night if I know that my grafana persistent volume is not easily accidentally deleted. (This keeps the kubernetes default of `Delete` but can be set to e.g. `Retain`)

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.52.1
+version: 6.52.2
 appVersion: 9.4.3
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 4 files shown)


zanhsieh (changes requested):
@kosta 
Thanks for the contribution. Could you bump the version in Chart.yaml and sign DCO please?

kosta:
I signed the CLA and bumped the helm chart version. Let me know if there's anything else I can do to move this forward :)

kosta:
Pushed the same code, but rebased on main with a single commit that has the required DCO.

zanhsieh (approved):

PR #2246, opened 2023-03-02
[promtail] fix ingress deployment

vquie:
I just discovered that we can now deploy an ingress and ran right into this issue that @rasta-rocket must've overseen in c0c9638.

I also created a ci values file that works just fine.

I didn't add it to any actions though.

Update:
This PR edits the ingress section to be configured per `extraPorts`.
Each `extraPort` will have its own ingress with own annotations, ingressclass, you name it.

Checks are working.

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.7.4
-version: 6.9.2
+version: 6.9.3
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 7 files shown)


rasta-rocket (approved):

rasta-rocket:
@zanhsieh do you think that you can review that ?
Sorry about that by the way 🙈

vquie:
I just discovered that this MR does not spawn a service yet. I will add it and push it here.

zanhsieh:
@vquie 
Also please don't forget sign DCO please.

vquie:
Had to no some more changes.

With this PR the ingress will be configured in the `extraPorts` section. You can add an ingress per `extraPort`. You can add multiple hosts per ingress. It will reuse the port configured.

I also added an ingress to the previous ci check and remove my own check for it.

Since the previous release was not working I would't consider this a breaking change. Otherwise I would.

Please check and ask if there are any questions.

vquie:
fixed both issues from the tests, the readme and the error in the ingress.yaml that could occur if `extraPorts` key is upper case.

zanhsieh (approved):

PR #2250, opened 2023-03-03
[grafana] rbac cluster-scoped resource creation conditions

gietschess:
as long as .Values.rbac.extraClusterRoleRules is explicitly defined the cluster-scoped rbac resources (clusterrole and its binding) should be created.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.51.4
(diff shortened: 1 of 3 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.51.3
+version: 6.51.4
```
```suggestion
version: 6.51.5
```

zanhsieh (approved):

PR #2253, opened 2023-03-03
[tempo] Make tmp directory read/write when using metricsGenerator

Kirchen99:
I got an error when I enabled metricsGenerator:
```
/tmp/tempo: read-only file system
```

To fix this, this PR makes /tmp mounted to emptyDir


Kirchen99:
I think it is a bugfix so I just bumped the PATCH. Feel free to tell me if I should bump the MINOR or the MAJOR.

zanhsieh (approved):

PR #2255, opened 2023-03-03
Fix for ingress template variables

awsprof:
Have understood that the ingress api version is called grafana.ingress.isStable within the helpers tpl file - while it is typed promtail.ingress.isStable which is never declared anywhere. Is it right to fix this variable here?


awsprof:
To @unguiculus @Whyeasy  - I am new to this commit process. I had to do a rebase (as suggested by the Git system) for including my signed-off message. Must have made a mistake somewhere in the commit process as I seemed to have pulled the changes by some other users. Closing this pull request and reopening again.

PR #2258, opened 2023-03-04
[loki-distributed] Update schema to use boltdb-shipper and/or v12 schema

LinTechSo:
Hi,
According to the official Loki GitHub repository, the schema configuration updated to 2022-01-11 v12.
appreciate it if you also add it to Loki distributed helm chart.
This PR makes schema configuration be configurable.
thanks.

### DIFF
--- a/charts/loki-distributed/values.yaml
+++ b/charts/loki-distributed/values.yaml
@@ -118,7 +118,18 @@ loki:
     {{- if .Values.loki.schemaConfig}}
     schema_config:
     {{- toYaml .Values.loki.schemaConfig | nindent 2}}
-    {{- end}}
+    {{- else }}
+    schema_config:
+      configs:
+        - from: 2022-01-11
+          store: boltdb-shipper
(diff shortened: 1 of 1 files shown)


charts/loki-distributed/values.yaml
```diff
@@ -118,7 +118,18 @@ loki:
     {{- if .Values.loki.schemaConfig}}
     schema_config:
     {{- toYaml .Values.loki.schemaConfig | nindent 2}}
-    {{- end}}
+    {{- else }}
+    schema_config:
```
zanhsieh, line 122:
```suggestion
    schema_config:
    {{- if .Values.loki.schemaConfig}}
    {{- toYaml .Values.loki.schemaConfig | nindent 2}}
    {{- else }}
```

zanhsieh (changes requested):
@LinTechSo 
Thanks for the contribution. Could you bump the version in Chart.yaml to latest and sign DCO as the bottom of this PR page please?

PR #2269, opened 2023-03-08
[grafana] feat: allow kubeVersion override via values

julien-sugg:
Greetings,

This is a small PR to easily be able to specify the kubeVersion via the values for use-cases like `kustomize`.

> Note that I wasn't able to generate the doc yet as it seems to override the existing one. Shall it be manually updated ?

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.52.1
+version: 6.53.0
 appVersion: 9.4.3
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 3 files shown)


zanhsieh (changes requested):
@julien-sugg 
Thanks for the contribution. However, I don't think grafana would adapt this semVer / kubeVersion approach because:
1. Backward compatibility
2. Tag on certain semVer / kubeVersion does not mean the API would be available all the time across all kind k8s, especially managed k8s. You might search over the issues in grafana to see yourself.  Even worse, the engineers work on grafana only could touch semVer / kubeVersoin; they can't assign kubeVersion to vendor specific version, e.g. *-gke.*, *-eks-*, etc. Handling those would be a big mess.

PR #2273, opened 2023-03-13
[grafana] Revert "Allow setting persistentVolumeReclaimPolicy for persistent volume"

zanhsieh:
Reverts grafana/helm-charts#2237


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.52.2
+version: 6.52.1
```
```suggestion
version: 6.52.3
```

PR #2275, opened 2023-03-15
[grafana] docs: Add some usage instructions about plugins

chxcode:
Hi, I think this PR is very helpful for users with poor network conditions. Thanks


zanhsieh (changes requested):
@chxcode 
Thanks for the contribution. Could you bump the version in Chart.yaml and sign DCO please?

PR #2288, opened 2023-03-23
[fluent-bit] Add automountServiceAccountToken variable

stgrace:
This PR adds the option to disable automounting serviceaccount tokens, both on the serviceaccount level and on the pods. 
This option is important as a best practice, because pods that don't make calls to the Kubernetes API should not have a Kubernetes API token in the first place.

### DIFF
--- a/charts/fluent-bit/Chart.yaml
+++ b/charts/fluent-bit/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: "v1"
 name: fluent-bit
-version: 2.4.0
+version: 2.5.0
 appVersion: v2.1.0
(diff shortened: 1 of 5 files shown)


zanhsieh (changes requested):
@stgrace 
Thanks for the contribution. Couple things:
- Could you describe why we need this PR by updating the PR description?
- Since this PR advance a minor version, could you update README.md that describes what would the impact for the old version user for upgrading? e.g. upgrade procedure, incompatibilities, etc.

stgrace:
Hi @zanhsieh,

I updated the README and the description. If any other info is needed, please let me know.

zanhsieh (approved):

PR #2292, opened 2023-03-24
[loki-distributed] Updated query_range to use addresses instead of host

avantasia1975:
I am getting an error when using memcached_frontend
```
level=warn ts=2023-03-23T15:09:31.785668334Z caller=memcached_client.go:228 msg="error updating memcache servers" err="lookup _memcached._tcp.loki-distributed-memcached-frontend on [169.254.20.10:53:](https://169.254.20.10:53/) no such host"
```
Using service instead works for me

Also fixed wrong naming


avantasia1975:
Ok, i have no idea how to fix this DCO issue.
I added the wanted line to the commit message using an email that exists and was verified. 
Also it is public 
No idea what else i can do

PR #2308, opened 2023-03-30
[tempo-distributed] Add possibility to override DNS config for tempo …

mchahar:
fixes grafana/tempo#1462


zalegrala:
Interesting, I didn't know you could do this on a deployment.  The PR looks good to me once the linting is fixed up.  Thanks for the contribution and the issue reference.

mchahar:
Thank you for the review. 
Regarding linting, helm-docs has different results when run on Windows/Linux (I guess because of CRLF). Everything should be good now!

zalegrala (approved):
Nice work, thanks for the contribution.

PR #2382, opened 2023-05-02
[promtail] fix: extraPort metadata placement

MGlants:
As promtail helm template working with annotation, level, it should be moved to root level of ports in example values.yaml


MGlants:
@unguiculus
@Whyeasy
could you check please

zanhsieh (approved):

PR #2389, opened 2023-05-07
[grafana] ADD: dashboard LoadBalancer file

Lee2532:
When developing or testing locally, every time 
example `kubectl --namespace default port-forward $POD_NAME 3000`
It is inconvenient to enter these commands, so add the corresponding values file under the CI directory to add convenience

### DIFF
--- a/charts/grafana/ci/with-only-local-values.yaml
+++ b/charts/grafana/ci/with-only-local-values.yaml
@@ -1,0 +1,4 @@
+service:
+  enabled: true
+  type: LoadBalancer
+  port: 3000


zanhsieh:
@Lee2532 
Thank you for the contribution, although I don't think I would approve this PR. The maintainer would consider merging the PRs based the best interest  of general public, not individual convenience.

Lee2532:
However, many developers think it will be used a lot to develop local projects for Toy projects and other learning. I think it's also good to add the CI file when you first create it for them

PR #2391, opened 2023-05-08
[grafana] Add envValueFrom to image-renderer

Red-M:
This allows using secrets or configmaps in-cluster to configure options for the image-renderer.
This is useful for setting the renderer token via an in-cluster secret.
From grafana/grafana-image-renderer#420

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.56.1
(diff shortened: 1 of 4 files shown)


zanhsieh (changes requested):
@Red-M 
Thank you for the contribution. Would you mind to bump the version in Chart.yaml to latest so the CI can merge this PR once approved?

Red-M:
Done

zanhsieh (approved):

PR #2399, opened 2023-05-10
[tempo] Correctly sets the Tempo version for 2.1.1

joe-elliott:
The appversion was updated in the chart, but the actual images use these tags for some reason.

### DIFF
--- a/charts/tempo/values.yaml
+++ b/charts/tempo/values.yaml
@@ -12,7 +12,7 @@ annotations: {}
 
 tempo:
   repository: grafana/tempo
-  tag: 2.0.1
+  tag: 2.1.1
   pullPolicy: IfNotPresent
   ## Optionally specify an array of imagePullSecrets.
   ## Secrets must be manually created in the namespace.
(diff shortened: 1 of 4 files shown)


faustodavid (approved):
LGTM, thank you @joe-elliott .

I have created the issue https://github.com/grafana/helm-charts/issues/2401 to avoid this in the future.

faustodavid (changes requested):
Can you also update the chart?

charts/tempo/values.yaml
```diff
@@ -12,7 +12,7 @@ annotations: {}
 
 tempo:
   repository: grafana/tempo
-  tag: 2.0.1
+  tag: 2.1.1
```
faustodavid, line 15: We can set these 2 tags to null and modify the statefulset.yaml to also get the version from the chart AppVersion

```
image: {{ .Values.tempo.repository }}:{{ .Values.tempo.tag | default .Chart.AppVersion | quote }}
```
and
```
image: {{ .Values.tempoQuery.repository }}:{{ .Values.tempoQuery.tag | default .Chart.AppVersion | quote }}
```

joe-elliott: nice. totally agree with this.

faustodavid (approved):

Issue #2404, opened 2023-05-12
[tempo-distributed] query-frontend-discovery is never ready 


tcoville-coveo:
When deploying tempo using the tempo-distributed chart, the querier uses the `query-frontend-discovery` service to lookup the query-frontend pods. 
But it seems that the same `query-frontend` pods wait for `querier` pods to register in order to report a ready status.
Unfortunately, the `query-frontend-discovery` only expose `ready` pods so the querier cannot connect to the frontend and thus the frontend is never ready.  This is a chicken & egg situation.

Adding `publishNotReadyAddresses: true` to the  `query-frontend-discovery` service seems to fix the issue. It also seems to be set in the various other projects that use the same pattern and in the [`jsonnet` version of the deployment:](https://github.com/grafana/tempo/blob/6299d56be6785e306d7c9b077135391211e84cb5/operations/jsonnet/microservices/frontend.libsonnet#L80-L89) 
but not in the helm chart: https://github.com/grafana/helm-charts/blob/23c8f967d1daabb3da022dc0fbafa5cef1084141/charts/tempo-distributed/templates/query-frontend/service-query-frontend-discovery.yaml#L12-L40

PR #2405, opened 2023-05-12
[tempo-distributed] fix missing `publishNotReadyAddresses` in  query-frontend-discovery service

tcoville-coveo:
The `publishNotReadyAddresses` flag seems to be set in the jsonnet version of the deployment but is missing in the helm chart and prevents the `query-frontend-discovery` service to expose its underlying pods. 

fixes: https://github.com/grafana/helm-charts/issues/2404


zalegrala (approved):
Seems reasonable to me.  Thanks for the PR.

PR #2434, opened 2023-05-27
[grafana] Use tpl in envSecrets

jkroepke:
All .Values.env values are already pass through tpl. I would like to have the same behavior for secrets, too.

https://github.com/grafana/helm-charts/blob/eb55960453f49814d3234fd4417ffb2d5b127b80/charts/grafana/templates/_pod.tpl#L950-L953

### DIFF
--- a/charts/grafana/templates/secret-env.yaml
+++ b/charts/grafana/templates/secret-env.yaml
@@ -9,6 +9,6 @@ metadata:
 type: Opaque
 data:
 {{- range $key, $val := .Values.envRenderSecret }}
-  {{ $key }}: {{ $val | b64enc | quote }}
+  {{ $key }}: {{ tpl ($val | toString) $ | b64enc | quote }}
 {{- end }}
 {{- end }}
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.56.6
+version: 6.57.0
 appVersion: 9.5.2
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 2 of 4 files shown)


charts/grafana/templates/secret-env.yaml
```diff
@@ -9,6 +9,6 @@ metadata:
 type: Opaque
 data:
 {{- range $key, $val := .Values.envRenderSecret }}
-  {{ $key }}: {{ $val | b64enc | quote }}
+  {{ $key }}: {{ tpl ($val | toString) $ | b64enc | quote }}
```
zanhsieh, line 12: @jkroepke 
I really puzzle in the use cases. Can you give me an example that value.yaml would refer to some secret somewhere else? And what does this mean / impact to the user? Do they need to escape / check character within their secret somewhere? Isn't it worthy to mention in README.md?

zanhsieh (changes requested):

charts/grafana/templates/secret-env.yaml
jkroepke: > Can you give me an example that value.yaml would refer to some secret somewhere else?

For example: `grafana\.ini.auth\.google.client_secret` or `grafana\.ini.database.password`

The grafana.ini configuration is stored in a ConfigMap and can contains sensitive values. For example the OAUTH credentials. for SSO logins. Instead define the client_secret in a ConfigMap, I would like to define an environment variable `GF_AUTH_GOOGLE_CLIENT_SECRET` which comes from this secret. My current blocker is that the grafana.ini directive support `tpl` while the secret configuration not. `.Values.env` is passed to tpl function, too. It's like everywhere, expect on `envRenderSecret`.

https://github.com/grafana/helm-charts/blob/eb55960453f49814d3234fd4417ffb2d5b127b80/charts/grafana/templates/configmap.yaml#L25

https://github.com/grafana/helm-charts/blob/eb55960453f49814d3234fd4417ffb2d5b127b80/charts/grafana/templates/_pod.tpl#L950-L953

> Do they need to escape / check character within their secret somewhere? Isn't it worthy to mention in README.md?

I added nots on README.md and values.yaml

jkroepke: @zanhsieh does it explain it well?

zanhsieh (approved):

PR #2453, opened 2023-06-07
[promtail] make config positions block more easily configurable

tamcore:
This saves the enduser from having to copy and modify the entire `config.file` block, just to update the `positions` block. As that might be required, as the default of `/run/promtail/positions.yaml` will not work everywhere, as `/run` might be a tmpfs.


zanhsieh (changes requested):
@tamcore 
Thank you for the contribution. Would you mind to sign DCO so the CI can merge after maintainer approval please?

zanhsieh (approved):

PR #2462, opened 2023-06-14
[tempo-distributed] Move overrides configuration to separate runtime configmap

### LINKED ISSUE #2461: Updating runtime overrides in tempo-distributed chart causes cluster rolling restart
I've run into an issue withe the tempo-distributed helm chart where, if I want to update the runtime overrides for a single tenant, it recomputes the checksum for all components and performs a rolling restart on the whole cluster, which isn't desirable.

It looks like it's because its computing the content of the configmap defined here https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/templates/configmap-tempo.yaml

[Discussion](https://grafana.slack.com/archives/C01D981PEE5/p1686621447526099) in Grafana Labs Community #tempo channel

Coming from Mimir they use a different configmap to handle the content for the runtime config which doesn't cause a checksum to be computed and updated on the annotations used in pod specs. I'd like to see this behavior be emulated on the tempo-distributed helm chart as well!

TaylorMutch:
Fixes #2461

* Moves runtime/overrides configuration to separate configmap

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -33,6 +33,9 @@ configStorageType: ConfigMap
 # -- Name of the Secret or ConfigMap that contains the configuration (used for naming even if config is internal).
 externalConfigSecretName: '{{ include "tempo.resourceName" (dict "ctx" . "component" "config") }}'
 
+# -- Name of the Secret or ConfigMap that contains the runtime configuration (used for naming even if config is internal).
+externalRuntimeConfigName: '{{ include "tempo.resourceName" (dict "ctx" . "component" "runtime") }}'
+
 # -- When 'useExternalConfig' is true, then changing 'externalConfigVersion' triggers restart of services - otherwise changes to the configuration cause a restart.
 externalConfigVersion: '0'
 
@@ -1119,7 +1122,7 @@ storage:
 
 # Global overrides
 global_overrides:
-  per_tenant_override_config: /conf/overrides.yaml
+  per_tenant_override_config: /runtime-config/overrides.yaml
   metrics_generator_processors:
     - service-graphs
     - span-metrics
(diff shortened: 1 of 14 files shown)


zalegrala:
This looks reasonable to me.  Were you able to test that it works as expected for all components?

TaylorMutch:
> This looks reasonable to me. Were you able to test that it works as expected for all components?

Testing this week - will get back to you!

TaylorMutch:
> > This looks reasonable to me. Were you able to test that it works as expected for all components?
> 
> Testing this week - will get back to you!

@zalegrala I managed to test this our in my sandbox environment today. However when upgrading I wasn't able to successfully get the `global_overrides` section to update the setting for `per_tenant_override_config` to `/runtime-config/overrides.yaml`. I had to explicitly set it in my values file directly to get it to update.

Looking into some nested logic within the `config` section which seems to depend on values being rendered into the values file also.

TaylorMutch:
I also found that a net new deployment works completely fine. It's just an upgrade that seems to be broken. I am also leveraging this as a subchart so that might also be throwing off the rendering, so will check that.

charts/tempo-distributed/values.yaml
```diff
@@ -1111,7 +1114,7 @@ storage:
 
 # Global overrides
 global_overrides:
-  per_tenant_override_config: /conf/overrides.yaml
+  per_tenant_override_config: /runtime-config/overrides.yaml
```
TaylorMutch, line 1117: This line is what seems to be tripped up on upgrades

TaylorMutch:
> I also found that a net new deployment works completely fine. It's just an upgrade that seems to be broken. I am also leveraging this as a subchart so that might also be throwing off the rendering, so will check that.

Okay false alarm it was indeed an issue with my use as a subchart. I had not packaged it correctly and run `helm dependency update` correctly. Upgrading from 1.4.2 to 1.4.4 does indeed work now on my setup!

@zalegrala when you have time can you review this again?

zalegrala (approved):
Nice, this looks good to me.  Thanks for putting this up.

PR #2465, opened 2023-06-16
[grafana] Add the possibility to configure extra volumeMounts for the alerts sidecar

Paul-Michaud:
Hello,

This PR adds the possibility to add extra volumeMounts to the alerts sidecar. For example you could use it to add scripts in the sidecar that you could run via the `sidecar.alerts.script` variable.


zanhsieh (changes requested):
@Paul-Michaud 
Thanks for the contribution. Since you bump to the minor version, would you mind to mention what the users need to know for upgrading in README.md? E.g. Do they need to scale Deployment/Statefulset to 0 in order to upgrade? What to do if they would like to keep some existing data? etc.

Paul-Michaud:
> @Paul-Michaud Thanks for the contribution. Since you bump to the minor version, would you mind to mention what the users need to know for upgrading in README.md? E.g. Do they need to scale Deployment/Statefulset to 0 in order to upgrade? What to do if they would like to keep some existing data? etc.

Hello, since this is an optional parameter I don't think anything is really needed in order to upgrade, if you don't want to use the parameter nothing will change (It's the same as the `sidecar.dashboards.extraMounts`). Maybe I shouldn't have bumped the minor version ? (Sorry If I misunderstood your question)

zanhsieh:
> Hello, since this is an optional parameter I don't think anything is really needed in order to upgrade, if you don't want to use the parameter nothing will change (It's the same as the `sidecar.dashboards.extraMounts`). Maybe I shouldn't have bumped the minor version ? (Sorry If I misunderstood your question)

Please bump to patch version. Thank you.

zanhsieh (approved):

PR #2443, opened 2023-06-03
Enable OCI chart push

jkroepke:
Fixes #2395

This PR start to publish OCI helm charts to ghcr.io/grafana/charts/, e.g. : `ghcr.io/grafana/charts/grafana`

Bitnami recently switches to OCI as default source for helm charts

* https://blog.bitnami.com/2023/04/httpsblog.bitnami.com202304bitnami-helm-charts-now-oci.html

and prometheus-community enable OCI based helm charts, too:

* https://github.com/prometheus-community/helm-charts/pull/3433

Fetch charts over OCI has multiple benefits:
* It's not longer necessary to download the 1.6M chart [index.yaml](https://grafana.github.io/helm-charts/index.yaml) each time.
* Mirroring to air-gapped environments would much easier, since existing service like Azure Container Registry, Elastic Container Registry and GoHarbor can be used to store/cache the helm charts.

I'm aware that this PR might be a bit aggressive, but I also would like to trigger the discussion here. As I know, this needs to be done at the loki chart, too.

PR #2474, opened 2023-06-22
Revert "Enable OCI chart push"

jkroepke:
Reverts grafana/helm-charts#2443

@zanhsieh Here is the rollback. 

This PR set the error to an warning which give the time to do the correct config while the workflow is not blocked.

In https://github.com/prometheus-community/helm-charts/blob/8b91a619e9407c0e830df7071ee73cd160937b6b/.github/workflows/release.yaml#L52-L60, we are using the same workflow, it seems like that packages are not configured yet to allow an push from actions.

An Grafana org admin has to set this in order to allow helm push from the pipeline: https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-a-registry-using-a-personal-access-token

The documentations also describe that a Personal Access Token should not be used in this case.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,5 +53,7 @@ jobs:
             if [ -z "${pkg:-}" ]; then
               break
             fi
-            helm push "${pkg}" "oci://ghcr.io/${GITHUB_REPOSITORY_OWNER}/charts"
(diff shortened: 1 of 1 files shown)


zanhsieh (approved):

jkroepke:
@zanhsieh are you able to reach an admin team here to enable that given option?

PR #2479, opened 2023-06-27
[grafana] bump default grafana version 10.0.1

zzzinho:
update grafana default version to `v10.0.1`
[grafana v10.0.1 changelog](https://github.com/grafana/grafana/blob/main/CHANGELOG.md#1001-2023-06-22)

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 9.6.0
+version: 6.58.0
 appVersion: 10.0.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.57.5
+version: 7.0.0
 appVersion: 10.0.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 6.57.4
-appVersion: 9.5.5
+version: 6.57.5
+appVersion: 10.0.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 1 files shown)


BobDu:
maybe we also need dump chat major version to: 7.0.0 ?

zzzinho:
right! It should be 7.0.0 :)

brentheimer (approved):

charts/grafana/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 6.57.4
-appVersion: 9.5.5
+version: 7.0.0
+appVersion: 10.0.1
```
zanhsieh, line 4: @zzzinho 
Thanks for the contribution. Can you specify what break change that the helm chart user should know, such as values.yaml change, upgrading / rollback note or strategy in README.md?

zanhsieh (changes requested):

charts/grafana/Chart.yaml
rverenich: @zzzinho , please use 6.x version for chart. There's no breaking changes, no need for "fancy" version number:

> A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an incompatible breaking change needing manual actions.

natl-set:
waiting for this!

charts/grafana/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 6.57.4
-appVersion: 9.5.5
+version: 9.6.0
+appVersion: 10.0.1
```
zanhsieh, line 4: @zzzinho 
Please bump to:

```suggestion
version: 6.58.0
appVersion: 10.0.1
```

@rverenich is not one of the chart maintainers. You should not take his advice.

zanhsieh (changes requested):

rverenich (approved):

dbirks (approved):

charts/grafana/Chart.yaml
zzzinho: I committed suggestion :)

rverenich: Hey, my advice was not to go with 7.0.0 chart version. Just bump 6.x chart version with `appVersion: 10.0.1`
@zzzinho , please approve and let's go further:) Thanks in advance

LarsStegman:
@zanhsieh any chance this could be merged?

zanhsieh (approved):

PR #2484, opened 2023-06-30
[loki-distributed] fix helm warning: cannot overwrite table with non table

JeffreyVdb:
The affinity value is expected to be passed as a string. The default values for querier and compactor were empty maps. This meant that helm would complain with the following message:

```
warning: cannot overwrite table with non table ...
```

which is basically telling the user that the type of the value in the values.yaml file is an empty map, but the value you're overriding it with is a string.

This commit fixes that issue by using the default affinity that is set for all other components (hard node and soft zone anti affinity).


JeffreyVdb:
@zanhsieh Is there anything I can do to push this one forward?

zanhsieh (approved):

Issue #2497, opened 2023-07-11
Default interval of 1m for the serviceMonitor conflicts with the dashboard queries. 


winem:
Here we have a default scrape interval for the Grafana service monitor of 1 minute: 
https://github.com/grafana/helm-charts/blob/c7196457d92a911af8348d7a1c09f7506a36b083/charts/grafana/values.yaml#L205

This setting conflicts with the `rate`/`irate` queries in the dashboard used by the promtheus-operator: https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/grafana-dashboardDefinitions.yaml

We need at least 2 datapoints within the rates timeframe but have just one if the scrape and the rate interval is the same.

I'll provide a PR addressing this issue soon.

PR #2498, opened 2023-07-11
[grafana] set the interval for the serviceMonitor to 30s so that irate()[1m] queries in the operators default Grafana dashboard work

winem:
This PR addresses the issue I reported in #2497 

With this change the Grafana metrics are scraped every 30 seconds so that the dashboards that use irate queries with an interval of `1m` do not show `No data` anymore. 

In my opinion it's more suitable to update the scrape interval to be aligned with the defaults of the other service monitor than to update the queries in the dashboard: https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/grafana-dashboardDefinitions.yaml


zanhsieh (changes requested):
@winem 
Thanks for the contribution. Can you bump the version in Chart.yaml to latest so CI can merge this PR once approved?

zanhsieh (approved):

PR #2509, opened 2023-07-14
[tempo-distributed] Use the recommended minio chart repo

zalegrala:


zalegrala:
I believe since CI passed, the change to `ct.yaml` is acceptable, but perhaps more charts need to be updated for this change as well.

zanhsieh (approved):

PR #2512, opened 2023-07-16
[tempo-distributed] create service-federation-frontend service only if enabled

hd-deman:
Create service-federation-frontend service only if enabled


mapno (commented):
Thanks for fixing this!

LGTM, but you'll need to bump the version.

hd-deman:
@mapno review please

zanhsieh (approved):

PR #2523, opened 2023-07-19
[tempo-distributed] Use default empty values for metrics-generator processors

zalegrala:
Without this change, metrics-generator is disabled by default, but processors are enabled by default.  This misalignment means that users who never enable metrics-generator have an unnecessary amount of log noise on the distributor.  Here we set the default processors to empty.   Users who have not configured the metrics-generator processors and have relied the previous default processors will **need to set the processors you desire**.


zalegrala:
Thoughts on a major version bump for this change?

zanhsieh (approved):

PR #2530, opened 2023-07-24
Correctly template compactor affinity values

### LINKED ISSUE #2529: [loki-distributed] .Values.compactor.affinity isn't passed through `tpl` causing invalid manifest
A default `affinity` block was introduced in https://github.com/grafana/helm-charts/commit/62d86eb2e3e2f3025050538d64085436fbbc1aa3 which according to the associated comment should be passed through `tpl` to be rendered. Unfortunately though, for the compactor deployment this isn't done (see https://github.com/grafana/helm-charts/blob/2548984be77c7d1ca35e970902e52bd699974e1c/charts/loki-distributed/templates/compactor/deployment-compactor.yaml#L107-L110) so it results in an invalid yaml manifest being produced.

In comparison, the `querier` that also had a default affinity added does correctly make use of the `tpl` command - https://github.com/grafana/helm-charts/blob/2548984be77c7d1ca35e970902e52bd699974e1c/charts/loki-distributed/templates/querier/deployment-querier.yaml#L114-L117

This was introduced in loki-distributed 0.70.1

AverageMarcus:
Fixes #2529

Passes the affinity string through `tpl` to ensure it is correctly templated as the default value now contains template strings that cause the manifest rendered to be invalid


JeffreyVdb:
@AverageMarcus I think you'll have to increase the patch version of the Chart, here: https://github.com/grafana/helm-charts/blob/4c754fa633623d8cdfbe576300a6a3af86bfd6da/charts/loki-distributed/Chart.yaml

EStork09:
@AverageMarcus did you mean to close this? Is there another PR that this conflicts with?

AverageMarcus:
Yeah it was fixed here: https://github.com/grafana/helm-charts/pull/2536

EStork09:
Ah cool, thanks for the quick response!

PR #2536, opened 2023-07-25
[loki-distributed] fixed an issue with the compactor affinity

sherifkayad:
The compactor affinity rules in the deployment template didn't use the tpl function which was producing wrong configs at the end


sherifkayad:
@zanhsieh can you please also have a look at this one?

zanhsieh (approved):

PR #2540, opened 2023-07-26
[agent-operator] Add host aliases

YuleZ:

### DIFF
--- a/charts/agent-operator/templates/operator-deployment.yaml
+++ b/charts/agent-operator/templates/operator-deployment.yaml
@@ -62,7 +62,7 @@ spec:
       {{- end }}
       {{- with .Values.hostAliases }}
       hostAliases:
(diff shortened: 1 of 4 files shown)


charts/agent-operator/templates/operator-deployment.yaml
```diff
@@ -60,6 +60,10 @@ spec:
       imagePullSecrets:
       {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.hostAliases }}
+      hostAliases:
+        {{- toYaml . | nindent 8 }}
```
zanhsieh, line 65:
```suggestion
      {{- toYaml . | nindent 8 }}
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #2542, opened 2023-07-26
[enterprise-logs-simple] Add host aliases

YuleZ:


charts/enterprise-logs-simple/values.yaml
```diff
@@ -164,6 +164,11 @@ structuredConfig: {}
 tokengen:
   # -- Whether the job should be part of the deployment
   enable: true
+  # -- hostAliases to add
+  hostAliases: [ ]
```
zanhsieh, line 168:
```suggestion
  hostAliases: []
```

zanhsieh (approved):

PR #2544, opened 2023-07-26
[fluent-bit] Add host aliases

YuleZ:

### DIFF
--- a/charts/fluent-bit/values.yaml
+++ b/charts/fluent-bit/values.yaml
@@ -45,7 +45,7 @@ config:
   #     json_date_format: time
 
 # -- hostAliases to add
-hostAliases: [ ]
+hostAliases: []
 #  - ip: 1.2.3.4
 #    hostnames:
   #      - domain.tld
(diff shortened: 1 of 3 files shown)


charts/fluent-bit/values.yaml
```diff
@@ -44,6 +44,12 @@ config:
   #     Format: json
   #     json_date_format: time
 
+# -- hostAliases to add
+hostAliases: [ ]
```
zanhsieh, line 48:
```suggestion
hostAliases: []
```

zanhsieh (changes requested):

zanhsieh (changes requested):
@YuleZ 
The final step: Sign DCO. Thanks.

YuleZ:
> @YuleZ
> The final step: Sign DCO. See the bottom of this page next to DCO for instructions (click Details). Thanks.


I believe this fails because I used Github's possibility to apply suggestions made by you via the webUI where it cannot be signed correctly. I think I have to put this PR together again.

zanhsieh (approved):

PR #2545, opened 2023-07-26
[loki-canary] add host aliases

YuleZ:

### DIFF
--- a/charts/loki-canary/values.yaml
+++ b/charts/loki-canary/values.yaml
@@ -13,7 +13,7 @@ namespace:
 imagePullSecrets: []
 
 # -- hostAliases to add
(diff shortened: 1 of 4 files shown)


charts/loki-canary/values.yaml
```diff
@@ -12,6 +12,12 @@ namespace:
 # -- Image pull secrets for Docker images
 imagePullSecrets: []
 
+# -- hostAliases to add
+hostAliases: [ ]
```
zanhsieh, line 16:
```suggestion
hostAliases: []
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #2546, opened 2023-07-26
[promtail] Add host aliases

YuleZ:

### DIFF
--- a/charts/promtail/values.yaml
+++ b/charts/promtail/values.yaml
@@ -69,6 +69,12 @@ image:
 # -- Image pull secrets for Docker images
 imagePullSecrets: []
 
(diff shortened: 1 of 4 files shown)


charts/promtail/values.yaml
```diff
@@ -69,6 +69,12 @@ image:
 # -- Image pull secrets for Docker images
 imagePullSecrets: []
 
+# -- hostAliases to add
+hostAliases: [ ]
```
zanhsieh, line 73:
```suggestion
hostAliases: []
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #2547, opened 2023-07-26
[rollout-operator] Add host aliases

YuleZ:

### DIFF
--- a/charts/rollout-operator/values.yaml
+++ b/charts/rollout-operator/values.yaml
@@ -11,7 +11,7 @@ image:
 imagePullSecrets: []
 
 # -- hostAliases to add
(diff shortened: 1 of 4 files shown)


charts/rollout-operator/values.yaml
```diff
@@ -9,6 +9,13 @@ image:
   tag: ""
 
 imagePullSecrets: []
+
+# -- hostAliases to add
+hostAliases: [ ]
```
zanhsieh, line 14:
```suggestion
hostAliases: []
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #2548, opened 2023-07-26
[synthetic-monitoring-agent] Host aliases for synthetic monitoring agent

YuleZ:


charts/synthetic-monitoring-agent/values.yaml
```diff
@@ -30,6 +30,11 @@ image:
 
 # -- Docker registry secret names as an array.
 imagePullSecrets: []
+# -- hostAliases to add
+hostAliases: [ ]
```
zanhsieh, line 34:
```suggestion
hostAliases: []
```

zanhsieh (changes requested):

zanhsieh (approved):

PR #2563, opened 2023-07-31
[tempo-distributed] make block version default to Tempo

joe-elliott:
- Make block version null and rely on Tempo defaults.
- Add a note to not auto scale ingesters unless you're a wizard

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.5.0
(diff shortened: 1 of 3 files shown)


zalegrala (approved):

eviln1:
> Add a note to not auto scale ingesters unless you're a wizard

:wave: Hey where can I read more about this wizard thing - in the sense of what's exactly happening when the ingester scales up/down, how much data we risk losing etc.. ?
I understand it's related to the distribution of trace hashes in the ring, but I couldn't find any specifics.

Thanks in advance !

gigawhat:
@joe-elliott I would also like some more info regarding auto scaling ingester. Where can we find more info about that?

PR #2519, opened 2023-07-18
Add clarifying notes to Loki charts.

JStickler:
I had a user who was confused as to which of the Loki charts they should be using.  Adding a sentence to each Loki Helm Chart to help distinguish which chart is recommended.

PR #2566, opened 2023-08-01
[loki-canary] Update README

JStickler:
I had a user who was confused as to which of the Loki charts they should be using. Adding a sentence to each Loki Helm Chart to help distinguish which chart is recommended.

Breaking up #2519 into separate PRs.

### DIFF
--- a/charts/loki-canary/Chart.yaml
+++ b/charts/loki-canary/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-canary
 description: Helm chart for Grafana Loki Canary
 type: application
 appVersion: 2.8.3
-version: 0.13.0
+version: 0.13.1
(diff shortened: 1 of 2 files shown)


charts/loki-canary/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: loki-canary
 description: Helm chart for Grafana Loki Canary
```
zanhsieh, line 3: @JStickler 

Thanks for the contribution. The helm-doc render would README.md.gotmpl -> read this line (`.chart.description`) -> output to README.md, so:

```suggestion
description: Helm chart for Grafana Loki Canary. This is a community maintained chart."
```

zanhsieh (changes requested):

PR #2568, opened 2023-08-01
[loki-distributed] Update README

JStickler:
I had a user who was confused as to which of the Loki charts they should be using. Adding a sentence to each Loki Helm Chart to help distinguish which chart is recommended.

Breaking up https://github.com/grafana/helm-charts/pull/2519 into separate PRs.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.8.3
-version: 0.71.1
+version: 0.71.2
 home: https://grafana.github.io/helm-charts
 sources:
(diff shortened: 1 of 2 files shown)


MichelHollands (approved):

charts/loki-distributed/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
```
zanhsieh, line 3: @JStickler 

Thanks for the contribution. The helm-doc render would README.md.gotmpl -> read this line (`.chart.description`) -> output to README.md, so:

```suggestion
description: "Helm chart for Grafana Loki in microservices mode. This is a community maintained chart."
```

zanhsieh (changes requested):

Issue #2428, opened 2023-05-24
[promtail] Regex issue with the pipelineStages YAML value


adesprez:
Given the following values for `.Values.config.snippets.pipelineStages`:
```
config:
[...]
    pipelineStages:
    - multiline:
        firstline: '^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}'
        max_wait_time: 3s
```

Expected result for the promtail `Secret`:
```
scrape_configs:
  # See also https://github.com/grafana/loki/blob/master/production/ksonnet/promtail/scrape_config.libsonnet for reference
  - job_name: kubernetes-pods
    pipeline_stages:
      - cri: {}
      - multiline:
          firstline: '^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}'
          max_wait_time: 3s
```

Actual result for the promtail `Secret`:
```
scrape_configs:
  # See also https://github.com/grafana/loki/blob/master/production/ksonnet/promtail/scrape_config.libsonnet for reference
  - job_name: kubernetes-pods
    pipeline_stages:
      - cri: {}
      - multiline:
          firstline: ^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}
          max_wait_time: 3s
```
(no single quotes `'` enclosing the regex).

As a consequence, Promtail is not processing that `firstline` as a regex. Single quotes are mandatory here.

Code responsible of that, `charts/promtail/values.yaml`:
```
    scrapeConfigs: |
      # See also https://github.com/grafana/loki/blob/master/production/ksonnet/promtail/scrape_config.libsonnet for reference
      - job_name: kubernetes-pods
        pipeline_stages:
          {{- toYaml .Values.config.snippets.pipelineStages | nindent 4 }}
```

The `toYaml` function is stripping the single quotes.

Something that could solve that issue, is to use `tpl` function to render `pipelineStages`, as suggested here by tiagodj: https://github.com/grafana/helm-charts/issues/2412

xyfleet:
I had same issue here. Any progress for this one?

adesprez:
@xyfleet , I haven't found a workaround.
I tried multiple syntaxes to escape characters and/or protect the single quotes. No luck. 
I have parked this, since then.

chrizel:
Having the same problem, I can confirm that the `tpl` function can be used to work around this problem by moving the regex to its own custom values variable and then using it in the `pipelineStages` definition:

```yaml
[...]

Issue #2412, opened 2023-05-16
Render pipelineStages with tpl function to support templates in pipelines


tiagodj:
In the values file we have this config:

```yaml
    scrapeConfigs: |
      # See also https://github.com/grafana/loki/blob/master/production/ksonnet/promtail/scrape_config.libsonnet for reference
      - job_name: kubernetes-pods
        pipeline_stages:
          {{- toYaml .Values.config.snippets.pipelineStages | nindent 4 }}
```
https://github.com/grafana/helm-charts/blob/main/charts/promtail/values.yaml#LL425C4-L425C4

Would it be possible to update the code to render the contents of `.Values.config.snippets.pipelineStages` with the `tpl` function? This would allow us to write template code in the pipeline definition.

This is already done in other parts of the definition, for example:

```yaml
 477  {{- tpl (toYaml .Values.config.clients) . | nindent 2 }}
```
https://github.com/grafana/helm-charts/blob/main/charts/promtail/values.yaml#LL477C29-L477C29

PR #2575, opened 2023-08-02
[promtail] Render pipelineStages with tpl function

chrizel:
Implements the suggestion by #2412 which allows us to work around the `toYaml` problems described under #2428

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 2.8.3
-version: 6.13.1
+version: 6.13.2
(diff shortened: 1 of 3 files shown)


zanhsieh (changes requested):
@chrizel 
Thank for the contribution. Can you bump version in Chart.yaml so the CI can merge this PR once approved?

zanhsieh (approved):

conorevans:
As discussed [here](https://github.com/grafana/helm-charts/issues/2428#issuecomment-1667405261) this breaks any Go templating... you should consider reverting this. More importantly, you should also consider your semver - this is not a bugfix, it is a change in behaviour, which warrants a minor version bump. I mean at present it's a breaking change but even if it worked, it would be a minor change in behaviour.

PR #2579, opened 2023-08-07
[tempo-distributed] fix: make auto mount sa token configurable

zhongchen:
The `automountServiceAccountToken` is configurable in the mono-tempo helm chart, but not in the distributed version.

https://github.com/grafana/helm-charts/blob/main/charts/tempo/templates/serviceaccount.yaml#L20

We want to set the field to be true in our tempo deployment, but we can't overwrite it easily right now.


zalegrala (approved):
This looks good.  Thanks for the PR.

PR #2583, opened 2023-08-08
[tempo-distributed] fix: add flush all on shutdown for ingestors

zhongchen:
The option is in the documentation but missing in the helm chart. https://grafana.com/docs/tempo/latest/configuration/#ingester


zalegrala (approved):
Thanks for the PR.

PR #2647, opened 2023-09-11
[grafana] Fix  missing label for dashboard configmap

goodard:
Fix missing sidecar.dashboards.label on custom  dashboard configmap

Signed-off-by: Marko Stojanovic <markostojanovic77@gmail.com>


zanhsieh (changes requested):
@goodard 
Thanks for the contribution. Can you fix the CI error?

zanhsieh (approved):

JefeDavis:
This causes issues if your label value is string "true" or "false" as it converts it from a string back to a bool during templating and then complains that it can't unmarshal a bool into go struct field metadata.labels

needs to be changed to ```{{ $.Values.sidecar.dashboards.labelValue | quote }}```

charts/grafana/templates/dashboards-json-configmap.yaml
```diff
@@ -9,6 +9,9 @@ metadata:
   labels:
     {{- include "grafana.labels" $ | nindent 4 }}
     dashboard-provider: {{ $provider }}
+    {{- if $.Values.sidecar.dashboards.enabled }}
+    {{ $.Values.sidecar.dashboards.label }}: {{ $.Values.sidecar.dashboards.labelValue }}
```
sjoukedv, line 13: @zanhsieh We should set the `labelValue` by default to `1` so that it does not try to create a label with an empty string value (or alternatively use helm `default` function). This broke the chart for me.

sjoukedv: And parse the value to `quote` to ensure it is a string

```suggestion
    {{ $.Values.sidecar.dashboards.label }}: {{ $.Values.sidecar.dashboards.labelValue | default "1" |  quote }}
```

sjoukedv: Now I need to enter

```yaml
        sidecar:
          dashboards:
            enabled: true
            label: grafana_dashboard
            labelValue: "'1'"
```

goodard: @sjoukedv applied your suggestion in https://github.com/grafana/helm-charts/pull/2693

PR #2664, opened 2023-09-21
[tempo-distributed] adjust minReplicas for ingester

dheeg:
If autoscaling is turned for the ingester, the distributor component permanently fails.

```
ingester:
  autoscaling:
    enabled: true
```

This will result in the following Go error in the distributor component:

`level=error ts=2023-09-21T07:26:44.537318884Z caller=rate_limited_logger.go:27 msg="pusher failed to consume trace data" err="at least 2 live replicas required, could only find 1"`

Therefor I propose to change the minReplicas default value for the ingester component:

```
ingester:
  autoscaling:
    minReplicas: 2
```

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.6.4
+version: 1.6.5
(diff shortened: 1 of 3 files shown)


zanhsieh (changes requested):
@dheeg 
Thanks for the contribution. Can you sign DCO, bump the version in Chart.yaml to latest, and update README.md with helm-doc please?

zanhsieh (approved):

PR #2681, opened 2023-09-28
[tempo-distributed] Moves to GET 2.2.1, reverts MinIO dependency.

hedss:
This primarly fixes a breaking change where the MinIO chart dependency had been reset to the prior, deprecated Chart, which broke the installation of this Chart from source.
This occured as part of this PR: https://github.com/grafana/helm-charts/pull/2384 presumably due to a failure to rebase between initial raising and merging.

Have also updated the GET version from 2.1.0 to 2.2.1, inline with latest release.

Fully E2E tested via GET k8s cluster, with Grafana and GrAgent setup.

### DIFF
--- a/charts/tempo-distributed/Chart.lock
+++ b/charts/tempo-distributed/Chart.lock
@@ -1,0 +1,9 @@
+dependencies:
+- name: minio
+  repository: https://charts.min.io/
+  version: 4.0.12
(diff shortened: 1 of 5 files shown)


zalegrala (approved):
I agree, this is the chart we want.

PR #2693, opened 2023-10-04
[grafana] Fix issue with default tag on grafana chart dashboards

goodard:


charts/grafana/templates/dashboards-json-configmap.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     {{- include "grafana.labels" $ | nindent 4 }}
     dashboard-provider: {{ $provider }}
     {{- if $.Values.sidecar.dashboards.enabled }}
-    {{ $.Values.sidecar.dashboards.label }}: {{ $.Values.sidecar.dashboards.labelValue }}
+    {{ $.Values.sidecar.dashboards.label }}: {{ $.Values.sidecar.dashboards.labelValue  | default "1" |  quote }}
```
zanhsieh, line 13: @goodard 
Thanks for the contribution. Can you describe what this scenario apply? Why we should set labelValue to "1" as default value? Why not just eliminate that label if labelValue is empty?

zanhsieh (changes requested):

charts/grafana/templates/dashboards-json-configmap.yaml
cwrau: I assume this PR stems from https://github.com/grafana/helm-charts/pull/2647#discussion_r1346148822 and tries to fix the same problem for which I also just created [a PR](https://github.com/grafana/helm-charts/pull/2696)

sjoukedv: Eliminating the label entirely also solves the problem for me. I have seen some charts that use `1` as the default value for their dashboard configmaps that's why I made this suggestion. It would be great if we can get this merged.

zanhsieh:
Close this PR in favor of #2647.

PR #2695, opened 2023-10-05
[grafana] add global image registry support

QuentinBisson:
This PR allows overriding of the grafana registry at a global level

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.60.2
+version: 6.61.0
 appVersion: 10.1.2
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -59,6 +59,7 @@ This version requires Helm >= 3.1.0.
 | `readinessProbe`                          | Readiness Probe settings                      | `{ "httpGet": { "path": "/api/health", "port": 3000 } }`|
 | `securityContext`                         | Deployment securityContext                    | `{"runAsUser": 472, "runAsGroup": 472, "fsGroup": 472}`  |
 | `priorityClassName`                       | Name of Priority Class to assign pods         | `nil`                                                   |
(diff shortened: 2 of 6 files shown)


zanhsieh (changes requested):
@QuentinBisson 
Sorry but I can't see what is the benefit for specifying registry separated form repository. Can you elaborate more please?

QuentinBisson:
Sure thing @zanhsieh there are a few use cases for this:
- Pull through caching (some users use a pull through registry to act as a cache to store all their images so only the registry changes but not the repository). For instance, I've seen a lot of users deploy harbor for this.
- Security and Compliance (I've seen a lot of users having their own custom images (added tools, CVE fixes and so on) of upstream images and add tools to them).

In general, it's easier for users to be able to set the registry at a global level but for that to work we need to be able to set individual registries

zanhsieh (approved):

pvallone:
@QuentinBisson,  did you consider naming this property `global.imageRegistry` instead of `global.image.registry`? That's the convention followed by the Bitnami Helm charts, the Prometheus charts, and the other charts I've used. It would also be more consistent with the existing `global.imagePullSecrets` property defined by the Grafana chart.

QuentinBisson:
@pvallone I followed what is used in the loki ssd helm chart but I found out today that it's configured differently in tempo-distributed. 
Personally, I think global.image.registry and global.image.pull.secrets and global.image.pull.policy makes more sense though

pvallone:
@QuentinBisson I could see an argument for `global.image.registry` and `global.image.pullSecrets`  if we were starting from scratch. In this case, it feels like the vast majority of Helm charts have standardized on `global.imageRegistry` and `global.imagePullSecrets`. For global parameters like this, I think it's especially important to follow what most other charts do. Myself or someone from my team are going to put up a PR to propose a rename. We'll add you to that PR too.

pvallone:
I'll include a discussion of renaming the others in my PR. Renaming the parameter is technically a breaking change, and I think it'd be less disruptive to scope this to Grafana for now, as the parameter was added very recently.

pvallone:
@QuentinBisson it turns out I can't add you to my PR, but it's here if you're interested: https://github.com/grafana/helm-charts/pull/2735

PR #2700, opened 2023-10-09
[loki-distributed] Query scheduler and index gateway join the memberlist

verejoel:
According to the Grafana Loki [documentation](https://grafana.com/docs/loki/latest/get-started/hash-rings/):
* the query scheduler must be part of a hashring
* the index gateway may optionally be part of a hashring

This PR updates the chart such that the query scheduler and index gateway pods will expose the memberlist container port 7946, and include the member list selector label. 

The index gateway may optionally be prevented from joining the memberlist by setting `.Values.indexGateway.joinMemberlist` to false (default true), reflecting the current state of the documentation.

In addition, the `appProtocol` field may be optionally configured for the ruler, query scheduler, and index gateway.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.9.1
-version: 0.74.6
+version: 0.75.0
 home: https://grafana.github.io/helm-charts
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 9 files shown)


z0rc:
IMO if documentation says "must", it should be enabled in helm chart without option to disable it. Same story for index gateway, if it's beneficial keep it enabled by default with option to disable.

I ain't concerned about backward compatibility. This chart is still version 0.x.x, which allows to change anything without compatibility concerns https://semver.org/#spec-item-4.

verejoel:
@zanhsieh would you be able to take a look?

MasslessParticle (approved):

MasslessParticle:
LGTM, thanks for the contribution!

PR #2702, opened 2023-10-10
[tempo] fix: Conditionally enable jaeger-metrics port on service monitor

maandr:
We recently encountered a situation where the service monitor stopped scraping data. This occurred because the service monitor contained the port configuration `jaeger-metrics` that did not exist in the service unless `.Values.tempoQuery.enabled` was enabled.

Since the presence of the `jaeger-metrics` port in the service is conditional, it should also be subject to the same condition in the service monitor.


mapno (changes requested):
Thanks for the fix! This change will require a bump in the chart's version to merge

PR #2712, opened 2023-10-16
[grafana] Range over extraVolumes 

efbicief:
Previously, this was ranging over extraVolumeClaims instead of extraVolumes, which meant extraVolumes was never read. This caused volumes mounted as extraVolumes to be mounted as emptyDirs.
This also correctly inserts all data from hostPath volumes, and allows configMaps to be used as an extraVolume.

This should fix https://github.com/prometheus-community/helm-charts/issues/3851

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.60.7
+version: 6.60.8
 appVersion: 10.1.5
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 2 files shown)


zanhsieh (changes requested):
@efbicief 
Thanks for the contribution. Can you sign DCO and CLA please?

efbicief:
Apologies for the delay, hopefully that is okay now.

efbicief:
I see we are now on 6.61.1 - should version be bumped to 6.61.2?

zanhsieh (approved):

PR #2716, opened 2023-10-17
[grafana] add externalTrafficPolicy support

schneidermr:
In some cases, the externalTrafficPolicy value should also be configured. The default value of externalTrafficPolicy is 'Cluster' which needs to change for example if you use an internal LoadBalancer type service in Microsoft Azure environment.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 6.61.0
+version: 6.61.1
 appVersion: 10.1.5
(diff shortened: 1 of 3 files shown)


zanhsieh (changes requested):
@schneidermr 
Thanks for the contribution. Can you bump the version in Chart.yaml please?

schneidermr:
> @schneidermr Thanks for the contribution. Can you bump the version in Chart.yaml please?

I bumped the version number of the Grafana chart

zanhsieh (approved):

PR #2722, opened 2023-10-18
[tempo-distributed] Add the ability to add extraObjects

1Psy:
Add the ability to create extraObjects in helm chart.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.6.12
+version: 1.6.13
 appVersion: 2.2.3
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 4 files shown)


zalegrala:
Thank your the PR.  I'm wondering if there is another way you might achieve your goal.  Since you are attempting to render any object here without referencing anything internally, it seems that could live anywhere in your deployment and doesn't necessarily need to live here, is that correct?  Do we gain additional value by having this change as part of the helm chart?

1Psy:
> Thank your the PR. I'm wondering if there is another way you might achieve your goal. Since you are attempting to render any object here without referencing anything internally, it seems that could live anywhere in your deployment and doesn't necessarily need to live here, is that correct? Do we gain additional value by having this change as part of the helm chart?

Similar manifests are used in many diagrams, for example, https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/extra-manifests.yaml.
In my opinion, this is normal practice.
For example, when declaring a VirtualService for Istio or when working with operator secrets.

zalegrala (approved):
I see we do this in the other charts as well.

zanhsieh (approved):

PR #2728, opened 2023-10-20
Adding support for workflow with Github App authentication

yorugac:
Authentication with Github App requires two vars being set in input (APP ID and PEM) and additional step to generate a temporary token. This PR adds support for such authentication in `update-helm-release` workflow.

It is implemented with conditionals in order to remain backwards compatible to setups of other projects that are using just one token at the moment.

It was tested with k6-operator:
- example CI workflow: https://github.com/grafana/k6-operator/blob/main/.github/workflows/helm-release.yaml
- successful run: https://github.com/grafana/k6-operator/actions/runs/6586421178

### DIFF
--- a/.github/workflows/update-helm-repo.yaml
+++ b/.github/workflows/update-helm-repo.yaml
@@ -108,7 +108,7 @@ jobs:
           fetch-depth: 0
           repository: grafana/helm-charts
           path: helm-charts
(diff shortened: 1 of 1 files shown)


krajorama (approved):
LGTM, I checked the successful run and the token is masked in the output as expected

PR #2747, opened 2023-10-28
[loki-distributed] manage compactor as statefulset

rasta-rocket:
Hey there !

Small contribution here to allow users of the chart to manage the compactor as a statefulset

:warning: Also I draw you attention on the fact that in the `values.yaml` I set the compactor field `Kind` as `Statefulset` starting now : this might break existing setup. I did that since Statefulset is kind of the default way to manage stateful software in k8s IMHO, but let me know what you think about it :wink:

If you have any question, do not hesitate to reach me :wink:

Cheers :sunny:

### DIFF
--- a/charts/loki-distributed/templates/compactor/statefulset-compactor.yaml
+++ b/charts/loki-distributed/templates/compactor/statefulset-compactor.yaml
@@ -1,21 +1,30 @@
 {{- if .Values.compactor.enabled }}
-{{- if eq .Values.compactor.kind "Deployment"}}
+{{- if eq .Values.compactor.kind "StatefulSet"}}
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
 metadata:
   name: {{ include "loki.compactorFullname" . }}
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "loki.compactorLabels" . | nindent 4 }}
+    app.kubernetes.io/part-of: memberlist
   {{- with .Values.loki.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.compactor.replicas }}
+  podManagementPolicy: Parallel
+  updateStrategy:
+    rollingUpdate:
+      partition: 0
+  serviceName: {{ include "loki.compactorFullname" . }}-headless
   revisionHistoryLimit: {{ .Values.loki.revisionHistoryLimit }}
-  strategy:
-    type: Recreate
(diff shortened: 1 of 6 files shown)


zanhsieh (approved):

zanhsieh:
@zalegrala
Can you review this please?

charts/loki-distributed/templates/compactor/statefulset-compactor.yaml
```diff
@@ -0,0 +1,190 @@
...
+          secret:
+            secretName: {{ include "loki.fullname" . }}-config
+          {{- else }}
+          configMap:
+            name: {{ include "loki.fullname" . }}
+          {{- end }}
+        - name: runtime-config
+          configMap:
+            name: {{ template "loki.fullname" . }}-runtime
+        {{- if not .Values.ingester.persistence.enabled }}
```
verejoel, line 164: Copy paste error
```suggestion
        {{- if not .Values.compactor.persistence.enabled }}
```

charts/loki-distributed/templates/compactor/statefulset-compactor.yaml
```diff
@@ -0,0 +1,190 @@
...
+        - name: data
+          emptyDir: {}
+        {{- end }}
+        {{- with .Values.compactor.extraVolumes }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+  volumeClaimTemplates:
```
verejoel, line 171: Needs a guard here in case persistence is not enabled or the list of volume claims is empty.

verejoel (changes requested):
This will introduce bugs if the compactor PVC is disabled which needs to be addressed, also please fix the copy-paste error.

zalegrala:
Thanks for the PR.  What's the motivation to manage the compactor as a statefulset?  Is this the recommendation from Loki?

rasta-rocket:
Hello @zalegrala 👋 

I did that for two reasons:
- Statefulset is kind of the default way to manage stateful software in k8s 
- When doing rolling update on deployment + pvc, the rolling is not well managed most of the time, because you need to detach the PV from the deployment N manually before deployment N+1 arrives (because you have concurrent access on the volume) : which result in a helm deployment that fails. That issue never occurs with an STS.

The recommendation from loki is to have that component to be stateful (they do not mention statefulset or deployment + pvc)

Cheers ☀️

zalegrala (approved):

zalegrala:
Ah okay, I didn't realize the compactor was stateful but I suppose this makes sense since if the client doesn't stream.

PR #2748, opened 2023-10-30
[grafana] Add global image registry support for extraInitContainers

pvallone:
If a client sets `global.imageRegistry`, that registry should override whatever registries are specified in `extraInitContainers`.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 7.0.1
+version: 7.0.2
 appVersion: 10.1.5
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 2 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 7.0.1
+version: 7.0.2
```
pvallone, line 3: I consider this a bug fix, so I bumped the patch version.

zanhsieh (approved):

buroa:
This completely breaks _any_ customization of init containers to only the variables specified (name, image, volume mounts):

https://github.com/buroa/k8s-gitops/blob/master/kubernetes/apps/monitoring/grafana/app/helmrelease.yaml#L28-L34

We should maintain functionality, so either revert or revise.

charts/grafana/Chart.yaml
samip5: I'm sorry but this would be classified as a breaking change and as such a patch bump is not sufficient.

Issue #2759, opened 2023-11-01
[grafana] sidecar datasources is deployed twice (init and standard) if init is desired


bt909:
In Grafana Helm Chart 7.0.2 the sidecar for datasources is deployed as init container and as regular container if ` .Values.sidecar.datasources.initDatasources` is selected, but I think it makes no sense. 
If I want to use this as init container, than this container should not be deployed as regular container as well.

bt909:
After thinking about, maybe it makes sense to have both container running. If you think it is more generic to have both, than this issue and related PR can be closed.

I use Grafana with anonymous login and at the moment it is not useful to have both variants (init-container and container), see https://github.com/grafana/helm-charts/issues/981

bt909:
Hm, the more I think about, the more I am confused. What is the use case to have the container running as init container and as sidecar container with value `skipReload: true`? What is the sidecar container doing in this case? Can anyone explain?

bt909:
If this is not a useful situation (running as sidecar with skipReload true), I can contribute the chart to have always only init container or sidecar and not both. Maybe it is an idea to remove skipReload key? Because it wouldn't be needed if the support support either init container or sidecar.

bt909:
closed with: https://github.com/grafana/helm-charts/pull/2760
thx

PR #2760, opened 2023-11-01
fix(grafana): sc datasources only w/o init container config

bt909:
If init container is desired, only init container should be deployed and not both (init and standard), see: https://github.com/grafana/helm-charts/issues/2759


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 7.0.2
+version: 7.1.0
```
```suggestion
version: 7.0.3
```

zanhsieh (changes requested):

zanhsieh (approved):

Xtigyro (approved):

bt909:
If there is anything I should do, let me know. 
With this PR we have then the same behavior on the alert and datasources sidecar, but as I know the notifier sidecar is running twice if configured as sidecar. If you want, I can contribute also a change for the behavior.

Issue #981, opened 2022-01-20
Data sources are not loaded properly when Grafana is configured with anonymous log in


mihail-sky:
Hi, we've recently upgraded Grafana from 8.3.1 to 8.3.3 (helm chart versions 6.18.2 to 6.20.5) and noticed that the data sources are not loading as expected.

What we can see in the logs for the _grafana-sc-datasources_ sidecar:
```
Starting collector
No folder annotation was provided, defaulting to k8s-sidecar-target-directory
Selected resource type: ('secret', 'configmap')
Loading incluster config ...
Config for cluster api at 'https://192.168.176.1:443' loaded...
Unique filenames will not be enforced.
5xx response content will not be enabled.
Working on ADDED configmap monitoring/grafana-datasource-prometheus
POST request sent to http://localhost:3000/api/admin/provisioning/datasources/reload. Response: 403 Forbidden {
  "message": "Permission denied"
}

Working on ADDED configmap monitoring/grafana-datasource-bigquery
POST request sent to http://localhost:3000/api/admin/provisioning/datasources/reload. Response: 403 Forbidden {
  "message": "Permission denied"
}
```

I suspect this is happing because the [Admin HTTP API](https://grafana.com/docs/grafana/latest/http_api/admin/#admin-api) request basic authentication to be enabled whereas on our Grafana we've disabled this.

It seems the REQ_URL and REQ_METHOD are always specified starting from this commit(https://github.com/grafana/helm-charts/commit/ee0914e19e680a6cde7a5a4e9bcb43f2bef9517a). If I remove these two parameters, everything works as expected.

Should we introduce a variable in the helm chart that removes these two variables similar to the GF_SECURITY_ADMIN_USER and GF_SECURITY_ADMIN_PASSWORD which leave out the REQ_USERNAME and REQ_PASSWORD?

I'm happy to raise a PR for this if you agree is the way forward.

nokernel:
This issue is still there in 6.26.2

c4tz:
This problem (with another HTTP status code) seems to also appear when specifying `GF_SECURITY_ADMIN_USER` and `GF_SECURITY_ADMIN_PASSWORD` in `.Values.env`.

By doing so, `REQ_USERNAME` and `REQ_PASSWORD` are left out for the sidecar container, making the reload call fail with:
```
POST request sent to http://localhost:3000/api/admin/provisioning/datasources/reload. Response: 401 Unauthorized {"message":"Unauthorized"}
```

[...]

PR #2761, opened 2023-11-02
[grafana] allow sc-alerts as init container

bt909:
Sidecar for datasources and notifiers are available, but not for alerts.  As values.yaml in line 873 suggested, the idea is to have it as init container as well.

This PR enables the possibility to use sidecar-alerts as init-container too.

Why do we want this as init-container? To get alerts loaded, when using Grafana in anonymous mode. For datasources we have a solution, but not for alerts.

Some kind of related (but this PR don't fix it): https://github.com/grafana/helm-charts/issues/981

### DIFF
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -206,7 +206,7 @@ need to instead set `global.imageRegistry`.
 | `sidecar.datasources.resource`            | Should the sidecar looks into secrets, configmaps or both. | `both`                               |
 | `sidecar.datasources.reloadURL`           | Full url of datasource configuration reload API endpoint, to invoke after a config-map change | `"http://localhost:3000/api/admin/provisioning/datasources/reload"` |
 | `sidecar.datasources.skipReload`          | Enabling this omits defining the REQ_URL and REQ_METHOD environment variables | `false` |
-| `sidecar.datasources.initDatasources`     | Set to true to deploy the datasource sidecar as an initContainer in addition to a container. This is needed if skipReload is true, to load any datasources defined at startup time. | `false` |
+| `sidecar.datasources.initAlerts`     | Set to true to deploy the datasource sidecar as an initContaine. This is needed if skipReload is true, to load any datasources defined at startup time. | `false` |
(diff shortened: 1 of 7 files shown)


bt909:
I have implemented as init-container OR container. If you like to have the init-container additional to the container, I can adjust that. In my usecase, I use Grafana w/ anonymous login and the container additional to init-container makes no sense in my usecase, but maybe to have both is the more generic way.

charts/grafana/README.md
```diff
@@ -206,7 +206,7 @@ need to instead set `global.imageRegistry`.
 | `sidecar.datasources.resource`            | Should the sidecar looks into secrets, configmaps or both. | `both`                               |
 | `sidecar.datasources.reloadURL`           | Full url of datasource configuration reload API endpoint, to invoke after a config-map change | `"http://localhost:3000/api/admin/provisioning/datasources/reload"` |
 | `sidecar.datasources.skipReload`          | Enabling this omits defining the REQ_URL and REQ_METHOD environment variables | `false` |
-| `sidecar.datasources.initDatasources`     | Set to true to deploy the datasource sidecar as an initContainer in addition to a container. This is needed if skipReload is true, to load any datasources defined at startup time. | `false` |
+| `sidecar.datasources.initAlerts`     | Set to true to deploy the datasource sidecar as an initContaine. This is needed if skipReload is true, to load any datasources defined at startup time. | `false` |
```
zanhsieh, line 209:
```suggestion
| `sidecar.datasources.initAlerts` | Set to true to deploy the datasource sidecar as an initContainer. This is needed if skipReload is true, to load any datasources defined at startup time. | `false` |
```

zanhsieh (changes requested):
@bt909 
Can you fix the test error?

bt909:
Hm, I would like to fix the test, but I am a bit confused. The test failed since I merged main....

bt909:
Oh, I am stuck in rebase hell, because of DCO. I will make an new clean PR.

bt909:
I will close this PR in favor of: https://github.com/grafana/helm-charts/pull/2771
Sorry for that.

Issue #2762, opened 2023-11-02
[grafana] Alerts not loaded using Grafana w/ anonymous login


bt909:
When Grafana is used with anonymous login, the sidecar for alerts is running as a standard container in the pod and is not able to trigger Grafana reload, so the alerts are not loaded.

Allowing sidecar-alerts to be a init-container would solve this issue.

Related PR:
https://github.com/grafana/helm-charts/pull/2761

bt909:
The mentoined PR is merged and solve this issue, so I will close it as resolved. Thx for merging.

PR #2771, opened 2023-11-07
[grafana]: allow alerts sidecar as init container

bt909:
Sidecar for datasources and notifiers are available, but not for alerts. As values.yaml in line 873 suggested, the idea is to have it as init container as well.

This PR enables the possibility to use sidecar-alerts as init-container too.

Why do we want this as init-container? To get alerts loaded, when using Grafana in anonymous mode. For datasources we have a solution, but not for alerts.

Some kind of related (but this PR don't fix it): https://github.com/grafana/helm-charts/issues/981

This PR was created, but I endet up in git rebase hell, so this is successor of https://github.com/grafana/helm-charts/pull/2761

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 7.0.3
+version: 7.0.4
 appVersion: 10.1.5
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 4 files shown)


zanhsieh (approved):

zanhsieh:
@zalegrala
Can you review this please?

zalegrala:
Thank you for the PR.  Is the sidecar expected to continue running during operation?

bt909:
No, my idea is to have to run it as init container or sidecar, not both. But if you prefer to have running them both, I can adjust.
The datasource sidecar is configured to running as init container and sidecar (and I have no real idea why this should be done), so I have a PR for the datasource sidecar as well, to configure either init or sidecar (but what about the option skipReload?).
I have some of my thoughts about it here:
https://github.com/grafana/helm-charts/issues/2759 and as issue for this PR: https://github.com/grafana/helm-charts/issues/2762

zalegrala:
If I understand correctly then, the current situation is that the sidecars start, do their work and then die?  If so, I think your right that making these init containers makes some sense here.  Have you been able to test this change to confirm its functionality?

bt909:
As you see, I had to change to watch method to "LIST", that this container just list the alerts and put them in the right place and die afterwards. 
Thank you for your demand to test. Yes, I tested it and it works now as expected.
But @zanhsieh, I add some change, would you like to review again? Latest change is init-container name for alerts init container and watch method (change from watch to list, that the container behaves as init container and die and don't want to watch all the time, sorry that I missed that).

zalegrala (approved):

Xtigyro (approved):

dbluxo:
Can this be merged? Waiting for this feature 😁

PR #2775, opened 2023-11-08
[loki-distributed] config: expose "auth_enabled" and extra properties for "common"

evenh:
As we're testing migration from the [loki](https://github.com/grafana/helm-charts/tree/main/charts/loki) chart to loki-distributed, we found that we were missing a couple of properties. This PR introduces support for toggling whether `auth_enabled` is enabled or not. In addition it lets us plug in extra configuration to the `common` block of the configuration file.

I'm open to relocating these properties to other sections of the values structure if it makes more sense or if they need to be renamed.

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.9.2
-version: 0.76.1
+version: 0.76.2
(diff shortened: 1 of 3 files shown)


evenh:
There seems to be a flake pulling Docker images in CI. Could it please be retested? :-)

verejoel:
Why not just put this in `structureConfig`? IMO this PR is not necessary as the required changes can be injected on a case by case basis. No need to extend the existing values.

evenh:
Thanks @verejoel, I will try to go down that route.

PR #2785, opened 2023-11-14
[rollout-operator] Add support for webhooks

m1rp:
- updated deployment and service to expose correct ports
- added role, clusterroles and associated bindings (needed for CA creation)
- added webhooks needed for downscale/no-downscale
- updated readme


charts/rollout-operator/templates/webhook-role/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,35 @@
...
+      namespace: {{.Release.Namespace}}
+      path: /admission/no-downscale
+      port: 443
+  failurePolicy: Fail
+  matchPolicy: Equivalent
+  rules:
+  - apiGroups: [apps]
+    apiVersions: [v1]
+    operations: [UPDATE]
+    resources:
```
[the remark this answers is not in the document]
m1rp: 🤔 this part was mostly a copy/paste job [from the docs](https://github.com/grafana/rollout-operator#admissionno-downscale-1) 
if this PR needs updating then the docs are potentially slightly off.

charts/rollout-operator/templates/webhook-role/mutatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,36 @@
...
+webhooks:
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: {{ include "rollout-operator.fullname" .}}
+      namespace: {{.Release.Namespace}}
+      path: /admission/prepare-downscale
+      port: 443
+  failurePolicy: Fail
```
colega, line 19: This may create a deadlock when applying it. If this webhook is created but rollout-operator isn't updated yet, then the rollout operator won't have the https endpoint and will always fail, rejecting any operation like trying to update the rollout operator.

It should be deployed with `Ignore` strategy first, and then with `Fail`. Note that `Ignore` only ignores failed calls, it doesn't ignore explicit rejections.

m1rp: is this only true for the `mutatingwebhookconfiguration.yaml` or does this also apply to the `validatingwebhookconfiguration.yaml`? 

is something like this:
```
 {{- if eq((lookup "v1" "MutatingWebhookConfiguration" "{{.Release.Namespace}}" ) "" ) }}
  failurePolicy: Ignore
  {{- else }}
  failurePolicy: Ignore
  {{- end }}
```
what you had in mind?

colega: This applies to both webhooks. 

I don't understand the condition of your diff (not saying it's wrong, I'm just not familiar with Helm), but I could imagine this as an extra setting, something to be disabled by default, and to be enabled after rolling out for more robustness (so it wouldn't allow everything if rollout-operator is just crashing).

Also, your diff says `Ignore` in both branches :D

m1rp: thanks for the update/clarification, i think i get what you're saying.

charts/rollout-operator/templates/webhook-role/validatingwebhookconfiguration.yaml
```diff
@@ -0,0 +1,40 @@
...
+- name: no-downscale-{{.Release.Namespace}}.grafana.com
+  admissionReviewVersions: [v1]
+  clientConfig:
+    service:
+      name: {{ include "rollout-operator.fullname" .}}
+      namespace: {{.Release.Namespace}}
+      path: /admission/no-downscale
+      port: 443
+  {{- $failpol := (lookup "v1" "ValidatingWebhookConfiguration" "{{.Release.Namespace}}" "no-downscale-{{.Release.Namespace}}") -}}
+  {{- if empty $failpol }}
```
colega, line 20: I think it would be nice to have an explicit switch that would disable the failurePolicy, just in case something goes wrong here (like: the webhook is created, but the rollout-operator isn't updated so it's not updating its TLS certs)

m1rp: by disable, do you mean set `failurePolicy: Ignore`?
sorry i'm a bit out of the loop honestly this PR has been going on for a while 😄

m1rp:
could i get an update on this? 
unfortunately i don't have as much time as is used to to spend on this so updates on my end are really slow

paulojmdias:
Hi @m1rp, You want to revive this PR, or can I pull it in another PR with your reference?

m1rp:
hiya @paulojmdias, not sure i can "revive" this as the issue is closed.
also looks like your PR was already merged, so i'm not sure if this is necessary anymore.
> can I pull it in another PR with your reference?

yes!

tcp13equals2:
A related PR has been created at https://github.com/grafana/helm-charts/pull/3859/files

PR #2789, opened 2023-11-15
[grafana] fix: Change sidecar.alerts.initDatasources to sidecar.alerts.initAlerts

kvanzuijlen:
Missed in #2760

~@zanhsieh Not sure if this should actually be a major version bump?~


zanhsieh (approved):

bt909:
Why should be initAlerts changed to initDatasources? This is a value for the alert sidecar and has nothing to to with datasources?

charts/grafana/values.yaml
```diff
@@ -872,7 +872,7 @@ sidecar:
     skipReload: false
     # This is needed if skipReload is true, to load any alerts defined at startup time.
     # Deploy the alert sidecar as an initContainer.
-    initAlerts: false
+    initDatasources: false
```
bt909, line 875: this is affecting only the sidecar alerts and should NOT named initDatasources. Why do you want to change this?
This came with: https://github.com/grafana/helm-charts/pull/2771

Is this what you are looking for in line 980?
https://github.com/grafana/helm-charts/blob/grafana-7.0.6/charts/grafana/values.yaml#L980

bt909 (changes requested):

charts/grafana/values.yaml
kvanzuijlen: Ah, I was completely confused by the changes from #2760. It's changes in the README.md are incorrect. I'll update this PR to fix those mistakes.

bt909: Thank you very much. I saw this PR and were shocked for the moment, because I introduced initAlerts (for this init container) and use this and thought this will be removed. That would be not that good.
But 2760 configures sidecar datasources only as sidecar, but the comment in line 978 needs to be adjusted, because this is not the case anymore. If you would like to adjust this, thank would be great.

Xtigyro (approved):

zanhsieh (approved):

PR #2813, opened 2023-11-30
[tempo-distributed] add optional statefulset and traces_storage config for metrics generator

jdgeisler:
Adding support for the new experimental metrics summary api capability https://grafana.com/docs/tempo/latest/api_docs/metrics-summary/

Specifically, the docs mention that this feature will cause the generator to use considerably more resources, including disk space. To handle this at scale and to ensure data persistence and availability, I am adding support for the metrics-generator to optionally be deployed as a statefulset with a pvc.

If `kind: StatefulSet` is set and persistence is enabled, then the metrics-generator will be deployed as a statefulset instead of a deployment and the walEmptyDir won't be used. To ensure backwards compatibility though, by default the metrics-generator is still a deployment with the walEmptyDir.

Additionally, the metrics summary api requires a new config `traces_storage`. This is used by the local blocks processor to store a wal for the traces. https://github.com/grafana/tempo/blob/main/modules/generator/config.go#L31
- Without this config, the following error is thrown 
```
level=warn ts=2023-11-29T19:29:38.039769755Z caller=server.go:1184 method=/tempopb.MetricsGenerator/PushSpans duration=5.298406ms msg=gRPC err="could not initialize processors: local blocks processor requires traces wal"
```


gespi1:
any updates here. This is very useful as it is recommended in [tempo 2.2.0+ to use statefulsets to support a new `processor`](https://github.com/grafana/tempo/releases/tag/v2.2.0) for the metrics-generator

zanhsieh (approved):

mapno (approved):
Thanks!

PR #2828, opened 2023-12-08
Remove the deprecated minio chart from the call-update-helm-repo GH action

MichelHollands:
The old min.io helm is no longer available. This is blocking Loki Helm chart releases: see https://github.com/grafana/loki/actions/runs/7059033470/job/19281713160 for an example.


dimitarvdimitrov (approved):
thanks

PR #2221, opened 2023-02-20
[promtail] Allow parent charts to override registry hostname and credentials

johnswarbrick-napier:
Reworked PR #2219 with a rebase and incremented minor version.

Add support for optionally overriding the registry hostname and credentials from a parent chart.

This is important for air-gapped installations which do not have direct access to the public internet and must use a private, internal container registry.

It also avoids having to modify the Promtail Helm chart directly.

FAO: @unguiculus and @Whyeasy

PR #2833, opened 2023-12-10
[promtail] Avoid unnecessary pod restart on each helm chart version

jkroepke:
This PR changes the the input of the config checksum which triggers the reload.

On main branch, the whole configmap/secret checksum is included as checksum input. The configmap contains labels which includes the helm chart version. Each new helm chart version forces a pod restart which is not always necessary.

For example: #2221 would not require a pod restart.


zanhsieh (approved):

jkroepke:
@zanhsieh I fixed the version conflict.

zanhsieh:
@zalegrala Can you review this PR please?

zalegrala:
I believe this will mean that you never restart your pods, even if there is a major version bump, correct?  Are we sure this is desireable?

jkroepke:
If a major version of a helm chart does not modify config or podSpec, whats the reason for the restart?

zalegrala:
Would a major version bump of promtail be caught here?

jkroepke:
> Would a major version bump of promtail be caught here?

Sure, if the major version bump changes podSpec (which includes the image version) or config, it would restart the promtail as expected.

zalegrala (approved):

PR #2853, opened 2023-12-15
improvement: Allow for external config as ConfigMap.

dtrejod:
## Problem
Currently there are 2 issues this PR is attempting to solve. 
1. In order to use [`existingSecretForConfig`](https://github.com/grafana/helm-charts/blob/cc1ebb8174f6a72dfb7aed54b46310fba7184885/charts/loki-distributed/values.yaml#L77-L78), the config must be a K8s Secret. In comparison within the [Mimir](https://github.com/grafana/mimir/blob/facbb92120699d0ccbdc9c24fe9006596f504a6f/operations/helm/charts/mimir-distributed/values.yaml#L83-L89) helm-chart it's possible to have an external config be a ConfigMap when `.Values.configStorageType == "ConfigMap"`. Having the external config be a ConfigMap over a Secret allows for operators to more quickly see the current Loki configuration without having to base64 decode the value if it were a Secret. 
2. The lack of existence of a `externalConfigVersion` entirely. The `externalConfigVersion` allows operators to change this field to force the `checksum/config` annotation to change force rolling pods in the cluster. Today the `checksum/config` is always a shasum of the configmap regardless if `existingSecretForConfig` is set or not.

### ConfigMap vs Secret for existingSecretForConfig
Allows for the external config to be a ConfigMap rather than a Secret. The loki helm chart already has `.Values. configAsSecret` which is a bool. So instead of introducing a new variable to align with the Mimir corresponding `.Values.configStorageType` variable, I opted to reuse this existing variable to toggle when a ConfigMap is used for the external config. 

### Introduce externalConfigVersion
This PR introduces a new field `externalConfigVersion` in `.Values` that will allow for versioning of the `checksum/config` annotation. The enterprise-logs chart already has a this as a `.Value` so this copies that variable naming convention. 
https://github.com/grafana/helm-charts/blob/cc1ebb8174f6a72dfb7aed54b46310fba7184885/charts/enterprise-logs/values.yaml#L41

## Notes for Reviewer
1. I'll be starting vacation on Dec 16, 2023, but will be available to respond to feedback/comments come Jan 3, 2024
2. I intend to follow up with similar changes to the enterprise-logs helm chart so it too can have an external config be a ConfigMap.
3. There does not seem to be a CHANGELOG.md for the loki-distributed helm chart. Happy to add a changelog where needed to document the above changes.

### DIFF
--- a/charts/loki-distributed/templates/_helpers.tpl
+++ b/charts/loki-distributed/templates/_helpers.tpl
@@ -153,7 +153,7 @@ Return the appropriate apiVersion for HorizontalPodAutoscaler.
 {{- end -}}
 
 {{- define "loki.ingester.readinessProbe" -}}
-{{- with .Values.ingester.readinessProbe }}  
+{{- with .Values.ingester.readinessProbe }}
 readinessProbe:
   {{- toYaml . | nindent 2 }}
 {{- else }}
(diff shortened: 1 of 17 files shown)


charts/loki-distributed/templates/_helpers.tpl
```diff
@@ -177,5 +177,30 @@ livenessProbe:
...
+{{- end }}
+
+{{/*
+The volume to mount for loki configuration
+*/}}
+{{- define "loki.configVolume" -}}
+{{- if .Values.loki.existingSecretForConfig }}
+{{- if .Values.loki.configAsSecret }}
+secret:
+  secretName: {{ .Values.loki.existingSecretForConfig }}
```
dtrejod, line 194: 💭 I think this maybe a break for some users if previously they were using `existingSecretForConfig` but didn't set `configAsSecret: true`. The break would be with these proposed changes they would need to also set `configAsSecret: true` to ensure the externalConfig is still pulled in from a Secret rather than a ConfigMap.

charts/loki-distributed/values.yaml
```diff
@@ -76,6 +76,9 @@ loki:
     allowPrivilegeEscalation: false
   # -- Specify an existing secret containing loki configuration. If non-empty, overrides `loki.config`
   existingSecretForConfig: ""
+  # -- When 'existingSecretForConfig' is not empty, then changing 'externalConfigVersion' triggers restart of services -
+  # otherwise changes to the configuration cause a restart.
+  externalConfigVersion: '0'
```
dtrejod, line 81: 💭 None of these variables seem to be documented in the README.md for loki-distributed. Therefore I left out docs that explain the addition of this new `.Values` variable.

PR #2866, opened 2023-12-21
[grafana] Enhance Deployment Template Flexibility for Zero Replicas

anotherglitchinthematrix:
I've encountered an issue while attempting to deploy the Grafana chart with zero replicas. The problem arises from the replicas value being set to zero, which falsifies an if block in the deployment template, leading Kubernetes to default to a replication count of 1.

This pattern is not unique and is present in multiple deployment templates across the repository. The current implementation limits configuration options for manual intervention, restricting users from customizing the replication count as needed.

I suggest modifying this pattern in other deployment templates within the repository to improve consistency and provide users with greater flexibility in their configurations.

The current pattern undermines some configuration opportunities for manual intervention, hindering users from tailoring deployment settings according to their specific requirements.

Your feedback on these proposed changes is highly appreciated. Please share your thoughts on whether this modification aligns with our goals for configuration flexibility. Let's discuss the best approach to enhance the user experience in managing deployments.


zanhsieh (changes requested):
@anotherglitchinthematrix 
Can you bump the version in Chart.yaml and sign DCO please?

zanhsieh (approved):

zanhsieh:
@jkroepke @Xtigyro 
Can you review this?

charts/grafana/templates/deployment.yaml
```diff
@@ -14,7 +14,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  {{- if (not .Values.autoscaling.enabled) }}
+  {{- if and (not .Values.autoscaling.enabled) (hasKey .Values "replicas") }}
```
jkroepke, line 17: To my best knowledge, this does not work, if grafana is embedded as sub-chart

anotherglitchinthematrix: This issue seems to be resolved in the recent versions

anotherglitchinthematrix:
This issue seems to be resolved in the recent versions

This PR was fixing the falsy evaluation of the `(.Values.replicas)` when set to zero.

```diff
- {{- if and (not .Values.autoscaling.enabled) (.Values.replicas) }}
+ {{- if and (not .Values.autoscaling.enabled) (hasKey .Values "replicas") }}
```

The current conditional block eliminated that case.

```
{{- if (not .Values.autoscaling.enabled) }}
```

PR #2908, opened 2024-01-15
Add support for autoscaling for metrics-generator component

shohamy7:
# Description
This PR aims to add support for autoscaling for the metrics generator component.

# Use case/Motivation
As we increase the number of metrics-generator instances - the total active series will increase as well.
However, when we have less traffic, we can decrease the replicas number and essentially decrease the number of active series generated.
Adding autoscaling will enable this use case - we can scale down when we have less traffic and scale back to normal when we need to.
Also, added a note about using this feature that explains the consequences that may happen when using feature.


zanhsieh (changes requested):
@shohamy7 
Can you bump the version in Chart.yaml and update README.md via helm-doc?

shohamy7:
@zanhsieh 
done :)

PR #2918, opened 2024-01-18
[grafana-sampling] add sampling helm chart

rlankfo:
This helm chart deploys a layered set of agents that can be used for OTLP load balancing, metrics generation, and trace sampling. This provides an opinionated solution that customers can use as somewhat of a "black box" on top of the agent. It does quite a bit of the river/flow configuration heavy-lifting.


charts/sampling/templates/_otelcol_processor_transform.river.txt
```diff
@@ -0,0 +1,69 @@
...
+      "set(attributes[\"pod\"], attributes[\"k8s.pod.name\"])",
+      "set(attributes[\"namespace\"], attributes[\"k8s.namespace.name\"])",
+      "set(attributes[\"loki.resource.labels\"], \"pod, namespace\")",
+    ]
+  }
+
+  trace_statements {
+    context    = "resource"
+    statements = [
+      "set(attributes[\"k8s.cluster.name\"], \"{{ .Values.k8s.cluster.name }}\")",
```
mar4uk, line 39: in k8s-monitoring-helm they [always check](https://github.com/grafana/k8s-monitoring-helm/blob/8e8c0fb312508ced0f4c6b9b12fe02e1606f63cf/charts/k8s-monitoring/templates/agent_config/_traces.river.txt#L21
) if `{{ .Values.k8s.cluster.name }}` exists:
```
// Accessing a map with a key that does not exist will return nil.
  "set(attributes[\"k8s.cluster.name\"], \"{{ required ".Values.cluster.name is a required value. Please set it and try again." .Values.cluster.name }}\") where attributes[\"k8s.cluster.name\"] == nil",
```

we probably should do the same here

charts/sampling/templates/_prometheus_remote_write.river.txt
```diff
@@ -0,0 +1,35 @@
...
+    basic_auth {
+      username = env("GRAFANA_CLOUD_PROMETHEUS_USERNAME")
+      password = env("GRAFANA_CLOUD_API_KEY")
+    }
+    queue_config {
+      retry_on_http_429 = false
+    }
+  }
+  external_labels = {
+    "__metrics_gen_instance" = env("POD_UID"),
```
mar4uk, line 31: I started to think that adding this label would not fully work because when metrics generation is done on the edge, `target_info` metric name is used instead of `traces_target_info`.

Using "__metrics_gen_instance" will not apply adaptive metrics aggregations to `target_info` metric, because default rule will apply to `traces_target_info` metric name: https://github.com/grafana/deployment_tools/blob/44243c57fcb391d8853061210fe7bbe611bf5742/ksonnet/lib/mimir-aggregations/default-aggregations.json#L127-L134

charts/sampling/templates/_otelcol_processor_transform.river.txt
```diff
@@ -0,0 +1,69 @@
...
+    ]
+  }
+
+  output {
+    logs   = [otelcol.processor.batch.default.input]
+    traces = [otelcol.exporter.loadbalancing.default.input]
```
mar4uk, line 45: shall batch processor be used for traces pipeline as well?

charts/sampling/templates/_otelcol_processor_k8sattributes.river.txt
```diff
@@ -0,0 +1,32 @@
+{{- define "deployment.processor.k8sattributes" -}}
+otelcol.processor.k8sattributes "default" {
```
cyrille-leclerc, line 2: Is this enriching incoming OTel signals with K8s attributes or is it just for signals produced by the grafana agent for its internal behavior?

Shouldn't we require that the incoming OTLP signals are properly enriched before entering the "sampling box" with all the `k8s.*`, `cloud.*`, and `host.*` metadata? Moreover, if we enrich incoming signals with K8s metadata in the samplign box, aren't we at risk of adding the wrong context?

charts/sampling/templates/_otelcol_processor_transform.river.txt
```diff
@@ -0,0 +1,69 @@
+{{- define "deployment.processor.transform.add_resource_attributes" -}}
+otelcol.processor.transform "add_resource_attributes_as_metric_attributes" {
```
cyrille-leclerc, line 2: Ditto, is this for incoming metrics or for metrics produced internally?
I'm not sure to catch the role for a "sampling box".

charts/sampling/templates/_otelcol_processor_transform.river.txt
```diff
@@ -0,0 +1,69 @@
...
+
+  output {
+    metrics = [otelcol.processor.batch.default.input]
+  }
+}
+
+{{ end }}
+
+{{- define "deployment.processor.transform.add_k8s_attributes" -}}
+otelcol.processor.transform "add_k8s_attributes" {
```
cyrille-leclerc, line 22: Ditto, is this for incoming logs&traces or for logs&traces produced internally?
I'm not sure to catch the role for a "sampling box".

cyrille-leclerc:
> This helm chart deploys a layered set of agents that can be used for OTLP load balancing, metrics generation, and trace sampling.

Doing RED & Svc Graph Metrics Generation + Sampling with the needed load-balancing for scalability is what I expected. I don't catch why there is also a K8s attributes processor, I think we don't want to enrich the incoming signals with any contextual metadata, we require metadata to be provided before entering the sampling box.

LittleLadySunshine:
@cyrille-leclerc with @rlankfo, I might not be able to get this addressed until morning. FYI. @gouthamve @mar4uk

mar4uk:
This is the architecture deployed by this helm chart:
sampling-recommended-arch-agent-scaled

The first layer (deployment) receives all OTLP signals and enriches them with k8s.* resource attributes. Then it exports metrics (app metrics) and logs to Grafana Cloud, and load-balances traces to the second layer with load-balancing exporter.

The second layer (statefulset) receives traces from the first layer does metrics generation and sampling, and exports generated metrics and sampled traces to Grafana Cloud.

So basically, the second layer is the "sampling box", it indeed receives enriched data

cyrille-leclerc:
As discussed with @mar4uk , we don't see reasons to enrich incoming OTel  signals with K8s Metadata.

The `k8s.*` enrichment, as well as the `cloud.*` and `host.*` enrichments, should be done before the Grafana Sampling layer as shown in the architecture diagram below
image

mar4uk:
Shall we put this diagram (the one Cyrille added) to the [DesignDoc](https://docs.google.com/document/d/1DB3jagr6V-2TWs-DRhPddz73lHgYu1upIPg7rVFUxUw/edit#heading=h.3ik6vnncm0n)?
So it would be clear that the sampling helm chart covers only the Sampling Box part of the diagram.

The data enrichment processors
- `otelcol.processor.transform "add_resource_attributes_as_metric_attributes"`
- `otelcol.processor.transform "add_k8s_attributes"`
- `otelcol.processor.k8sattributes "default"`
- `resourcedetectionprocessor` (later, when it is released. As a separate PR)

are going to be moved out of the Sampling Box and be a part of k8s-monitoring helm chart (they are already a part of the k8s-monitoring helm chart)

wdyt, Robbie?

cyrille-leclerc:
> The data enrichment processors ... are going to be moved out of the Sampling Box and be a part of k8s-monitoring helm chart (they are already a part of the k8s-monitoring helm chart)

Do we want to enrich meta-monitoring metrics and logs. I guess Grafana Agent's OTel Collector components publish meta-monitoring metrics similarly to what the Otel Collector does.

rlankfo:
> Shall we put this diagram (the one Cyrille added) to the [DesignDoc](https://docs.google.com/document/d/1DB3jagr6V-2TWs-DRhPddz73lHgYu1upIPg7rVFUxUw/edit#heading=h.3ik6vnncm0n)? So it would be clear that the sampling helm chart covers only the Sampling Box part of the diagram.
> 
> The data enrichment processors
> 
> * `otelcol.processor.transform "add_resource_attributes_as_metric_attributes"`
> * `otelcol.processor.transform "add_k8s_attributes"`
> * `otelcol.processor.k8sattributes "default"`
> * `resourcedetectionprocessor` (later, when it is released. As a separate PR)
> 
> are going to be moved out of the Sampling Box and be a part of k8s-monitoring helm chart (they are already a part of the k8s-monitoring helm chart)
> 
> wdyt, Robbie?

I'll remove these for now but my gut feeling is that not everyone who uses this helm chart will use the k8s monitoring helm chart as well and eventually we may get requests to make some of these processors configurable in the chart.

rlankfo:
> Do we want to enrich meta-monitoring metrics and logs? I guess Grafana Agent's OTel Collector components publish meta-monitoring metrics similarly to what the Otel Collector does.

I don't think we want to enrich any debug metrics coming out of the agent.  I believe these metrics would need to be scraped from the agents /metrics endpoint.

LittleLadySunshine:
Let's make sure we keep meta monitoring on our horizon for future work. cc: @rlankfo

charts/sampling/templates/_otelcol_processor_k8sattributes.river.txt
rlankfo: I'll remove these as discussed; however, I imagine users will ultimately want this as a configuration option given that they may not be using this chart in tandem with our k8s monitoring chart. 

> Is this enriching incoming OTel signals with K8s attributes or is it just for signals produced by the grafana agent for its internal behavior?

This was intended to enrich the OTEL signals. Collection of agent metrics would probably occur elsewhere imo, although we could eventually add an option to have these agents scrape themselves to collect the debug metrics.

> Shouldn't we require that the incoming OTLP signals are properly enriched before entering the "sampling box" with all the `k8s.*`, `cloud.*`, and `host.*` metadata? Moreover, if we enrich incoming signals with K8s metadata in the sampling box, aren't we at risk of adding the wrong context?

I don't think this should be a hard requirement but for the first milestone we can exclude any enrichment. As already mentioned, I imagine customers will likely ask for this possibility.  I don't see the risk of adding the wrong context -- the information is sourced by the connection IP with this configuration in the deployment.

charts/sampling/templates/_otelcol_processor_transform.river.txt
rlankfo: Given that we're excluding any enrichment for now, I'll just remove this.

charts/sampling/templates/_prometheus_remote_write.river.txt
rlankfo: @mar4uk you're correct, we'd also have to include a default rule for `target_info`.  As an alternative I can change this to `collector_id`, but we'd need default rules for all metrics with that label in this case.

charts/sampling/templates/_otelcol_processor_transform.river.txt
rlankfo: initially I didn't enable this as the batch processor docs indicate this should happen after sampling; however, it also says that it should be configured on every collector:
"It is highly recommended to configure the batch processor on every collector. The batch processor should be defined in the pipeline after the memory_limiter as well as any sampling processors. This is because batching should happen after any data drops such as sampling."

wdyt?

charts/sampling/templates/_otelcol_exporter_loadbalancing.river.txt
```diff
@@ -0,0 +1,21 @@
+{{- define "deployment.exporter.loadbalancing" -}}
+otelcol.exporter.loadbalancing "default" {
+  // https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.exporter.loadbalancing/
+  resolver {
+    dns {
+      hostname = "{{ .Release.Name }}-grafana-agent-statefulset.{{ .Release.Namespace }}.svc.cluster.local."
+    }
```
petewall, line 7: Does it make more sense to use the `kubernetes` resolver?
https://grafana.com/docs/agent/latest/flow/reference/components/otelcol.exporter.loadbalancing/#kubernetes-block

petewall (commented):
I'd love to see a quick README to briefly explain what's going on here.

Otherwise, looking good.

charts/sampling/templates/_otelcol_exporter_loadbalancing.river.txt
rlankfo: What are the main benefits it provides over the DNS resolver with a headless service? I think either resolve will work here. Currently this is setup to create a headless service and use the DNS resolver.  

We will likely have to make this resolver configurable however once we add support for multi-cluster setups where users having complex request life-cycles that live across more than a single kubernetes cluster. In this case, they'd need to deploy this sampling chart in each cluster but will have to provide ingress to the second layer, and the exporters will need to be configured to send to this ingress point.

mar4uk:
> I'll remove these for now but my gut feeling is that not everyone who uses this helm chart will use the k8s monitoring helm chart as well and eventually we may get requests to make some of these processors configurable in the chart.

We are going to advise users to use k8s monitoring helm chart if they run their applications in Kubernetes. If they prefer to not use k8s monitoring helm chart anyway (I'm wondering what could be the reasons for not using it), they would still deploy their custom Grafana Agent or Collector. We probably can provide some recommended configurations for this custom Agent/Collector. 

The downside of having enhancement processors in the sampling box in case when users don't use k8s monitoring helm chart is we will enhance only traces. Users still would need to enhance metrics and logs somewhere before the sampling box

charts/sampling/templates/_prometheus_remote_write.river.txt
mar4uk: yep. I created the dedicated page about how to configure adaptive metrics https://github.com/grafana/app-o11y/pull/626/files#diff-4ce44250fd12c678b647e0a4d2d119842486f6124e90c08a0005afb61cab96d4, the rules are not default, they need to be configured by user. Not ideal, but that could be good for the beginning. Later we can move it to default rules

charts/sampling/templates/_otelcol_processor_transform.river.txt
mar4uk: I think we need to add the batch processor.
The part "The batch processor should be defined in the pipeline after the memory_limiter as well as any sampling processors. This is because batching should happen after any data drops such as sampling." is applicable within one collector. In our case, we have two separate collectors, and we treat them separately

cyrille-leclerc: From what I remember, @gouthamve recently said about a Comply Advantage escalation that the memory limiter processor should be removed, that it may be deprecated.
@jpkrohling  @gouthamve  can you please help us?

charts/sampling/templates/_prometheus_remote_write.river.txt
rlankfo: Any reason not to just add these as default?

mar4uk: not really :) Probably only align with the Agent team to use the same rules (currently they suggested using `agent_id` label, instead of `collector_id`) 
I can create PR to deployment_tools and cc agent team

rlankfo:
> We are going to advise users to use k8s monitoring helm chart if they run their applications in Kubernetes. If they prefer to not use k8s monitoring helm chart anyway (I'm wondering what could be the reasons for not using it), they would still deploy their custom Grafana Agent or Collector. We probably can provide some recommended configurations for this custom Agent/Collector.

We provide the recommended configurations but we don't provide a way to deploy them. The suggestion was to make the processors configurable in this helm chart when users do not deploy k8s monitoring.  There could be several reasons such as cost, having other monitoring solutions implemented for k8s, etc.

> The downside of having enhancement processors in the sampling box in case when users don't use k8s monitoring helm chart is we will enhance only traces. Users still would need to enhance metrics and logs somewhere before the sampling box

This is not true. The way it was originally configured, it allowed to send metrics and logs through the first layer as well.

charts/sampling/templates/_otelcol_exporter_loadbalancing.river.txt
rlankfo: Updating to the k8s resolver as the DNS resolver could result in a several second delay after scaling in some cases. I'm also overcomplicating the multi-cluster setup and we can solve this by just adding ingress into the first layer and all clusters would send traces to this ingress.  This should usually not be necessary as requests living across clusters is probably more of an edge case. My assumption is most users will want this helm chart deployed once per cluster.

charts/sampling/templates/_agent_config_statefulset.river.txt
```diff
@@ -0,0 +1,17 @@
+{{- define "agent.config.statefulset" -}}
+  {{- include "statefulset.receiver.otlp" . }}
+  {{- if .Values.metricsGeneration.enabled -}}
+    {{- include "statefulset.connector.spanmetrics" . }}
```
mar4uk, line 4: I included one more processor to the config in docs `otelcol.processor.transform "drop_unneeded_resource_attributes"`: https://github.com/grafana/app-o11y/blob/main/docs/sources/setup/sampling/assets/agent-config-head-sampling.river#L96-L118

It is about dropping unneeded resource attributes in traces before passing them to spanmetrics connector.
That was feedback from Paulin, also I found we do the same for Tempo metrics generation: https://github.com/grafana/app-observability-plugin/blob/671336f758aaf3c0c7a1006c8c4d95baafcc749a/src/constants/overrides.ts#L11

We probably should include this processor in this helm chart as well

rlankfo: thanks @mar4uk, nice catch. I'll add this processor here as well.

hedss:
Hey folks. I've been pointed at this chart draft, and want to raise a few comments, not about the code itself but more a general conversation around usage.

We now have several different Agent charts for different purposes, including the 'vanilla' agent install and the k8s one. An issue with layering more charts in is that we're now starting to advise users to potentially put in entire pipelines of Agents in their infrastructure to get appropriate signals into our backend. This becomes problematic, and it's something we've seen in GTM.

The alternative is a user having to pick the chart that does the 'heaviest lifting' for them to start with, and start to override config, or at worst rewrite bits of charts, to actually do what they need.  I would very much welcome a conversation here, or elsewhere, to discuss this.

cyrille-leclerc:
Thanks for reaching out @hedss, with great pleasure to have a conversation, I'll DM you.

mar4uk:
Looks good! Left a couple of comments, I think they should be fixed before merging.

Did you exclude logs and metrics pipelines because they should be handled by a separate agent?

Let me know if you need help with the diagram update (logs and metrics arrows should be removed, right?)

rlankfo:
> Looks good! Left a couple of comments, I think they should be fixed before merging.
> 
> 
> 
> Did you exclude logs and metrics pipelines because they should be handled by a separate agent?
> 
> 
> 
> Let me know if you need help with the diagram update (logs and metrics arrows should be removed, right?)

Thanks Irina! I'll make these updates. I would always appreciate any help with the diagram. It should only support traces.

charts/grafana-sampling/templates/_otelcol_connector_spanmetrics.river.txt
```diff
@@ -0,0 +1,26 @@
...
+  histogram {
+    unit = "s"
+
+    explicit {
+      buckets = ["0s", "0.005s", "0.01s", "0.025s", "0.05s", "0.075s", "0.1s", "0.25s", "0.5s", "0.75s", "1s", "2.5s", "5s", "7.5s", "10s"]
+    }
+  }
+
+  output {
+    metrics = [otelcol.processor.transform.use_grafana_metric_names.input]
```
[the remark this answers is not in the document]
rlankfo: thanks @mar4uk, made the update!

mar4uk:
Hey Robbie! Will you update the diagram?

mar4uk (approved):

mar4uk:
LGTM!

LittleLadySunshine (approved):

PR #2959, opened 2024-02-09
[loki-distributed] Add kind & apiVersion to volumeClaims

rufusnufus:
[ArgoCD](https://argo-cd.readthedocs.io/en/stable/) always tries to delete kind & apiVersion from manifests to achieve GitOps, that leads to resyncing the loki-distributed Application
image

This PR adds those kind & apiVersion to volumeClaimTemplates


zanhsieh (approved):

Skaronator:
I had the same issue when enabling Server Side apply.

You need to enabled Server Side diff in argocd in order to fix this.

https://argo-cd.readthedocs.io/en/stable/user-guide/diff-strategies/#server-side-diff

Issue #2973, opened 2024-02-14
[loki-distributed] Compactor replica count should be adjustable


alita1991:
In my Kubernetes cluster, I'm utilizing Mimir, Loki, and Tempo, and I've observed that only the Loki compactor deployment has a fixed replica count of 1, which cannot be adjusted. I'm interested in scaling up or down the Compactor deployment to meet changing workload demands.

IamGroot19:
I thought this was a deliberate decision on the part of maintainers due to this warning mentioned in the docs:
> Note: There should be only 1 compactor instance running at a time that otherwise could create problems and may lead to data loss.
- source: https://grafana.com/docs/loki/latest/operations/storage/boltdb-shipper/#compactor

PR #2982, opened 2024-02-19
[loki-distributed] FIX: compactor replica configuration

ashleyprimo:
This is a fairly simple PR that fixes a minor oversight that prevents `replica` setting (_as shown [here](https://github.com/grafana/helm-charts/blob/071814887d3b8c49bca3f3d8724af763925413ab/charts/loki-distributed/values.yaml#L1335) from having any impact_) due to hard coding of the `replica` value within the `Deployment` template. 

Addressing issue: https://github.com/grafana/helm-charts/issues/2973


P6rguVyrst (commented):
I remember Loki compactor used to support only a single replica. 

I think that may be changed now? In which case - the chart may have to support compactor ring - not sure if it does right now :eyes: 

If this is possible, might need a bit more than a oneliner - maybe good people in Grafana know better :)

ashleyprimo:
Yep, I think you're correct. 

https://github.com/grafana/helm-charts/blob/071814887d3b8c49bca3f3d8724af763925413ab/charts/loki-distributed/values.yaml#L213 Is currently missing ring configuration.

Which should look something like: 
```
+      compactor_ring:
+        kvstore:
+          store: memberlist
```
Looking at the code: https://github.com/grafana/loki/blob/ffc61fbbf4990ccb00aaed107327fd5ff4d66e82/pkg/compactor/compactor.go#L87

zanhsieh (approved):

Sheikh-Abubaker (commented):
Hey @ashleyprimo Thanks for your contribution! unfortunately this chart is deprecated in favor of https://github.com/grafana/helm-charts/pull/3975, feel free to explore the issues and contribute to the charts that are actively being maintained. We encourage your valuable contributions to the community!

PR #2985, opened 2024-02-21
add support for zipkin receiver

dickon:
Currently the tempo chart expoes a port for zipkin but does not configure
zipkin as an ingester. This simply changes the config template to use the port it exposes. Without this, I was seeing incoming zipkin traffic getting their TCP connections dropped, and with this change it works fine.


zanhsieh (changes requested):
@dickon 
Can you bump the version in Chart.yaml and sign DCO please?

Sheikh-Abubaker:
@dickon could you please address the failed checks?

PR #2991, opened 2024-02-26
[grafana] Add value for pod.spec.automountServiceAccountToken

jkroepke:
fixes https://github.com/grafana/helm-charts/issues/2986
fixes https://github.com/grafana/helm-charts/pull/2987

PR #2977, opened 2024-02-15
[grafana] autoMount set to false in values.yaml

Sheikh-Abubaker:
Fixes #2976

PR #2997, opened 2024-02-27
[grafana] Add `serviceAccount.automountServiceAccountToken` and document `automountServiceAccountToken`

jkroepke:
Follow-up PR from #2991 

* Document `automountServiceAccountToken`
* Replace `serviceAccount.autoMount` by `serviceAccount.automountServiceAccountToken` - other charts using the same naming, too (https://github.com/grafana/helm-charts/blob/df7ae36c21bce12712c04a91115c165a04b7cdbf/charts/tempo/templates/serviceaccount.yaml#L20) - `serviceAccount.autoMount` is still supported and preferred to avoid a breaking change here.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 7.3.3
+version: 7.3.4
 appVersion: 10.3.3
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -137,6 +137,7 @@ need to instead set `global.imageRegistry`.
 | `extraSecretMounts`                       | Additional grafana server secret mounts       | `[]`                                                    |
 | `extraVolumeMounts`                       | Additional grafana server volume mounts       | `[]`                                                    |
 | `extraVolumes`                            | Additional Grafana server volumes             | `[]`                                                    |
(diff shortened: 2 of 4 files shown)


philipfreude:
Not sure if having both `automountServiceAccountToken` and `serviceAccount.automountServiceAccountToken` is really necessary. It may result in confusion.

In my opinion the best solution would be to have the same code state as **before** #2977 was merged. Just with the variable renaming and backwards compatibility you added.

Either way would work for me, but I have the feeling it overcomplicates things.

jkroepke:
> Not sure if having both `automountServiceAccountToken` and `serviceAccount.automountServiceAccountToken` is really necessary. It may result in confusion.

Yup, it is.

The security best practice is that set `automountServiceAccountToken: false` on each serviceAccount and set `automountServiceAccountToken: true` to all deployments they need one.

At least, this pattern make all security scanner more than happy.

* https://kyverno.io/policies/other/restrict-sa-automount-sa-token/restrict-sa-automount-sa-token/
* https://securecloud.blog/2021/08/17/azure-aks-reviewing-recommendations-from-security-center-disabling-automounting-api-credentials/
* https://avd.aquasec.com/misconfig/kubernetes/general/avd-ksv-0036/

> Either way would work for me, but I have the feeling it overcomplicates things.

Yeah, technically, the `automountServiceAccountToken` should disabled by default in Kubernetes. But this isn't possible unless we have Kubernetes 2.0 (which allows breaking changes) - https://github.com/kubernetes/kubernetes/issues/57601

Xtigyro (approved):

zanhsieh (approved):

PR #3016, opened 2024-03-11
upgrade helm-docs to v1.13.1

rlankfo:
This upgrades the helm-docs dependency from v1.8.1 to v1.13.1. I found in v1.8.1, some annotations were not properly honored.

PR #3019, opened 2024-03-12
update helm-docs to v1.13.1

rlankfo:
Signed-off-by: Robbie Lankford <robert.lankford@grafana.com>


rlankfo:
opening this PR to update CI only as suggested in https://github.com/grafana/helm-charts/pull/3016. This is failing as updating to this version of `helm-docs` results in changes to existing charts. 

imo it makes sense to merge this with the generated chart doc changes from this new version bypassing the CI check. @zanhsieh wdyt?

Footur:
I think some README files should be patched within this change.

zanhsieh (approved):

zalegrala (approved):

zalegrala:
@rlankfo Looks good.  Please also update the docs in this PR for the lint.

zanhsieh:
@zalegrala I don't think @rlankfo can update any doc to pass the PR lint. This PR is for Github workflow upgrade, not for a specific chart. We have to find someone who has force-merge permission (nor I have).

zalegrala:
Hmm, I see the lint failing because the new version writes a different doc output.  I think we'd need to commit those changes, or they will still fail on the next PR.  I think I'd rather update all the doc in one PR, rather than have each chart require a new PR with doc updates.  What do you think?

rlankfo:
> Hmm, I see the lint failing because the new version writes a different doc output. I think we'd need to commit those changes, or they will still fail on the next PR. I think I'd rather update all the doc in one PR, rather than have each chart require a new PR with doc updates. What do you think?

I agree, but that will cause a different lint failure, as it changes multiple charts in one PR, as seen here: https://github.com/grafana/helm-charts/pull/3016/files

We'll have to bypass one of the lint checks and merge with the failure.

zanhsieh:
@zalegrala I don't think update all doc in one PRs would work under the current CI implementation.

PR #3037, opened 2024-03-20
[grafana] fix: Add missing version, kind to volumeClaimTemplates

Nickmman:
This PR fixes the issue described in https://github.com/bitnami/charts/issues/15526, which is also present in the current helm chart.


zanhsieh (approved):

donovanrost:
@Xtigyro  Any chance you can find some time to review this? 
This is causing ArgoCD to show out of sync for kube-prometheus-stack

jkroepke (approved):
Thanks for linking the bitnami issue. that was worth reading it.

Sounds like bitnami has already qualified this change, so LGTM

PR #3044, opened 2024-03-26
[tempo-distributed] fix service monitor

KyriosGN0:
when using the tempo-distributed helm chart in argocd it complains its always out of sync
this is due to the face that in the service-monitor.tpl there is not definition to action. which is inserted automatically

this fixes that


KyriosGN0:
hey @mapno @swartz-k @BitProcessor @faustodavid, any chance i could get a review ?

mapno (approved):

PR #3054, opened 2024-04-02
fixing imagePullSecrets for enterprise components

jordan-simonovski:
Ran into the following issue when trying to deploy GET via ArgoCD:

```
Error: template: tempo-distributed/templates/enterprise-gateway/gateway-dep.yaml:40:23: executing "tempo-distributed/templates/enterprise-gateway/gateway-dep.yaml" at <.Values.image.pullSecrets>: nil pointer evaluating interface {}.pullSecrets Use --debug flag to render out invalid YAML
```

After some debugging, I found some differences in the way that the `pullSecrets` are derived in the `adminApi` and the `enterpriseGateway` which I've now standardised to fix up some of the complexity when configuring this.


charts/tempo-distributed/templates/admin-api/_helpers.tpl
```diff
@@ -0,0 +1,7 @@
+{{/*
+adminApi imagePullSecrets
+*/}}
+{{- define "tempo.adminApiImagePullSecrets" -}}
+{{- $dict := dict "tempo" .Values.tempo.image "component" .Values.adminApi.image "global" .Values.global.image -}}
+{{- include "tempo.adminApiImagePullSecrets" $dict -}}
```
zalegrala, line 6: Is this `include` desired here?  It looks to me like we are defining a function and then recursively calling it.  I suspect we want the same approach to all the components attempting to pull images from private registries, and so creating one template per component seems a little verbose.  Can we unify the templating?

zalegrala: Based on `templates/distributor`/helpers-distributor.tpl` it looks like we can change this include line to point to the top level include `tempo.imagePullSecrets`

```
{{/*
distributor imagePullSecrets
*/}}
{{- define "tempo.distributorImagePullSecrets" -}}
{{- $dict := dict "tempo" .Values.tempo.image "component" .Values.distributor.image "global" .Values.global.image -}}
{{- include "tempo.imagePullSecrets" $dict -}}
{{- end }}
```

jordan-simonovski: Yep that was my bad there. Thanks for that. I'm also looking at updating the `tokengenJob` to support this too.

zalegrala:
This looks good to me.  We need a version bump again for the PR that went in before this one.

zalegrala (approved):
Looks good to me.  Thanks for the contribution.

joe-elliott (approved):

PR #3059, opened 2024-04-03
[tempo] set unique cluster_label for tempo memberlist

hobbsh:
Addressing https://github.com/grafana/tempo/issues/2766 where Loki/Tempo ingesters can join each other's ring because cluster_label seems to be defaulted to something common to both components. This change should mitigate the issue in the future.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 1.11.0
+version: 1.12.0
 appVersion: 2.5.0
 engine: gotpl
 home: https://grafana.net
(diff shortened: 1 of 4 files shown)


bilsch:
A similar patch to mimir and loki should be made as well ( guessing separate pr?). I ran into an issue in my cluster where loki and mimir pods were seeing each others rings and causing all sorts of problems ( the loki app would not even start up correctly it just failed with strange errors )

Our fix was to set this cluster_label to a unique value for each of the 3 apps. When the pods restarted they correctly were ignored and the clusters formed happily

nlamirault:
any news on this feature ?
i've got some strange logs in Loki :( 

```
autoforget removed ingester tempo-ingester-2 from the ring because it was not healthy after 1m0s
```

joe-elliott (approved):

zalegrala (approved):

Sheikh-Abubaker (approved):

joe-elliott:
Thanks @hobbsh!  🙏

PR #3075, opened 2024-04-11
[tempo-distributed] Adds service specific features for enterprise gateway inline with gateway.

hedss:
Namely:
* Service type
* Service port
* Cluster IP address (if type is ClusterIP)
* Loadbalancer IP address (if type is LoadBalancer)

The gateway component already supports these, inline with similar config options in Mimir and Loki. The enterprise gateway component does not.

Can help in situations where specific cluster/loadbalancer port mappings are required for external IPs for a cluster.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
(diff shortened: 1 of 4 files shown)


zalegrala (approved):

joe-elliott (approved):
this probably won't destroy everything forever

hedss:
> this probably won't destroy everything forever

Phew, very glad to hear that!

PR #3111, opened 2024-05-03
[tempo-distributed] fix invalid selector configuration for tokengenjob

jordan-simonovski:
I've run into an error when trying to deploy the helm chart via ArgoCD which is getting stuck on trying to execute the `tokengen-job`.

The error is as follows:

```
  PostSync  Job.batch "tempo-tokengen-job" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{"app.kubernetes.io/component":"tokengen-job", "app.kubernetes.io/instance":"tempo", "app.kubernetes.io/managed-by":"Helm", "app.kubernetes.io/name":"tempo", "app.kubernetes.io/version":"2.4.1", "batch.kubernetes.io/controller-uid":"1aef82a1-1901-4360-a92d-bfd2b8ac568c", "batch.kubernetes.io/job-name":"tempo-tokengen-job", "controller-uid":"1aef82a1-1901-4360-a92d-bfd2b8ac568c", "helm.sh/chart":"tempo-distributed-1.9.5", "job-name":"tempo-tokengen-job"}: `selector` does not match template `labels`, spec.selector: Invalid value: "null": field is immutable]
2024-05-03T15:07:55+10:00  batch         Job       tempo    tempo-tokengen-job   Running  SyncFailed    PostSync  job.batch/tempo-tokengen-job created
```

After a bit of investigation, the potential culprit has been found to be the `selector:` which is evaluating to a null value.

PR #3119, opened 2024-05-06
[tempo-distributed] Removing Job Selector Config

jordan-simonovski:
Many apologies! This is on me for not properly verifying the [Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/) prior to raising my previous PR.

Templating with a config like the following 

```
  selector:
  template:
    metadata:
      labels:
        app.kubernetes.io/component: tokengen-job-2
```

will throw the error: 

```
selector` does not match template `labels`, spec.selector: Invalid value: "null": field is immutable
```

On the other hand, adding in the selector labels will also fail as Kubernetes will throw the following error:

```
 `selector` not auto-generated
```

I've given this a much more rigorous test, and apologies for my previous bad change 🙏 

The Selector should not be manually specified unless otherwise configured in spec. Removing the selector altogether ensures the job configuration is valid, while also allowing for templating without null values causing errors.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.9.8
+version: 1.9.9
 appVersion: 2.4.1
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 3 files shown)


kvrhdn:
This is a revert of https://github.com/grafana/helm-charts/pull/3111 right?

kvrhdn (approved):

joe-elliott (approved):

zalegrala:
The chart version will need a bump and the docs need an update.

jordan-simonovski:
> This is a revert of #3111 right?

It's not so much a revert as much as it is just removing the whole selector key from the config.
We could revert and keep `selector:` but this still seems to trigger the issue when templating which throws `Invalid value: "null"`

I have verified that upon removing this from the template that Kube will still add the selector config with the auto-generated selector value.

PR #3120, opened 2024-05-06
Improve dashboard variable substitution

toanju:
The Grafana dashboard variable syntax allows varnames without curly braces ($datasource and ${datasource}). A substitution can only be applied for the latter. This enables extended regular expressions for sed to allow both substitutions to take place to include dashboards that are written in either form.


zanhsieh (approved):

toanju:
I am happy to update this once more, but I'd really appreciate if this time I could have a 2nd review. Many thanks @zanhsieh for already looking into this.

zanhsieh:
@zalegrala Can you review this plz?

zalegrala:
Have you had a chance to test this?  It seems like we should be matching the closing brace also.

toanju:
The matches were modified to match braces as optional. Hence, the closing brace will match if it is there, e.g `${DATASOURCE}` and `$DATASOURCE` will match. So far only `${DATASOURCE}` was matching.

zalegrala:
Looking at the old regex, I can test from the CLI with the following.
```bash
❯ echo '$variable' | sed '/-- .* --/! s/${{"{"}}{{ .name }}}/{{ .value }}/g'
$variable
```

With the new regex and the `-E` flag on sed, I get the following.
```bash
❯ echo '$variable' | sed -E '/-- .* --/! s/\$\{{"{"}}?{{ .name }}\}?/{{ .value }}/g'
sed: -e expression #1, char 54: Invalid content of \{\}
```

How can we validate this before merge?

toanju:
Please be aware that this is a helm chart and hence first processed by helm. Thus `{{"{"}}` will replaced with only `{` as in the original regex*. In addition `{{ .name }}` and `{{ .value }}` will be replaced according to the desired substitution.

*) in the new regex this could be omitted since there are no longer 3 `{` in a row, but I left it to ease comparison with the original regex

toanju:
Hi @zalegrala,
was the above description helpful or do you have any additional questions? In addition, I updated the branch once more.

jkroepke (commented):
My question is, would that sed also work on an alpine environment? It is not known to me, if sed E is a gnu extension?

If not, 2 simple sed commands are needed.

jkroepke:
> This could be mitigated by having to sed commands. The first as it was so far and the second without the curly braces but including the quotes.

I'm more a fan of practical solution over cool solutions. If there is already an potential use case, that could break and it can be avoided, then we go of this.

Pretty sure there is a 1 in 100 that a dashboard using ds as prefix for something.

charts/grafana/templates/_config.tpl
```diff
@@ -121,7 +121,7 @@ download_dashboards.sh: |
     {{- end }}
     {{- if kindIs "slice" $value.datasource }}
       {{- range $value.datasource }}
-        | sed '/-- .* --/! s/${{"{"}}{{ .name }}}/{{ .value }}/g' \
+        | sed -E '/-- .* --/! s/\$\{{"{"}}?{{ .name }}\}?/{{ .value }}/g' \
```
jkroepke, line 124:
```suggestion
        | {{`sed -E '/-- .* --/! s/\$\{?`}}{{ .name }}{{`\}?/`}}{{ .value }}/g' \
```

Is that more readable? No idea.

toanju: as mentioned above I'd go for `sed -E '/-- .* --/! s/\$\{?{{ .name }}\}?/{{ .value }}/g'`, I just left the original syntax to keep the changes minimal

PR #3136, opened 2024-05-17
[grafana] Update app version to v10.4.3

xvilo:
As 10.4.3 has been released some time ago https://github.com/grafana/grafana/releases/tag/v10.4.3

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 7.3.11
-appVersion: 10.4.1
+version: 7.3.12
+appVersion: 10.4.3
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
 home: https://grafana.com


terop:
Also version [11.0.0](https://github.com/grafana/grafana/releases/tag/v11.0.0) has been released a few days ago. Would it make sense to update directly to it?

xvilo:
Personally, I would definitely like to wait a while before updating to 11, so getting as much from 10.x as possible is preferred. Why not create a new release with this latest 10.x version anyway, I'm happy to create another one but I don't know if it needs additional changes in the chart, it's a major release after all.

terop:
I updated my own Grafana instance to 11.0.0 and it seems to work fine. However my dashboards are very simple and thus do not "test" the changes in 11.0.0 thoroughly.

xvilo:
Yeah, that is not the case for us. We have lots of dashboards and been using it since Grafana 5 or 6

jkroepke:
> Also version [11.0.0](https://github.com/grafana/grafana/releases/tag/v11.0.0) has been released a few days ago. Would it make sense to update directly to it?

#3139

zanhsieh (approved):

LarsStegman:
@gritzkoo @zalegrala since you recently this Chart, could you review this PR?

PR #3138, opened 2024-05-18
[grafana] Set `GOMEMLIMIT` environment variable based on container resources

jnoordsij:
Set ~~`GOMAXPROCS` and~~ `GOMEMLIMIT` environment variables based on container resources. This should reduce potential CPU throttling and OOMKills on containers.

The [`resourceFieldRef`](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-resourceFieldRef) is a very specific Kubernetes directive that is created specifically for passing resource-related values, which rounds up the CPU value to the nearest whole number (e.g. 250m to 1) and passes the memory as a numeric value; so `64Mi` would result in the environment variable being set to `67108864`. This by design makes it completely compatible with Go's API.

An example is documented within Kubernetes documentation itself: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-container-fields-as-values-for-environment-variables.

Inspired by https://github.com/traefik/traefik-helm-chart/pull/1029.

Implementation notes:
- This was only added to the "main" Grafana container, given it's only relevant for Go-based applications
- This idea can probably be applied to other charts within this repository; I currently have no intention of following-up with this myself, but anyone coming across this, feel free to do so!

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -1196,12 +1196,6 @@ containers:
         value: {{ (get .Values "grafana.ini").paths.plugins }}
       - name: GF_PATHS_PROVISIONING
         value: {{ (get .Values "grafana.ini").paths.provisioning }}
-      {{- if (.Values.resources.limits).cpu }}
-      - name: GOMAXPROCS
-        valueFrom:
-          resourceFieldRef:
-            resource: limits.cpu
-      {{- end }}
       {{- if (.Values.resources.limits).memory }}
       - name: GOMEMLIMIT
         valueFrom:
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -1196,6 +1196,18 @@ containers:
         value: {{ (get .Values "grafana.ini").paths.plugins }}
       - name: GF_PATHS_PROVISIONING
         value: {{ (get .Values "grafana.ini").paths.provisioning }}
+      {{- if (.Values.resources.limits).cpu }}
+      - name: GOMAXPROCS
+        valueFrom:
+          resourceFieldRef:
+            resource: limits.cpu
+      {{- end }}
+      {{- if (.Values.resources.limits).memory }}
+      - name: GOMEMLIMIT
+        valueFrom:
+          resourceFieldRef:
+            resource: limits.memory
+      {{- end }}
       {{- range $key, $value := .Values.envValueFrom }}
       - name: {{ $key | quote }}
         valueFrom:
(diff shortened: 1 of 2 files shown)


jkroepke:
I would not recommend to set `GOMAXPROCS` like that. It's counterproductive.

GOMAXPROCS sets the maximum number of CPUs that can be executing simultaneously

However, if you give grafana cpu=2 on a 16 core machine, then the Grafana pod will still get 16 cores assigned, but the pod is only able to use 1/8 of each CPU core.

With GOMAXPROCS, grafana may only use 2 cores of them and would not longer benefit form all assigned resources.

See also a 7 years old discussion that promethues-operator: https://github.com/prometheus-operator/prometheus-operator/issues/501

jnoordsij:
I'm all fine with leaving `GOMAXPROCS` out, even though I still think it's beneficial from the things I've seen/read. From what I've seen, as soon as you actually use Kubernetes limits, you will experience some degree of throttling in practice and by setting `GOMAXPROCS` accordingly this is somewhat mitigated. The https://github.com/uber-go/automaxprocs is a widely-used tool that does a very similar thing, which [was also added as opt-in to Prometheus operator](https://github.com/prometheus/prometheus/pull/10498).

However, I do agree all results and opinions I've seen are mixed and this is potentially not a better-for-everyone solution. If removing this part (or making it opt-in/opt-out rather than mandatory) is considered preferable, just let me know.

charts/grafana/templates/_pod.tpl
```diff
@@ -1196,6 +1196,18 @@ containers:
         value: {{ (get .Values "grafana.ini").paths.plugins }}
       - name: GF_PATHS_PROVISIONING
         value: {{ (get .Values "grafana.ini").paths.provisioning }}
+      {{- if (.Values.resources.limits).cpu }}
+      - name: GOMAXPROCS
+        valueFrom:
+          resourceFieldRef:
+            resource: limits.cpu
+      {{- end }}
```
jkroepke, line 1204: With looking forward on Go 1.25 and https://go.dev/blog/container-aware-gomaxprocs this automatic discover should be favorite.

It additionally adjust the GOMAXPROCS, if the cgroups limits has been changed.

```suggestion
```

jkroepke:
please bump the chart version, ty!

jkroepke (approved):

charts/grafana/templates/_pod.tpl
```diff
@@ -1196,6 +1196,12 @@ containers:
         value: {{ (get .Values "grafana.ini").paths.plugins }}
       - name: GF_PATHS_PROVISIONING
         value: {{ (get .Values "grafana.ini").paths.provisioning }}
+      {{- if (.Values.resources.limits).memory }}
```
jkroepke, line 1199: May ask what happens here, if a user set `.Values.resources=null` ?

PR #3145, opened 2024-05-27
[tempo-distributed] feat: add zone aware replication

### LINKED ISSUE #3069: [tempo-distributed] add support for zone aware replication in ingesters
i would like to be the same as mimir
because right now its not really possible when using the helm chart

KyriosGN0:
this will close #3069
this is mostly copied from the mimir chart implementation (only i put the functions in the ingester helper file, as i don't know if the metrics generator needs zone aware replication support)

i also removed the migration values as i don't know if the mimir approach is applicable here. 

@mapno @faustodavid @BitProcessor  @swartz-k, would like you review here as this a bigish addition to the chart
Thanks in advance


charts/tempo-distributed/values.yaml
```diff
@@ -224,6 +224,55 @@ ingester:
   appProtocol:
     # -- Set the optional grpc service protocol. Ex: "grpc", "http2" or "https"
     grpc: null
+  zoneAwareReplication:
+    # -- Enable zone-aware replication for ingester
+    enabled: true
```
zalegrala, line 229: I don't know what to expect if we go from default `false` to default `true` here.  Perhaps a default `false` makes sense here.

KyriosGN0: yeah, make sense

zalegrala:
I'll give some time to test this out, but I won't be able to get to it for a week.

lukasmrtvy:
@KyriosGN0 
Rendered chart does contain some errors for `tempo-ingester-*` statefulsets :
```
      affinity:
        Error: 'error converting YAML to JSON: yaml: line 6: mapping values are not allowed
          in this context'
```

charts/tempo-distributed/templates/ingester/_helpers-ingester.tpl
```diff
@@ -1,7 +1,180 @@
...
+{{- end -}}
+{{- end -}}
+{{/*
+Calculate anti-affinity for a zone
+Params:
+  rolloutZoneName = name of the rollout zone
+  topologyKey = topology key
+*/}}
+{{- define "ingester.zoneAntiAffinity" -}}
+{{- if .topologyKey -}}
```
lukasmrtvy, line 33: bad indentation. this works

```
{{- if .topologyKey -}}
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
        - labelSelector:
            matchExpressions:
              - key: rollout-group
                operator: In
                values:
                  - ingester
              - key: zone
                operator: NotIn
                values:
                  - {{ .rolloutZoneName }}
            topologyKey: {{ .topologyKey | quote }}
{{- else -}}
```

KyriosGN0: @lukasmrtvy fixed! thx for the review

KyriosGN0:
Hey @zalegrala , we're you able to test my PR? 
Sorry for nagging

zalegrala:
No problem, sorry for the delay.  I've had a chance to test this and take a closer look.  I think we need some documentation or clarity around how to adjust the replicas per zone.  Otherwise this looks to deploy and do the right thing.  Thanks for putting this together.

zalegrala:
I think a small example in the readme for this release would be good.  You can edit the template and then regenerate the readme.  Similar to [here](https://github.com/KyriosGN0/helm-charts/blob/6dc301d4c1fdddd943ca983480c1c68cda7f5f83/charts/tempo-distributed/README.md.gotmpl#L68), for example.

KyriosGN0:
added docs

charts/tempo-distributed/values.yaml
zalegrala: In my testing, I had to fix the ring after destroying the old instances, one remained unhealthy in the ring and had to forget it.  Once I dropped that member, everything looks pretty good.

zalegrala:
Looks good, though I see the default value true is back.  Is that intended?  Also you'll need to resolve the conflicting files for changes that went in before this one.

charts/tempo-distributed/values.yaml
KyriosGN0: its possibe we need to use the `-ingester.ring.unregister-on-shutdown`

KyriosGN0:
@zalegrala sorry i tested something and forgot to revert
fixed the conflicts

zalegrala (approved):

joe-elliott:
I'm good on this PR as long as this is clearly marked experimental.

KyriosGN0:
> I'm good on this PR as long as this is clearly marked experimental.

I'll update the docs and values file to mark this as experimental

KyriosGN0:
@joe-elliott updated docs to reflect that

joe-elliott (approved):
Thanks for the contribution!

charts/tempo-distributed/README.md
```diff
@@ -45,6 +46,11 @@ The command removes all the Kubernetes components associated with the chart and
 
 A major chart version change indicates that there is an incompatible breaking change needing manual actions.
 
+### from Chart versions < 1.11.0
```
z0rc, line 49: This should be `< 1.12.0`

z0rc:
Something is wrong with range logic when zone aware replication is disabled (default). I'm validating upgrade of tempo-distributed of my installation from 1.12.0 to 1.13.0 (btw changelog  is incorrect about version) and helm-diff says that existing ingester statefulset will be removed and nothing to be created to replace it.

z0rc:
Also rollout-operator should be installed only when zone aware replication is enabled. Right now it's installed by default.

charts/tempo-distributed/Chart.yaml
```diff
@@ -31,3 +31,8 @@ dependencies:
     version: 0.2.2
     repository: https://grafana.github.io/helm-charts
     condition: metaMonitoring.grafanaAgent.installOperator
+  - name: rollout-operator
+    alias: rollout_operator
+    repository: https://grafana.github.io/helm-charts
+    version: 0.15.0
+    condition: rollout_operator.enabled
```
z0rc, line 38: This variable isn't defined in values, which for some reason enabled installation of rollout-operator by default.

KyriosGN0:
@z0rc, opened a PR to fix this

Starefossen:
Yeah, this seams to have messed up the latest version of the tempo-distributed chart. Had to downgrade to v1.12.0 to get things rolling again.

PR #3148, opened 2024-05-28
feat(grafana): make possible to globally override a specific image registry

yann-soubeyrand:
This feature is necessary in cases where repositories coming from different registries are stored in different namespaces in your registry. For example, this is the case with AWS ECR pull through cache feature, which uses the repository prefix to know where to fetch the image from.

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -29,7 +29,7 @@ initContainers:
 {{- end }}
 {{- if ( and .Values.persistence.enabled .Values.initChownData.enabled ) }}
   - name: init-chown-data
-    {{- $registry := .Values.global.imageRegistry | default .Values.initChownData.image.registry -}}
+    {{- $registry := index .Values.global.imageRegistries .Values.initChownData.image.registry | default .Values.global.imageRegistry | default .Values.initChownData.image.registry -}}
     {{- if .Values.initChownData.image.sha }}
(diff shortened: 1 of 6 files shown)


zanhsieh (approved):

charts/grafana/templates/_pod.tpl
```diff
@@ -29,7 +29,7 @@ initContainers:
 {{- end }}
 {{- if ( and .Values.persistence.enabled .Values.initChownData.enabled ) }}
   - name: init-chown-data
-    {{- $registry := .Values.global.imageRegistry | default .Values.initChownData.image.registry -}}
+    {{- $registry := index .Values.global.imageRegistries .Values.initChownData.image.registry | default .Values.global.imageRegistry | default .Values.initChownData.image.registry -}}
```
jkroepke, line 32: Could you please create a named template for this, and reuse rhe logic?

charts/grafana/values.yaml
```diff
@@ -2,6 +2,11 @@ global:
   # -- Overrides the Docker registry globally for all images
   imageRegistry: null
 
+  # -- Overrides the Docker registries globally for all images based on their origin registry
+  imageRegistries: {}
```
jkroepke, line 6: Please configure this in one of the ci test values.

You can use kiwigrid/k8s-sidecar as test, since it is published on multiple registries.

jkroepke (changes requested):

PR #3160, opened 2024-06-08
[promtail] Fix YAML parsing error

allanhung:
Re-procedure this issue:
```
$ helm pull --untar grafana/promtail --version 6.16.0
$ helm template --set serviceMonitor.enabled=true promtail/
Error: YAML parse error on promtail/templates/service-metrics.yaml: error converting YAML to JSON: yaml: line 11: did not find expected key

Use --debug flag to render out invalid YAML
```

### DIFF
--- a/charts/promtail/Chart.yaml
+++ b/charts/promtail/Chart.yaml
@@ -3,7 +3,7 @@ name: promtail
 description: Promtail is an agent which ships the contents of local logs to a Loki instance
 type: application
 appVersion: 3.0.0
-version: 6.16.0
+version: 6.16.1
 home: https://grafana.com/loki
 sources:
   - https://github.com/grafana/loki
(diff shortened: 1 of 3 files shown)


davinkevin:
I bypassed it using this extra `values.yaml`

```
service:
  labels:
    uselessLabel: because-of-generation-issue
```

I subscribe to get notified when merged, so I can remove this hack.

thx

pschichtel:
I usually prefer `with` instead of `if` in these cases

allanhung:
> I usually prefer `with` instead of `if` in these cases

Thanks. I have updated the code to reflect this preference.

allanhung:
duplicate with [#3164](https://github.com/grafana/helm-charts/pull/3164).

PR #3164, opened 2024-06-09
[promtail] Update service template to fix broken promtail.

Sheikh-Abubaker:
Fixes #3162

PR #3175, opened 2024-06-17
Fix promtail helm charts service metrics template

xinity:
This PR fixes the service-metrics template regarding labels and annotations.


gelicia:
Same work done in #3164  - thanks so much for the catch @Sheikh-Abubaker :D

PR #3176, opened 2024-06-17
fix promtail service-metrics template

xinity:
the service-metrics template is broken.

This PR add the according fix to make it work again :)


xinity:
Closing as the fix is already merged from #3164 😂

PR #3194, opened 2024-06-27
tempo-vulture: fix yaml indentation error

devoxel:
Using the following values:

	tempoAddress:
	  push: "http://p"
	  query: "http://q"
	annotations:
	  foo: bar
	  baz: fop

helm template produced yaml that looked like:

	  annotations:    baz: fop
	    foo: bar

Thye reason was that it was using indent, instead of nindent.


devoxel:
This issue is entirely breaking the use of the `annotations` param, and fix is simple - anybody free to review?

Sheikh-Abubaker (commented):
@devoxel Could you please fix the CI lint test error.

PR #3200, opened 2024-07-01
[promtail] prefer image.tag for app.kubernetes.io/version

yurrriq:
For example, when I set `image.tag=2.9.8`, I'd like the resulting annotation to be `app.kubernetes.io/version: 2.9.8` instead of the current `app.kubernets.io/version: 2.9.3` which uses the `.Chart.AppVersion`. I've copied from the grafana chart, which already has the correct behaviour.

### DIFF
--- a/charts/promtail/values.yaml
+++ b/charts/promtail/values.yaml
@@ -100,7 +100,7 @@ image:
   # -- Docker image repository
   repository: grafana/promtail
   # -- Overrides the image tag whose default is the chart's appVersion
-  tag: null
+  tag: ""
   # -- Docker image pull policy
   pullPolicy: IfNotPresent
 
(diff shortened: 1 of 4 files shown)


cyriltovena (approved):
LGTM

yurrriq:
I've changed the default `image.tag` from `null` to `""` which solves the linting issue, at least locally.

charts/promtail/values.yaml
```diff
@@ -100,7 +100,7 @@ image:
   # -- Docker image repository
   repository: grafana/promtail
   # -- Overrides the image tag whose default is the chart's appVersion
-  tag: null
+  tag: ""
```
yurrriq, line 103: With `null`, the linter complains about a type mismatch: https://github.com/grafana/helm-charts/actions/runs/9773179026/job/26978893209?pr=3200

yurrriq:
And now I've updated README.md :facepalm: Hopefully this time the linter is finalyl happy.

trevorwhitney (approved):
LGTM

PR #2922, opened 2024-01-21
[tempo] Add support for Network Policy.

Sheikh-Abubaker:
Fixes #2890

PR #3203, opened 2024-07-02
[tempo] Remove multiple - (dash) entries under from within ingress

Sheikh-Abubaker:
This PR addresses the user's comment https://github.com/grafana/helm-charts/pull/2922#issuecomment-2188167417 to remove multiple - (dash) entries under from in ingress to perform AND operation making the Network Policy more secure.

### DIFF
--- a/charts/tempo/templates/networkpolicy.yaml
+++ b/charts/tempo/templates/networkpolicy.yaml
@@ -52,7 +52,7 @@ spec:
         - namespaceSelector:
             {{- toYaml . | nindent 12 }}
         {{- end }}
-        - podSelector:
+          podSelector:
             matchLabels:
               {{- include "tempo.labels" . | nindent 14 }}
               role: read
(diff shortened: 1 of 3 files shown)


charts/tempo/templates/networkpolicy.yaml
```diff
@@ -52,7 +52,7 @@ spec:
         - namespaceSelector:
             {{- toYaml . | nindent 12 }}
         {{- end }}
-        - podSelector:
+          podSelector:
```
zanhsieh, line 55: What would impact if above L51 explicitNamespacesSelector is empty? Would that break anything?

zanhsieh (changes requested):

charts/tempo/templates/networkpolicy.yaml
Sheikh-Abubaker: @zanhsieh if `explicitNamespacesSelector` is empty, it wouldn't break anything since the code goes with `{{ with }}` in line 51, I tested this Network Policy previously by keeping `explicitNamespacesSelector` empty only, it didn't break anything and was working as expected.

zanhsieh (approved):

zalegrala (approved):

Issue #3072, opened 2024-04-11
Loading dashboards as files results in duplicate warnings in Grafana logs


nwsteenberg:
Using grafana chart 7.3.0 i am loading in some dashboards as files:
```
dashboards:
    someProvider:
      home:
        file: dashboards/someProvider/home.json
```
But also to pick up configmaps with the following configuration:
```
sidecar:
    dashboards:
      folderAnnotation: grafana_dashboard_folder
      searchNamespace: ALL
      # Configmap Dashboard labels
      label: grafana_dashboard
      labelValue: "1"
```
I get a lot of the following warnings in grafana:
```
logger=provisioning.dashboard t=2024-04-11T08:04:33.170882773Z level=warn msg="dashboards provisioning provider has no database write permissions because of duplicates" provider=someProvider orgId=1
logger=provisioning.dashboard t=2024-04-11T08:04:33.170889673Z level=warn msg="dashboards provisioning provider has no database write permissions because of duplicates" provider=sidecarProvider orgId=1
logger=provisioning.dashboard t=2024-04-11T08:05:03.171217608Z level=warn msg="the same UID is used more than once" orgId=1 uid=pS6ZuGV7z times=2 providers="[someProvider sidecarProvider]"
logger=provisioning.dashboard t=2024-04-11T08:05:03.171309409Z level=warn msg="the same UID is used more than once" orgId=1 uid=WmDYPIXGz times=2 providers="[sidecarProvider someProvider]"
logger=provisioning.dashboard t=2024-04-11T08:05:03.171331109Z level=warn msg="dashboard title is not unique in folder" orgId=1 title="Kubernetes cluster overview" folderID=1 times=2 providers="[someProvider sidecarProvider]"
logger=provisioning.dashboard t=2024-04-11T08:05:03.171349109Z level=warn msg="dashboard title is not unique in folder" orgId=1 title="Home" folderID=1 times=2 providers="[someProvider sidecarProvider]"
logger=provisioning.dashboard t=2024-04-11T08:05:03.171362609Z level=warn msg="dashboards provisioning provider has no database write permissions because of duplicates" provider=someProvider orgId=1
logger=provisioning.dashboard t=2024-04-11T08:05:03.171370409Z level=warn msg="dashboards provisioning provider has no database write permissions because of duplicates" provider=sidecarProvider orgId=1
```

[...]

PR #3215, opened 2024-07-07
[grafana] Removed dashboard labels from dashboard-json-configmap

nwsteenberg:
Remove the dashboard labels from dashboard-json-configmaps, as these are loaded seperately. 
This fixes: #3072

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 8.3.3
+version: 8.3.4
 appVersion: 11.1.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,10 +1,6 @@
 apiVersion: v2
 name: grafana
-<<<<<<< HEAD
-version: 8.3.4
-=======
-version: 8.3.5
->>>>>>> upstream/main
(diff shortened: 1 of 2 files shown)


jkroepke:
Could you please explain, what did you mean by

> as these are loaded seperately

nwsteenberg:
If i define in my values:

```
dashboardProviders: 
  dashboardproviders.yaml:
    apiVersion: 1
    providers:
    - name: 'default'
      orgId: 1
      folder: ''
      type: file
      disableDeletion: false
      editable: true
      options:
        path: /var/lib/grafana/dashboards/default

dashboards: 
  default:
    some-dashboard:
      json: |
        <json dashboard>
```


My dashboard is loaded through the provider default. This works as intented. My dashboard exists in grafana upon installation. 

However. If i add the following:
```
sidecar:
  dashboards:
    enabled: true
    label: grafana_dashboard
    labelValue: "1"
```

My dashboard will be loaded through the provider, created as a configmap and labled so it can be picked up by the sidecar. 
So the dashboard is loaded in twice. Once by the provider configuration, and then again by the sidecar. That i what i meant with "as these are loaded seperately (by the provider)"

Hope it makes sense

zanhsieh (approved):

jkroepke (approved):

mkwanqsr:
I need this change reversed as it doesn't cater for sidecar dashboard providers.

If the dashboardProviders is a side car (not default), then my custom dashboard doesn't get picked up, because the grafana_dashbaord: "1" label is not getting stamped as the previous change was trying to do.

PR #3227, opened 2024-07-12
[grafana] Fixed image renderer network policy namespace selector

elliotcourant:
The egress network policy for the image rendered expects the namespace of the chart to match `name: ${value}`, this label is not applied to namespaces by default in Kubernetes and would need to be added to the namespace in order for this network policy to work. However the ingress network policy uses `kubernetes.io/metadata.name: ${value}`, this label is applied to namespaces in kubernetes by default and will allow the network policy to work out of the box with reasonable defaults.

Without this patch, essentially grafana is able to communicate with the image renderer, but the image renderer is not able to make HTTP requests back to grafana in the same namespace without adding `name: ${value}` to the labels of the namespace grafana is deployed in.


jkroepke (changes requested):
Looks good, please bump the chart version.

elliotcourant:
> Looks good, please bump the chart version.

Done! sorry about that!

jkroepke (approved):

Sheikh-Abubaker (approved):
LGTM!

PR #3230, opened 2024-07-14
[tempo-distributed] feat: add cli flag for zone aware ingestion

KyriosGN0:
like in mimir, we need to pass `ingester.ring.instance-availability-zone` when enabled zone aware ingestion 
FYI @zalegrala

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.19.1
+version: 1.20.1
 appVersion: 2.6.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.15.2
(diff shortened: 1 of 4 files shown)


zalegrala:
I get the following when trying to pass that flag.
```
flag provided but not defined: -ingester.ring.instance-availability-zone
```

KyriosGN0:
@zalegrala weird, in our mimir we use this flag, but according to the Cortex Config it should be `-ingester.availability-zone`
i'll try to compile in jsonnet since it also has a Zone Aware config

KyriosGN0:
hm @zalegrala i looked into the jsonnet and they only set the env var `availability_zone` which corresponds to the `-ingester.availability-zone` cli flag according to the cortex docs, so i think its good enough.

zalegrala:
Each database exposes different config flags, so we can't assume that because its exposed in mimir that it would be exposed in tempo.  Have you had a chance to test this change?

KyriosGN0:
@zalegrala `-ingester.availability-zone` is equivalent to the env var set in the jsonnet file of Tempo, i'll try to deploy this locally in some KIND cluster to see that it recognizes the flag.

KyriosGN0:
@zalegrala after taking another look in the jsonnet file i've realized i've missed other things in the chart (like services) 
i would also need to open PR to tempo itself to support this cli flag (or figure out how to implement this with the env va)
will ping you when im ready, until then i've converted this to a Draft pr

KyriosGN0:
@zalegrala should be good now.

zalegrala (approved):

zanhsieh (approved):

PR #3244, opened 2024-07-26
[grafana] fix: do not create pvc manually when useStatefulSet

fengxsong:
when deploying grafana as stateful workload, it's unnecessary to create pvc manually since we use `statefulset.spec.volumeClaimTemplates` instead.


jkroepke:
please fix the CI

jkroepke (approved):

zanhsieh (approved):

PR #3250, opened 2024-07-30
[grafana] add value to make extraConfigmapMounts and extraSecretMounts optional

tibuntu:
Hi there, we are using that feature a lot but recently ran into some kind of race condition because a referenced ConfigMap wasn't existing yet. I noticed that there is now also an alerting sidecar which will probably improve our situation, but I still wanted to contribute this small changes which gives one the option to make the extraConfigmapMounts optional.

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -1156,6 +1156,9 @@ volumes:
   - name: {{ tpl .name $root }}
     configMap:
       name: {{ tpl .configMap $root }}
+      {{- with .optional }}
+      optional: {{ . }}
+      {{- end }}
       {{- with .items }}
       items:
         {{- toYaml . | nindent 8 }}
@@ -1261,6 +1264,9 @@ volumes:
     secret:
       secretName: {{ .secretName }}
       defaultMode: {{ .defaultMode }}
+      {{- with .optional }}
+      optional: {{ . }}
+      {{- end }}
       {{- with .items }}
       items:
         {{- toYaml . | nindent 8 }}
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 8.3.7
+version: 8.4.0
 appVersion: 11.1.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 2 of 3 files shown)


jkroepke:
Hi, could you add this feature to extraSecretMounts as well?

tibuntu:
> Hi, could you add this feature to extraSecretMounts as well?

That makes sense, done!

tibuntu:
Just noticed that I used different default values for the examples. What would you prefer?

charts/grafana/templates/_pod.tpl
```diff
@@ -1156,6 +1156,7 @@ volumes:
   - name: {{ tpl .name $root }}
     configMap:
       name: {{ tpl .configMap $root }}
+      optional: {{ .optional }}
```
jkroepke, line 1159:
```suggestion
      {{- with .optional }}
      optional: {{ .optional }}
      {{- end }}
```

charts/grafana/templates/_pod.tpl
```diff
@@ -1261,6 +1262,7 @@ volumes:
     secret:
       secretName: {{ .secretName }}
       defaultMode: {{ .defaultMode }}
+      optional: {{ .optional }}
```
jkroepke, line 1265:
```suggestion
      {{- with .optional }}
      optional: {{ .optional }}
      {{- end }}
```

jkroepke:
optional: false is the Kubernetes default, lets use that.

I added a suggestion, if optional is not present at values, the key will be omit in the template, too.

jkroepke (approved):

zanhsieh (approved):

makkes:
If the ConfigMap doesn't exist at the time the Pod is created, it won't be mounted into the container. Kubernetes itself doesn't restart the container as soon as the CM is created so what is the expected behaviour here? Should users delete the Pod after the CM is created? I suppose this can lead to a lot of confusion.

jkroepke:
> Should users delete the Pod after the CM is created? 

Yes, that expected from Kubernetes point of view. 

> I suppose this can lead to a lot of confusion.

Yes, avoid using optional flag would be the best possible option here.

Not sure, if you can use https://github.com/stakater/Reloader to automate the restart.


----

The general recommendation from an Kuberneter guy with much experience: Avoid optional unless you know the consequences.

makkes:
> The general recommendation from me with much experience: Avoid optional unless you know the consequences.

I agree and that's what my comment was all about that the behaviour is documented in the default `values.yaml` file accompanying this chart so that users don't expect the chart to do any magic around this but will rather just pass the flag to the mount. Not sure if this is a usual practice in charts, though.

PR #3251, opened 2024-07-30
[grafana] Update Helm test annotations to new format

ibakshay:
This PR updates all deprecated Helm test annotations in `grafana` charts to the new format as per the official Helm documentation.

#### Changes:

Replace `"helm.sh/hook": test-success` with `"helm.sh/hook": test`

#### Reference:
https://helm.sh/docs/topics/chart_tests/

Testing:
- [ ]  Verified that all tests still run correctly with the new annotation


jkroepke:
@zanhsieh it seems like this Repo has [CLAassistant](https://github.com/CLAassistant) AND DCO enabled. Do we really need both?

jkroepke (approved):

PR #3259, opened 2024-08-05
feat: add config for enable_virtual_node_label

davhdavh:
Add configuration item to enable the metrics generators virtual nodes to infer services call but not instrumented.

If set to false, it uses the default (which is false)

https://grafana.com/docs/tempo/latest/metrics-generator/service_graphs/#activate-enable_virtual_node_label/


Sheikh-Abubaker (commented):
@davhdavh Could you please sign the DCO and fix the CI lint errors ?

Issue #3264, opened 2024-08-07
[tempo-distributed] Ingesters are not configured with the correct availability_zone when zone-awareness is enabled.


sringel:
When zone-awareness is enabled the `availability_zone` for each ingester remains the default value of `""`.

This was solved in [jsonnet](https://github.com/grafana/tempo/blob/main/operations/jsonnet/microservices/multi-zone.libsonnet) by updating the config to use environment variable expansion.

Something similar will need to be done for helm.

sringel:
fixed in https://github.com/grafana/helm-charts/pull/3230

PR #3265, opened 2024-08-07
[tempo-distributed] Set ingester availability_zone when zoneAwareReplication is enabled.

sringel:
This PR updates the config template and statefulset to ensure the `availability_zone` is set for the ingester when zoneAwareReplication is enabled.

This resolves issue: https://github.com/grafana/helm-charts/issues/3264


sringel:
fixed in https://github.com/grafana/helm-charts/pull/3230

Issue #3268, opened 2024-08-10
[grafana] [image-renderer] Service monitor because the ingress network policy does not allow prometheus to connect


FirelightFlagboy:
Hello, I had configured the `grafana-image-renderer` with the `grafana/grafana` helm with the following configuration:

```yaml
imageRenderer:
    enabled: true

    serviceMonitor:
        enabled: true
```

But the service monitor wasn't able to fetch the metrics as it wasn't able to connect to the `grafana-image-renderer` pod with the TPC error `connection refused`.

It appears that the issue is caused by the configured `image-renderer-ingress` network policy that only allow connection from the `grafana` pod.

What I would expect is that when the service monitor is enabled that the ingress is configured to allow prometheus to scrap the metrics from the `image-renderer` pod.

## Workaround

Currently, I have a workaround that use the `extraIngressSelectors` to select the prometheus pod.

```yaml
imageRenderer:
    networkPolicy:
        extraIngressSelectors:
            - namespaceSelector:
                matchLabels:
                    kubernetes.io/metadata.name: siem
              podSelector:
                matchLabels:
                    app.kubernetes.io/instance: prometheus
                    app.kubernetes.io/name: prometheus
```

PR #3272, opened 2024-08-15
Fix network policy always applying to ingress

nomaster:
The if statement does not target the 'enabled' property, but the whole map of attributes. This leads to the network policy always listing the 'Ingress' policy type, which disrupts legitimate traffic.

fixes: https://github.com/grafana/helm-charts/issues/3268


nomaster:
Just found that the Helm chart values define

- networkPolicy.enabled (default false)
- networkPolicy.ingress (default true)
- networkPolicy.egress.enabled (default false)

... which is quite confusing to me.

This means that the behavior is intended and this PR can be closed.

PR #3282, opened 2024-08-21
[grafana] Update app version to v10.4.4

abubro77:
[grafana] Update app version to v10.4.4

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana
-version: 7.3.12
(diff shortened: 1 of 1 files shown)


jkroepke:
We are already on v11

abubro77:
Hi @jkroepke 

We don't have helm chart for Grafana image version 10.4.4, so raise the PR for releasing the new helm chart.
Correct me, If i am wrong.

Thanks,
Abu

jkroepke:
Hi, you can always manually override the version inside the values. The grafana helm chart always holds the latest version of grafana.

PR #3283, opened 2024-08-21
[grafana] Update app version to v10.4.5

abubro77:
[grafana] Update app version to v10.4.5


jkroepke:
We are already on v11

PR #3284, opened 2024-08-21
[grafana] Update app version to v10.4.6

abubro77:
[grafana] Update app version to v10.4.6


jkroepke:
We are already on v11

PR #3309, opened 2024-09-12
deprecating that chart the way helm best practices suggest it

### LINKED ISSUE #3300: Is loki-stack deprecated?
I wonder if loki-stack is depreacated or lacks maintenance?
Why I'm asking;
When deprecated it would be nice to add `deprecated: true` to the Chart.yaml
If it lacks maintenance we should just update it.

erkules:
Fixes #3300 
Helm suggest to change the Chart.yaml to announce deprecation of a HelmChart.
Tools like renovate, nova ... use that information.


zanhsieh (changes requested):
@erkules I think better to leave some significant words about deprecation in README.md.

PR #3324, opened 2024-09-24
[grafana-sampling] add batch config options and k8s.pod.name to metrics dimensions

rlankfo:
This PR makes batch processing options configurable for the deployment and statefulset batch processors. Additionally, k8s.pod.name is added to the set of default dimensions to avoid issue with faulty target_info metric generation.

### DIFF
--- a/charts/grafana-sampling/values.yaml
+++ b/charts/grafana-sampling/values.yaml
@@ -9,6 +9,7 @@ metricsGeneration:
     - service.version
     - deployment.environment
     - k8s.cluster.name
+    - k8s.pod.name
 
 sampling:
   # -- Toggle tail sampling.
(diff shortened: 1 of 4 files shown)


petewall (approved):

charts/grafana-sampling/values.yaml
```diff
@@ -9,6 +9,7 @@ metricsGeneration:
     - service.version
     - deployment.environment
     - k8s.cluster.name
+    - k8s.pod.name
```
mar4uk, line 12: should this list be the same as we use for [tempo metrics gen](https://github.com/grafana/app-observability-plugin/blob/fba1b1fe8fbc93e6285d0bc522a71b470986ab6f/plugin/src/constants/overrides.ts#L5-L11) and use in [app o11y docs](https://github.com/grafana/app-o11y/blob/1a09c19409b7c0cd74a4ca61f22141821c0e59b4/docs/sources/sampling/assets/otel-config-head-sampling.yaml#L92-L99)?

    dimensions:
      - service.namespace
      - service.version
      - deployment.environment
      - k8s.cluster.name
      - k8s.namespace.name
      - cloud.region
      - cloud.availability_zone

I remember we discussed `k8s.pod.name` as a dimension, but decided not to add it because of high cardinality.
Here is [DesignDoc](https://docs.google.com/document/d/1mnmPZnU60Bp87uM3qoeaxyDCSh1iqOO5hoF50MP9cgY/edit) about list of resource attributes we decided to add to generated metrics

PR #3352, opened 2024-10-10
[grafana] correct array formatting for grafana.ini

nathwill:
fixes array format in configmap grafana.ini.

previously produced: `org_mapping = ["a" "b" "c"]`, which grafana did not apply correctly
now produces: `org_mapping = ["a", "b", "c"]`, which grafana applies correctly

of note:
for some reason this only seems to work when applied as TOML style array lists `["a", "b", "c"]`, even though the file type is nominally ini, which would suggest the following would actually be appropriate:
```
org_mapping[] = "a"
org_mapping[] = "b"
org_mapping[] = "c"
```

🤷

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 8.5.2
+version: 8.5.3
 appVersion: 11.2.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_config.tpl
+++ b/charts/grafana/templates/_config.tpl
@@ -13,6 +13,8 @@ grafana.ini: |
   {{- if not (kindIs "map" $elemVal) }}
   {{- if kindIs "invalid" $elemVal }}
   {{ $elem }} =
+  {{- else if kindIs "slice" $elemVal }}
+  {{ $elem }} = {{ toJson $elemVal }}
   {{- else if kindIs "string" $elemVal }}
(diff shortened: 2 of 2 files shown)


jkroepke:
Could you link documentation, where array style is used?

for org_mapping, I can only find a space separated string:

```
org_mapping = org_foo:org_foo:Viewer org_bar:org_bar:Editor *:org_baz:Editor
```

jkroepke:
but with the join method, it will result into:

```
org_mapping = [a,b,c]
```

not

```
org_mapping = ["a","b","c"]
```

nathwill:
yeah, string-wrapping is optional as it's written and afaik only required when spaces are involved (?)... if we want to enforce string-wrapping i can update the PR accordingly, just wanted to only do the minimal needed so it could be made to work via values...

in our patched version of the chart we're passing values in like this:

```
    org_mapping:
    - '"SRE:1:Admin"'
    - '"Support:1:Editor"'
    - '"Data Scientists:1:Editor"'
    - '"Developers:1:Editor"'
```

which produces 

```
org_mapping = ["SRE:1:Admin", "Support:1:Editor", "Data Scientists:1:Editor", "Developers:1:Editor"]
```

jkroepke:
instead `join`, consider using [`toJson`](https://helm.sh/docs/chart_template_guide/function_list/#tojson-musttojson)

nathwill:
cool, i like it.

jkroepke (commented):
For conistancy reasons, could you has this to the config section (15 lines above) as well?

jkroepke (approved):

PR #3355, opened 2024-10-11
[grafana] add extraMounts to plugins sidecar

### LINKED ISSUE #3351: [Grafana] Adding extraVolumeMount to sidecar to run custom script
Problem:
The grafana chart has configuration to have sidecars and even a `script` env variable to set the path to a custom script. For the moment, there is no `extraVolumes` and `extraVolumeMounts` to use to mount a script in the sidecar container. The only additional volumes that can be add are for the main Grafana container. Maybe is there another way to run a custom script that I am not aware of. The only other way I thought of is to modify the sidecar image to already have my script but that seems like more work than needed.

Objective:
I would like to mount a custom script to mount to a sidecar and run it everytime plugins are reloaded like described [here](https://github.com/grafana/helm-charts/blob/350783e92680eb98ff0c2b75eb0502a2dcc79fcb/charts/grafana/values.yaml#L1053).

Solution:
Add `extraVolumes` and `extraVolumeMounts` to the sidecar options.

coutug:
Fixes #3351

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -857,6 +857,9 @@ containers:
     volumeMounts:
       - name: sc-plugins-volume
         mountPath: "/etc/grafana/provisioning/plugins"
+      {{- with .Values.sidecar.plugins.extraMounts }}
+      {{- toYaml . | trim | nindent 6 }}
+      {{- end }}
 {{- end}}
   - name: {{ .Chart.Name }}
     {{- $registry := .Values.global.imageRegistry | default .Values.image.registry -}}
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -649,6 +649,9 @@ containers:
     volumeMounts:
       - name: sc-datasources-volume
         mountPath: "/etc/grafana/provisioning/datasources"
+      {{- with .Values.sidecar.datasources.extraMounts }}
+      {{- toYaml . | trim | nindent 6 }}
+      {{- end }}
 {{- end}}
 {{- if .Values.sidecar.notifiers.enabled }}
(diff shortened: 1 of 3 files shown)


charts/grafana/templates/_pod.tpl
```diff
@@ -857,6 +857,9 @@ containers:
     volumeMounts:
       - name: sc-plugins-volume
         mountPath: "/etc/grafana/provisioning/plugins"
+      {{- with .Values.sidecar.plugins.extraMounts }}
+      {{- toYaml . | trim | nindent 6 }}
```
jkroepke, line 861: why trim?

jkroepke (commented):
Please bump the chart version, thanks!

charts/grafana/templates/_pod.tpl
coutug: not really required, I just left it the same as other extraMounts (like the sc-alerts one)

jkroepke (approved):

charts/grafana/templates/_pod.tpl
```diff
@@ -857,6 +857,9 @@ containers:
     volumeMounts:
       - name: sc-plugins-volume
         mountPath: "/etc/grafana/provisioning/plugins"
+      {{- with .Values.sidecar.plugins.extraMounts }}
```
jkroepke, line 860: please add this to all sidecars

coutug: Sure, do you want me to remove the trim function from existing extraMount or add it to the new extraMount for the sake of consistency?

jkroepke: Yes, add it for the sake of consistency.

jkroepke (approved):

zanhsieh (approved):

PR #3364, opened 2024-10-15
[tempo-distributed] Fixes minor regression in tempo-distributed compactor extraEnv

mtweten:
A recent change to add a global extraEnv (https://github.com/grafana/helm-charts/commit/2126cb60a0974de3369ad210ebce8be786d4ccaa) made a small typo for the compactor component, so that it was looking at `compactor.env` instead of `compactor.extraEnv` (in the case `global.extraEnv` is not set). 

This PR just fixes that.


mapno (approved):

Sheikh-Abubaker (approved):
LGTM!

PR #3115, opened 2024-05-03
Publish chart package to ghcr from the update-helm-repo workflow

narqo:
Fixes #3068 grafana/mimir#7628

This one is similar to #2443 #2998 but updates the helper workflow, Mimir (and maybe others) use for their helm charts.

PR #3366, opened 2024-10-16
Add OCI registry support for all Helm charts

domolitom:
Some Helm charts, such as Loki, are not being pushed to an OCI registry because OCI support is missing in the `update-helm-repo.yaml` workflow, which is used for external releases.

Solution: This PR introduces a step that ensures Helm charts are also pushed to the GitHub Container Registry (GHCR) as part of the release process, providing OCI support for charts that were previously excluded.

### DIFF
--- a/.github/workflows/update-helm-repo.yaml
+++ b/.github/workflows/update-helm-repo.yaml
@@ -91,8 +91,10 @@ jobs:
           fi
 
   release:
-    needs: [ setup ]
+    needs: [setup]
     runs-on: ubuntu-latest
+    permissions:
+      packages: write
     env:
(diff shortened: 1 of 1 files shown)


domolitom:
@Sheikh-Abubaker

domolitom:
@narqo that is true. I didn't see the already open PR. Btw the PR https://github.com/grafana/helm-charts/pull/3115 needs to be edited, the workflow needs write permission to packages. I just added it into this PR.

narqo (approved):
🔥 Thank you, works for me (_now we need to find one more +1_).

zalegrala (approved):

PR #3374, opened 2024-10-23
[grafana] fix: allow nameOverride for PVC

anthr76:
I have a usecase where using the volumeClaimTemplate `.metadata.name` as the `{{ include "grafana.fullname" . }}` would aid in flexibility. 

Users currently using a statefulSet would need to set `.Values.persistence.nameOverride` to `storage` to avoid a breaking change.

### DIFF
--- a/charts/grafana/templates/statefulset.yaml
+++ b/charts/grafana/templates/statefulset.yaml
@@ -42,7 +42,7 @@ spec:
   - apiVersion: v1
     kind: PersistentVolumeClaim
     metadata:
-      name: {{ include "grafana.fullname" . }}
+      name: {{ if .Values.persistence.nameOverride }}{{ .Values.persistence.nameOverride }}{{ else }}{{ include "grafana.fullname" . }}{{ end }}
     spec:
       accessModes: {{ .Values.persistence.accessModes }}
       storageClassName: {{ .Values.persistence.storageClassName }}
--- a/charts/grafana/templates/statefulset.yaml
+++ b/charts/grafana/templates/statefulset.yaml
@@ -42,7 +42,7 @@ spec:
   - apiVersion: v1
     kind: PersistentVolumeClaim
     metadata:
-      name: storage
+      name: {{ include "grafana.fullname" . }}
     spec:
       accessModes: {{ .Values.persistence.accessModes }}
       storageClassName: {{ .Values.persistence.storageClassName }}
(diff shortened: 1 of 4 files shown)


onedr0p (approved):

charts/grafana/templates/statefulset.yaml
```diff
@@ -42,7 +42,7 @@ spec:
   - apiVersion: v1
     kind: PersistentVolumeClaim
     metadata:
-      name: storage
+      name: {{ include "grafana.fullname" . }}
```
jkroepke, line 45:
```suggestion
      name: {{ .Values.persistence.storageName }}
```

and to it to storage by default to avoid a breaking change

anthr76: @jkroepke while you were reviewing I actually revised this in a similar approach.

For backwards compatibility users can set `.Values.persistence.nameOverride` to `storage` while using statefulSets.

Let me know if this works for you!

jkroepke: could you explain the general benefit, why it should be configurable? It's just the name of template which will be merged with the pod name above. And the pod name already contains `{{ include "grafana.fullname" . }}`

anthr76: I have a use-case where I would prefer minimal intervention converting a deployment based helm chart to a statefulSet based helm chart installation on upgrade. Having the PVC name remain the same between a deployment mount and a statefulSet `volumeClaimTemplate` would be useful as Kubernetes will elect to use the existing one since the name remains the same. 

Having the override configurable aids in backwards compatibility

jkroepke: but keep in mind, there name here is not the name of the final PVC. Your use-case and the proposed change doesn't match

anthr76: I see I would like the name of the PVC to be this https://github.com/grafana/helm-charts/blob/563977e617b8f6861fadfe567c9dd399a4c8c054/charts/grafana/templates/pvc.yaml#L5C8-L5C43 . Perhaps I'm mis understanding something?

jkroepke: This mention name here describes a Volume Claim **Template** which is part of the statefulset. The statefulset controller will use the template an create own PVC for each pod. The final PVC name is the name of the PVC followed by the full name of the pod itself, and always ending with an ordered number.

zanhsieh (changes requested):
@anthr76 
Can you fix DCO please?

jkroepke:
@zanhsieh I feel the PR is pointless. Since Requester miss-understand the volumeTemplate naming.

zanhsieh:
@jkroepke 
Let's just close this.

PR #3379, opened 2024-10-29
[tempo] Add commonLabels (#3319) 

ramon951:
Every object created by the chart can have custom labels.

Fixes https://github.com/grafana/helm-charts/issues/3319

I edited Tempo Charts using the same logic used for https://github.com/grafana/mimir/issues/9067

I didn't edit dependency charts; I'm waiting for a feedback on how to proceed https://github.com/grafana/mimir/issues/9067#issuecomment-2439073840

### DIFF
--- a/charts/tempo/values.yaml
+++ b/charts/tempo/values.yaml
@@ -1,3 +1,7 @@
+global:
+  # -- Common labels for all object directly managed by this chart.
+  commonLabels: {}
+
 # -- Overrides the chart's name
 nameOverride: ""
 
@@ -282,6 +286,7 @@ podAnnotations: {}
 podLabels: {}
 
 # Apply extra labels to common labels.
+# DEPRECATED: use global.commonLabels
 extraLabels: {}
 
 # -- Volumes to add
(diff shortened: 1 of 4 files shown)


charts/tempo/values.yaml
```diff
@@ -282,6 +286,7 @@ podAnnotations: {}
 podLabels: {}
 
 # Apply extra labels to common labels.
+# DEPRECATED: use global.commonLabels
```
Sheikh-Abubaker, line 289:
```suggestion

```
@ramon951 folks could still make use of extraLabels if there's a scenario where the labels might not needed to be applied to all components ? so I'd be more inclined towards keeping extraLabels in place yet enhancing the chart with the support of commonLabels which takes care of the scenario where the same labels are needed to be applied to all components.

ramon951: I was wrong. I'm fixing it

Sheikh-Abubaker (approved):
LGTM!

zanhsieh (approved):

PR #3388, opened 2024-10-30
[grafana] Make the Helm hook type for the testFramework configurable

baileymjensen:
## Story
As a user of the testFramework, I'd like to be able to run it automatically `post-install` and `post-upgrade` to check the health endpoint of my Grafana instance any time it is altered. Additionally, upon failure of this test, I'd like it to rollback which happens automatically when using the previously mentioned hook types.


jkroepke (approved):
LGTM 👍

petewall (approved):

PR #3389, opened 2024-10-30
`managed-by` is set at install time by the managing installer, and should not be templated.

bleggett:
`app.kubernetes.io/managed-by: Helm` is automatically applied to resources _installed_ by Helm via `helm install`, and does not need to be (read: should not be) manually templated. The label is an install-time label that should be applied by the tool actually installing the resources into the cluster.

This matters because `helm template` should (and by default will) produce YAML without this label, as it is unknown at template-time what will "manage" the resource - e.g. `helm template | kubectl apply -f -` should not produce resources with the `app.kubernetes.io/managed-by: Helm` label, because the resources are not managed by helm.

xref: https://github.com/istio/istio/issues/53698

### DIFF
--- a/charts/agent-operator/templates/_helpers.tpl
+++ b/charts/agent-operator/templates/_helpers.tpl
@@ -35,7 +35,6 @@ Common labels
 */}}
 {{- define "ga-operator.labels" -}}
 {{ include "ga-operator.selectorLabels" . }}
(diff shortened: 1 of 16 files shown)


jkroepke:
Hi @bleggett 

thanks for you request. Please create a dedicated PR for each chart and increase the version separately.

PR #3392, opened 2024-11-01
[synthetic-monitoring-agent] Add deployment strategy

BlackDex:
Added a deploymentStrategy for the synthetic-monitoring-agent helm.

This allows more controle on how the pods will be deployed.
Used the same syntax as other helm charts already do.

### DIFF
--- a/charts/synthetic-monitoring-agent/templates/deployment.yaml
+++ b/charts/synthetic-monitoring-agent/templates/deployment.yaml
@@ -5,6 +5,10 @@ metadata:
   labels:
     {{- include "synthetic-monitoring-agent.labels" . | nindent 4 }}
 spec:
+  {{- with .Values.deploymentStrategy }}
+  strategy:
+    {{- toYaml . | trim | nindent 4 }}
+  {{- end }}
   minReadySeconds: 10
   {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
--- a/charts/synthetic-monitoring-agent/templates/deployment.yaml
+++ b/charts/synthetic-monitoring-agent/templates/deployment.yaml
@@ -7,7 +7,7 @@ metadata:
 spec:
   {{- with .Values.deploymentStrategy }}
   strategy:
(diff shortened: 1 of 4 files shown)


charts/synthetic-monitoring-agent/templates/deployment.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   labels:
     {{- include "synthetic-monitoring-agent.labels" . | nindent 4 }}
 spec:
+  {{- with .Values.deploymentStrategy }}
+  strategy:
+    {{- toYaml . | trim | nindent 4 }}
```
Sheikh-Abubaker, line 10:
```suggestion
    {{- toYaml . | nindent 4 }}
```

BlackDex: Any reason to remove it? 
All other helm charts seen to have this if I'm correct.

Sheikh-Abubaker: could you tell me any reason to include it ? any trailing whitespaces would already be trimmed.

BlackDex: As mentioned, i copied it from other helm charts in this repo which all use the same.

Also, the `{{-` will not trim the input from value it self if I'm correct.

Sheikh-Abubaker: Yeah but the CI lint test won't pass if there are trailing spaces in the `deployment strategy` value, so you don't have to worry about that.

BlackDex: Ok, I'm fine with removing it, i just wanted it to match the rest of all the helm charts.

BlackDex: Done

zanhsieh (approved):

Sheikh-Abubaker (approved):
LGTM!

PR #3404, opened 2024-11-06
[tempo] Update to 2.6.1

farcaller:
The chart seems to lag behind the releases now. I've manually bumped the image and it seems to work just fine with the rest of the chart.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 1.12.0
-appVersion: 2.5.0
+version: 1.13.0
+appVersion: 2.6.1
 engine: gotpl
 home: https://grafana.net
 icon: https://raw.githubusercontent.com/grafana/tempo/master/docs/tempo/website/logo_and_name.png
--- a/charts/tempo/README.md
+++ b/charts/tempo/README.md
@@ -1,6 +1,6 @@
 # tempo
 
-![Version: 1.12.0](https://img.shields.io/badge/Version-1.12.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.5.0](https://img.shields.io/badge/AppVersion-2.5.0-informational?style=flat-square)
+![Version: 1.13.0](https://img.shields.io/badge/Version-1.13.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 2.6.1](https://img.shields.io/badge/AppVersion-2.6.1-informational?style=flat-square)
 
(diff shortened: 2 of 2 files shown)


charts/tempo/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
 version: 1.11.0
```
zanhsieh, line 5:
```suggestion
version: 1.11.1
```

zanhsieh (changes requested):

charts/tempo/Chart.yaml
```diff
@@ -2,8 +2,8 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 1.11.0
-appVersion: 2.5.0
+version: 1.11.1
```
Sheikh-Abubaker, line 5:
```suggestion
version: 1.12.1
```
@zanhsieh how about we bump it to `1.12.1` since the  `app version: appVersion: 2.6.1` introduces minor as well as patch version change ?

zanhsieh: Sure.

zanhsieh (approved):

gcaracuel:
Same change to be applied on tempo-distributed if not that will also 'lag'

farcaller:
Sorry about trampling on your commits. I updated all three charts and bumped their minor versions as appropriate.

Sheikh-Abubaker:
@farcaller Please note you need to create individual PRs for different charts, otherwise the tests won't pass, you could only introduce changes in one chart in one PR.

farcaller:
@Sheikh-Abubaker I think that should do it.

farcaller:
Oh, it's one _PR_ per chart, not one _commit_ per chart.

farcaller:
Moved the other charts into #3412 and #3413.

Sheikh-Abubaker:
@farcaller you need to bump the patch version also i.e: `version: 1.13.1` since the `appVersion: 2.6.1` bumps both minor as well as patch version.

farcaller:
@Sheikh-Abubaker your [CONTRIBUTING.md](https://github.com/grafana/helm-charts/blob/main/CONTRIBUTING.md) points to [best practices](https://helm.sh/docs/chart_best_practices/conventions/) which says to use SemVer for versioning. I'd agree with you if there was a bump to 2.6.0 first and 2.6.1 second, but as we skip 2.6.0 completely (and there's no real reason to add it at this point), the semver approach is to bump the minor version.

Are you sure that you want to have `1.13.1`? Because the way I read it, it goes against this repo's documentation.

Sheikh-Abubaker:
> Are you sure that you want to have `1.13.1`? Because the way I read it, it goes against this repo's documentation.

@farcaller my bad! minor version bump already includes patch!

Sheikh-Abubaker (approved):
LGTM!

Issue #3405, opened 2024-11-06
Tempo Distributed gossip-ring Service missing annotations


rbrady:
In tempo-distributed, the service-gossip-ring.yaml file is missing the annotations block that is found in the other services.

PR #3406, opened 2024-11-06
Adds annotations to gossip-ring service in tempo-distributed

rbrady:
In the tempo-distributed the gossip-ring service is the only service in the chart that does not have annotations. This PR adds the annotations based on .Values.tempo.memberlist as used elsewhere in the file.

Closes: #3405


zalegrala (approved):
Nice, thanks for the PR.  This will need a rebase and version bump for a change that went in right before this one.

rbrady:
Thanks @zalegrala! rebased and bumped.

joe-elliott (approved):

PR #3416, opened 2024-11-10
revert jump to tempo 2.6.1

trevatk:
revert jump, breaking change when tempo is installed using helm chart with `tempoQuery.enabled` is set to `true`

`
flag provided but not defined: -query.base-path
Usage of /tempo-query:
  -config string
        A path to the plugin's configuration file
`


Sheikh-Abubaker (commented):
Hey @trevatk PRs should be based on different branch other than main, as the best practice any modifications in the main branch is prohibited, you need to create a new branch off of main, make changes on that new branch and then you may propose a PR.

Issue #3068, opened 2024-04-08
Add OCI repository publish to ghcr.io for helm charts released from other repositories


krajorama:
Followup to #2998 #2395 

Implement releasing LGTM databases and other helm charts to github packages. 
Update 
https://github.com/grafana/helm-charts/blob/main/.github/workflows/update-helm-repo.yaml

benoitschipper:
Hey All, I see there are already some Helm Charts being published as OCI Artifact (packages) here: https://github.com/orgs/grafana/packages?repo_name=helm-charts , or even within their own respective project within the grafana/*** github page.

I'd love to be able to also collect the following Helm Charts through GHCR as OCI Artifact:

- Loki - Helm Chart: https://github.com/grafana/loki/tree/main/production/helm/loki
- Mimir Distributed - Helm Chart: https://github.com/grafana/mimir
- Pyroscope - Helm Chart: https://github.com/grafana/pyroscope/tree/main/operations/pyroscope/helm/pyroscope
- Alloy - Helm Chart: https://github.com/grafana/alloy/tree/29baf3fd21a26e6004d9fef8221035db155b63db/operations/helm/charts/alloy
- Beyla - Helm Chart: https://github.com/grafana/beyla/tree/main/charts/beyla

The reason I like the OCI artifact has to do with the fact that I often require to work within airgapped environments. If the helm charts are available as OCI artifacts it allows me to gather and airgap these helm charts as OCI artifacts, the same way as I do with Images from multiple artifact repo's such as Docker/GHCR and more. Using tooling such as Skopeo Sync and Renovate to automatically alert me of a new version and decrease the amount of LCM.

Hope this makes sense! <3

domolitom:
I have created a PR to add OCI publish to the workflow that is used by 'external' repositories:
https://github.com/grafana/helm-charts/pull/3366

What do you think @nlamirault ?

narqo:
For future references, to push an OCI package with a Helm chart from another repository, the `helm-charts` repo here needs explicitly give the permissions for original repository (e.g. `grafana/mimir`) to push a package. Refer to https://github.com/grafana/helm-charts/pull/3383 for details.

narqo:
> Mimir Distributed - Helm Chart: https://github.com/grafana/mimir/tree/main/operations/helm/charts/mimir-distributed

Starting this weekly release, the `mimir-distributed` helm chart is pushed to the ghcr.io also. Check:

```
[...]

PR #3418, opened 2024-11-11
update-helm-repo: fix oci image push

narqo:
This is the fixup for https://github.com/grafana/helm-charts/pull/3366

Relates to #3068

It turned out GitHub doesn't support publishing a package using GitHub App installation tokens (or fine-grained personal tokens). That is, as of today, the users are left with using either "classic" PAT with org-scopes or the magical `GITHUB_TOKEN` ([docs](https://docs.github.com/en/packages/learn-github-packages/publishing-a-package#publishing-a-package)).

This PR fixes the `update-helm-repo` workflow, updating the last step to use the `GITHUB_TOKEN` token. 

Note that, the package is pushed to the `GITHUB_REPOSITORY_OWNER` repository, under the name `helm-charts/<chart-name>`. The package the is automatically associated with the source repository, that run the push.

### DIFF
--- a/.github/workflows/update-helm-repo.yaml
+++ b/.github/workflows/update-helm-repo.yaml
@@ -94,8 +94,7 @@ jobs:
     needs: [setup]
     runs-on: ubuntu-latest
     permissions:
-      contents: write # to push chart release, create release, and push tags to github
-      packages: write # to push package to ghcr
+      packages: write # allows GITHUB_TOKEN to push package to ghcr
     env:
       github_app_id: ${{ secrets.github_app_id }}
     if: needs.setup.outputs.changed == 'true'
@@ -234,7 +233,9 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ env.AUTHTOKEN }}
+          # GitHub App installation token cannot push package to the org's package registry;
+          # using GITHUB_TOKEN here instead to avoid using PAT.
+          password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Push charts to GHCR
         run: |


.github/workflows/update-helm-repo.yaml
```diff
@@ -234,7 +233,9 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ env.AUTHTOKEN }}
+          # GitHub App installation token cannot push package to the org's package registry;
+          # using GITHUB_TOKEN here instead to avoid using PAT.
+          password: ${{ secrets.GITHUB_TOKEN }}
```
dimitarvdimitrov, line 238: what was the reason for using `AUTHTOKEN` at all in this workflow? isn't it that we want to inject a token from another repo which might have different permissions from the `GITHUB_TOKEN` that the helm-charts repo gets?

.github/workflows/update-helm-repo.yaml
```diff
@@ -94,8 +94,7 @@ jobs:
     needs: [setup]
     runs-on: ubuntu-latest
     permissions:
-      contents: write # to push chart release, create release, and push tags to github
```
dimitarvdimitrov, line 97: we still create a GH release. Don't we still need `contents`

https://github.com/grafana/helm-charts/blob/226cf91bb0222cc2a2cc66449a24395e7b041127/.github/workflows/update-helm-repo.yaml#L205-L219

.github/workflows/update-helm-repo.yaml
narqo: `GITHUB_TOKEN` doesn't let you push content cross-repo (i.e. from `mimir` to `helm-charts`), thus the options are either use a PAT or GitHub app installation. But packages live in the org — in `ghcr.io/grafana` — and only PAT or `GITHUB_TOKEN` can push there.

dimitarvdimitrov: ok, then i'm with you

.github/workflows/update-helm-repo.yaml
narqo: As I understand it, the `permissions` only apply to `GITHUB_TOKEN`. But this workflow publishes the content via the app install token, which has the scope assigned when the app is installed.

Not that the workflow never had `permissions`. They were added only recently in the https://github.com/grafana/helm-charts/pull/3366 (I don't think this was necessary)

dimitarvdimitrov (approved):
SGTM

krajorama (approved):
stamp

pracucci (approved):
👍

PR #3426, opened 2024-11-13
[tempo-distributed] Being TLS management

zalegrala:


zalegrala:
I think this is probably not needed, since we can create, mount and configure the TLS outside of the chart.  I will confirm and close this out.

zalegrala:
This won't be needed to get TLS.

PR #3443, opened 2024-11-21
Allow prefixing of rollout-group value and ingester name

### LINKED ISSUE #3442: [Helm] tempo Ingester rollout-group collides with other LTGM component ingesters
**Describe the bug**

The `tempo` Helm chart with `deploymentMode: Distributed` and `zoneAwareReplication` enabled (default) generates Ingester StatefulSets with labels for [rollout-operator](https://github.com/grafana/rollout-operator). Ex [here](https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/templates/ingester/_helpers-ingester.tpl#L73-L78):

```
{{- if and .ctx.Values.ingester.zoneAwareReplication.enabled .rolloutZoneName }}
name: {{ printf "%s-%s" .component .rolloutZoneName }}
rollout-group: {{ .component }}
zone: {{ .rolloutZoneName }}
{{- end }}
```
This is identical to the issue filed against loki, here: https://github.com/grafana/loki/issues/13168

Additionally, the anti-affinity rules for loki/tempo/mimir all incorporate the StatefulSet name. In all three cases it is explicitly set to the `component` name coupled with the target zone. When deploying multiple LTGM components to the same namespace this yields a first scheduled wins problem.

boldfield:
Allow for prefixing both the `rollout-group` label and `name` label of Tempo to avoid scheduling and rollout issues when deploying multiple elements of the LTGM stack to a single namespace.

Fixes #3442


boldfield:
The corollary of this PR for the Loki charts has been merged: https://github.com/grafana/loki/pull/15063

joe-elliott:
I'm concerned about the complexity this adds to the chart. I would strongly advise not running both Loki and Tempo in the same namespace. Is there a reason to put them in one?

boldfield:
@joe-elliott we have constraints within our environment that require it and are actively doing it. These changes are opt-in so they will not add additional complexity w/r to users who do not opt to use the feature.

boldfield:
I also have an alternative solution that can be found here: https://github.com/grafana/rollout-operator/pull/190

boldfield:
Hi @joe-elliott, any additional thoughts here? The alternative approach was nixed and kustomize isn't an option for us. Thanks.

danstadler-pdx:
Brian - can you provide any more information here about why a single namespace is needed? thanks

boldfield:
@danstadler-pdx The environment in which we run at Nvidia requires us to run all systems in a single team-level namespace. This is less than ideal but a constraint we need to live within unfortunately.

dewjam:
@danstadler-pdx This would be incredibly useful for us as well.  I don't think it's all that uncommon to run in a multi-tenant K8S cluster where all the LGTM components are constrained to a single namespace.  Any chance we can get some consensus on this?

I would be happy to pick this up and rebase the PR to get it moving.

PR #3459, opened 2024-11-28
[loki-distributed] wrong indent for server config

chuan99nd:
## Bug Description
The `server` configuration in the Helm template was being incorrectly indented due to the use of an absolute indentation level (`nindent 6`). This caused issues when the server field and its nested configuration were parsed into a Go map and then re-generated as YAML. Since the indentation in YAML is relative to the parent field, using an absolute indentation level resulted in malformed YAML output.
Here is Configmap template
```
{{- if and (not .Values.loki.existingSecretForConfig) (not .Values.loki.configAsSecret) -}}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "loki.fullname" . }}
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "loki.labels" . | nindent 4 }}
data:
  config.yaml: |
    {{- tpl (mergeOverwrite (tpl .Values.loki.config . | fromYaml) .Values.loki.structuredConfig | toYaml) . | nindent 4 }}
{{- end -}}

```
You can see other case similer in `values.yaml` file
```
    {{- if .Values.loki.schemaConfig}}
    schema_config:
    {{- toYaml .Values.loki.schemaConfig | nindent 2}}
    {{- end}}
```
## Testing
Verified that the generated YAML now has the correct relative indentation under the server field.
Ensured that the fix does not introduce any regression in other parts of the configuration.


Sheikh-Abubaker:
@chuan99nd could you please address the failed checks ?

PR #3460, opened 2024-11-29
[tempo] Add topologySpreadConstraints

rim99:
Add topologySpreadConstraints, so pods can spread in multiple AZs evenly


joe-elliott:
isn't this a single binary helm chart? do topology constraints impact a single pod statefulset?

rim99:
@joe-elliott I think this can be tuned to be a statefulset with mutiple pods. 
I've created a small cluster with several pods in my environment.

rim99:
@joe-elliott I think this can be tuned to be a statefulset with mutiple pods. 
I've created a small cluster with several pods in my environment.

PR #3473, opened 2024-12-10
[rollout-operator] allow setting extra args

DrFaust92:
to configure log level and format


andyasp:
Looks like this PR was missed, sorry about that! This was added in https://github.com/grafana/helm-charts/pull/3682 so I'll close this.

PR #3475, opened 2024-12-10
[grafana] Update to version 11.4.0

terop:


terop:
I am not sure whether the minor version of the chart version should be updated instead of patch when the application minor version is updated. Please let me know if it should be changed.

jkroepke (approved):

jkroepke:
> I am not sure whether the minor version of the chart version should be updated instead of patch when the application minor version is updated. Please let me know if it should be changed.

According to semver Patch for fixes and minor for new features. 

Bump minor would be a better option. 

I'm pre-approve that so the second reviewer could take note of that.

terop:
Thank you for the comment @jkroepke. I updated the version field per your suggestion.

Sheikh-Abubaker (approved):

PR #3477, opened 2024-12-11
[tempo-distributed] Disabled reporting usage by default

xogoodnow:
Hi,
IMHO, reporting usage must not be enabled by default (since the report is users privacy and should only be enabled if the user actually wants it).
This also is not the case in monologic version (disabled by default).

p.s: I did not update the conf doc yet.
[tempo-configuration](https://grafana.com/docs/tempo/latest/configuration/#usage-report)

Cheers

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -43,7 +43,7 @@ externalRuntimeConfigName: '{{ include "tempo.resourceName" (dict "ctx" . "compo
 externalConfigVersion: '0'
 
 # -- If true, Tempo will report anonymous usage data about the shape of a deployment to Grafana Labs
-reportingEnabled: true
+reportingEnabled: false
 
 tempo:
   image:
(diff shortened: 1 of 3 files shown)


charts/tempo-distributed/values.yaml
```diff
@@ -43,7 +43,7 @@ externalRuntimeConfigName: '{{ include "tempo.resourceName" (dict "ctx" . "compo
 externalConfigVersion: '0'
 
 # -- If true, Tempo will report anonymous usage data about the shape of a deployment to Grafana Labs
-reportingEnabled: true
```
zalegrala, line 46: I think we would like to leave this.  Feel free to call something out more explicit in the docs if you'd like.

zalegrala (changes requested):
The reporting helps us understand the deployments in the wider community and can help inform us when we deprecate or make breaking changes.  If we want something in the docs to more explicitly point out where to change the reporting, that sounds okay, but the default value I thin should remain.

charts/tempo-distributed/values.yaml
xogoodnow: Sure, the comments seems quite straight forward (if you have something else in mind please let me know)

xogoodnow:
Hi @zalegrala,

Sure, I understand.
It was just my opinion to have this disabled by default.
Thank you

Cheers

xogoodnow:
Hi @joe-elliott,

Sure, I understand.

Cheers

PR #3479, opened 2024-12-11
[tempo-distributed] Declare after installation

xogoodnow:
Added other components.

p.s: 
Enterprise gateway and the Memcached exporter are not included. should they be included as well?


zalegrala:
Including the enterprise components if enterprise is enabled would be a great update too.  Thanks for the PR.

xogoodnow:
> Including the enterprise components if enterprise is enabled would be a great update too. Thanks for the PR.

Added all components.
Also changed "gateway.gem.example.com" to "gateway.get.example.com" since its Enterprise Gateway for Traces and not metrics (sorry I'm a little OCD :) ).

Cheers

zalegrala (approved):

zalegrala:
Alignment in the names with the mimir chart is okay with me.  Do we know if this will be a breaking change?

xogoodnow:
> Alignment in the names with the mimir chart is okay with me. Do we know if this will be a breaking change?

Great, I do not think it would be a breaking change.
User can provide a domain for all services which have ingress and then override them when needed and the sub-domain for the service would contain the release name such as below:
query-front.{{ {{ .Release.Name }}.example.com

I would open a PR for this then.
Cheers

xogoodnow:
Hi @zalegrala,

mapno (approved):

xogoodnow:
Hi @zalegrala, @mapno ,

xogoodnow:
> @xogoodnow, Thanks! for the PR, could you please address the failed CI check.

Hi @Sheikh-Abubaker,
Sure, bumped the version

PR #3480, opened 2024-12-11
[tempo-distributed] Updated memcached image

xogoodnow:
Updated and tested Memcached from 1.6.29 to 1.6.33

Below is the link for the changes:

https://github.com/memcached/memcached/compare/1.6.29...1.6.33


zalegrala (approved):
We'll need to bump the chart version and update the docs.  Thanks for the PR.

PR #3486, opened 2024-12-13
Update values.yaml - enable alertmanager v2 api by default

jessebot:
This enables alertmanger's V2 api by default for the loki-distributed chart for loki ruler. Helps with making this more clear https://github.com/grafana/loki/issues/12859 (it should be fixed upstream, but for some reason, I still had to set it manually when I installed this chart)

### DIFF
--- a/charts/loki-distributed/Chart.yaml
+++ b/charts/loki-distributed/Chart.yaml
@@ -3,7 +3,7 @@ name: loki-distributed
 description: Helm chart for Grafana Loki in microservices mode
 type: application
 appVersion: 2.9.10
-version: 0.80.0
+version: 0.81.0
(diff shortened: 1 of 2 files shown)


Sheikh-Abubaker:
@jessebot Maybe it's due to the fact that the current loki application version doesn't use v2 by default:
https://github.com/grafana/helm-charts/blob/7f540d715efbfc9bcd941655fa9adf870b4f714b/charts/loki-distributed/Chart.yaml#L5


The official Loki docs too states it is enabled by default since  3.2.0: https://github.com/grafana/loki/blob/58840c5102c0ab8b13fb668aadacaf527dd026df/docs/sources/alert/_index.md?plain=1#L318

PR #3490, opened 2024-12-15
[tempo-distributed] Added global storage class

xogoodnow:
Added global storage class.

Generally, people do not set different storageClasses for different components.
So having a global variable would be quite convenient.

Cheers


joe-elliott:
Thanks for the PR but will need to updated to tip of main

xogoodnow:
> Thanks for the PR but will need to updated to tip of main

Done

xogoodnow:
Hi @joe-elliott ,

Sheikh-Abubaker (approved):

PR #3493, opened 2024-12-16
[tempo-distributer] Pdb ingestor

xogoodnow:
Hi,

the ".Values.podDisruptionBudget" condition is never met and the file is redundant.
Also uncommented the "Maxunavailable" parameter for "Ingestor "

Cheers


xogoodnow:
Hi @Sheikh-Abubaker,

PR #3498, opened 2024-12-22
[tempo-distributed] Ports/service related configuration not configureable

xogoodnow:
Hi,

Cheers


xogoodnow:
Hi,

The target port for gateway is still static. since the image is unprivileged, I did not want to change the logic (the nginx config should have been changed as well).

I see that all other components have the same problem.
So either the config section(for ports) should be removed and just refer to the documentation or they would actually be checked within the templates.
I will change the rest and send theme on this branch to be checked.

Cheers

PR #3501, opened 2024-12-22
[tempo-distributed] Static ports

xogoodnow:
Hi,

Cheers


xogoodnow:
I went through almost all services.
1. If the ports section are there so folks would be able to change the ports, they must not be set statically but if the section is just informative, we can commit them all and just provide a link to the documentation.
2. This way changing any default port would be possible in the future

Cheers

PR #1661, opened 2022-08-03
document how to re-generate the chart readme

neiser:
As suggested here, my first-time contribution, I struggled how to re-generate the readme. I've added documentation to help future first-time contributors.

PR #3524, opened 2025-01-14
[grafana] add session affinity config

### LINKED ISSUE #3520: [grafana]: Support `sessionAffinity` in `Service` template.
It would be nice to be able to configure `sessionAffinity` for Grafana service, something like:

```
service:
  sessionAffinity: ClientIP
```

or at very least make it `ClientIP` by default. For those who have only one replica - having default `ClientIP` or `None` does not make difference.

In my case the problem was that I wanted to have 2 replicas, but I couldn't login because service does round-robin, causing infinite redirects from and to login page. (I'm logged in on one instance obviously, but not the other.)

KyriosGN0:
Closes #3520

### DIFF
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -75,6 +75,7 @@ need to instead set `global.imageRegistry`.
 | `service.enabled`                         | Enable grafana service                        | `true`                                                  |
 | `service.ipFamilies`                      | Kubernetes service IP families                | `[]`                                                    |
 | `service.ipFamilyPolicy`                  | Kubernetes service IP family policy           | `""`                                                    |
+| `service.sessionAffinity`                 | Kubernetes service session affinity config    | `""`                                                    |
 | `service.type`                            | Kubernetes service type                       | `ClusterIP`                                             |
 | `service.port`                            | Kubernetes port where service is exposed      | `80`                                                    |
 | `service.portName`                        | Name of the port on the service               | `service`                                               |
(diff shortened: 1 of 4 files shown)


charts/grafana/README.md
jkroepke, line 1: Please revert all unrelated Changes in the README and only add the new value, thanks.

KyriosGN0: fixed that, worth to update the main CONTRIBUTING.md file that says to not run the docker step in the grafana folder.

jkroepke: @zanhsieh fyi - the info was introduced in https://github.com/grafana/helm-charts/pull/1661 but seems not applied for grafana chart yet.

charts/grafana/templates/service.yaml
```diff
@@ -48,6 +48,9 @@ spec:
   {{- with .Values.service.externalTrafficPolicy }}
   externalTrafficPolicy: {{ . }}
   {{- end }}
+  {{- if .Values.service.seessionAffinity }}
+  sessionAffinity: {{ .Values.service.sessionAffinity }}
```
jkroepke, line 52: For consistency reasons, it should be

```suggestion
  {{- with .Values.service.seessionAffinity }}
  sessionAffinity: {{ . }}
```

like the line above.

jkroepke (approved):
One small nit, one resolved, LGTM. Leaving room for the final reviewer.

zanhsieh (approved):

Issue #1556, opened 2022-07-01
[grafana] imagerenderer.enable doesn't set capture option in grafana.ini


tim-felix-lange:
When you enable the grafana image renderer via `imageRenderer.enabled: true`, it does create the deployment and service for the renderer as expected, but I would expect it to also enable images in alerts, since this seems to be the main feature for what the image renderer is needed. However, including images in alerts requires setting the capture option in grafana.ini
```
[unified_alerting.screenshots]
capture = true
```

I'm not sure whether this is considere a bug in the chart caused by the advent of unified alerting in Grafana 8 or it is some (weird) intended behaviour that is missing documentation. In the latter case a note that capture needs to be enabled separately would be helpful and would have saved me some days :-)

bbedward:
Is this still an issue? I was trying to figure out how to enable capture screenshots in the helm chart as well.

tim-felix-lange:
As far as I can tell yes. But if you're struggling with screenshot capture, also mind #795 though it's flagged as merged, but I'm not sure whether it made it into the release already.
I was able to get the screenshots in the my alerts with
```
imageRenderer:
  enabled: true
  networkPolicy:
    limitIngress: false
grafana.ini:
  unified_alerting.screenshots:
    capture: true
```

PR #3537, opened 2025-01-22
[grafana] set unified_alerting.screenshots.capture to true when imageRenderer is enabled

kefi550:
~~Set environment variable `GF_UNIFIED_ALERTING_SCREENSHOTS_CAPTURE` to true when `imageRenderer.enabled` is true.~~
I changed values.yaml so that unified_alerting.screenshots.capture is true in grafana.ini when imageRenderer is enabled.
This will cause screenshots to be attached to alerts when `imageRenderer.enabled` is set to true.

As mentioned in #1556, currently screenshot is not attached to alerts even though `imageRenderer.enabled` is set to true.
It would be natural for alerts to have a screenshot attached when the imageRenderer is enabled.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 8.8.5
+version: 8.8.6
 appVersion: 11.4.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -867,6 +867,8 @@ grafana.ini:
     url: https://grafana.net
   server:
     domain: "{{ if (and .Values.ingress.enabled .Values.ingress.hosts) }}{{ tpl (.Values.ingress.hosts | first) . }}{{ else }}''{{ end }}"
+  unified_alerting.screenshots:
+    capture: "{{ .Values.imageRenderer.enabled }}"
 ## grafana Authentication can be enabled with the following values on grafana.ini
  # server:
       # The full public facing url you use in browser, used for redirects and emails


jkroepke:
I feel this should be configurable. and please bump the chart version.

zanhsieh (approved):

jkroepke:
I read your message, but I think is should be configurable with default true. Sometimes, there are reasons to disable "cool" feature, for example for cost reasons.

kefi550:
I understood that setting in environment variables is higher priority than grafana.ini and is not truly configurable.

I changed values.yaml so that unified_alerting.screenshots.capture is true in grafana.ini when imageRenderer is enabled.
This allows configuring the behavior of unified_alerting.screenshots.capture by changing grafana.ini or env in values.yaml.

(The title and description of the PR have been changed accordingly)

jkroepke:
Hi @kefi550 

I guess your are using GenAI, I used GenAi as well:

Q: Should unified_alerting.screenshots.capture always set to true if the image renderer is enabled?

A: Based on the knowledge sources provided, I can answer your question about the unified_alerting.screenshots.capture setting in Grafana.  The unified_alerting.screenshots.capture setting should not always be set to true just because the image renderer is enabled. This setting specifically controls whether screenshots are captured for alert notifications, which is a separate feature from the general image rendering capabilities. According to the [Grafana documentation on image notifications](https://grafana.com/docs/grafana/latest/alerting/configure-notifications/template-notifications/images-in-notifications/#configuration), you need to explicitly enable screenshot capture for alert notifications by setting capture to true in the [unified_alerting.screenshots] section of the configuration:  
 # Enable screenshots in notifications. You must have either installed the Grafana image rendering # plugin, or set up Grafana to use a remote rendering service. # For more information on configuration options, refer to [rendering]. capture = false 
  The default value is false, which means screenshots are not captured for alert notifications even if the image renderer is enabled.  It's important to note that enabling this feature has some requirements and considerations:  1. You must have the image rendering capability set up, either through the Grafana Image Renderer plugin or a remote rendering service. 2. Capturing screenshots for alerts can be resource-intensive, especially in terms of memory usage. 3. You need to consider where the screenshots will be stored or uploaded, which may require additional configuration.  Therefore, you should only set unified_alerting.screenshots.capture to true if you specifically want to include screenshots in alert notifications and have ensured that your system meets the necessary requirements to support this feature.  For more detailed information on configuring and using images in notifications, you can refer to the [Grafana documentation on using images in notifications](https://grafana.com/docs/grafana/latest/alerting/configure-notifications/template-notifications/images-in-notifications/#use-images-in-notifications).


Based on that feedback, the value should not be forced to yes.

PR #3544, opened 2025-01-25
Configure ports

xogoodnow:
Hi,

I see that almost all other services (except for metrics-generator) are static.

Cheers


xogoodnow:
I went through almost all services and mildly modified the templates so the ports could be modified dynamically.

If the ports section are there so folks would be able to change the ports, they must not be set statically but if the section is just informative, we can commit them all and just provide a link to the documentation.
This way changing any default port would be possible in the future
Cheers

PR #3547, opened 2025-01-28
[tempo-distributed] Fix Memcached StatefulSet volumeMounts formatting issue

baileymjensen:
## Description
In using this chart, I came across a formatting error. The conditional for disabling memcached-exporter does not wrap all of its possible settings. If you use volumeMounts and disable memcached-exporter, it still adds the volumeMounts and results in adding a second duplicate section to the memcached container config.

This change fixes that. 

I also wonder if we should set the memcached-exporter extraVolumeMounts to be configured via:
`.Values.memcachedExporter.extraVolumeMounts` rather than the current `.Values.memcached.extraVolumeMounts` so that it is possible to mount different things into each container in the case you may need to rather than them both getting the same thing.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.32.1
+version: 1.32.2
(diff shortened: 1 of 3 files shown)


javiermolinar (approved):
LGTM

SimKev2 (approved):

PR #3585, opened 2025-02-28
[grafana] Process sidecar configmap/secret label and labelValue with tpl

a-abella:
This change modifies the sidecar configmap `label` and `labelValue` values to be processed with `tpl`. This is a convenience feature for users with multiple Grafana helm releases installed to the same namespace.

This would allow users to configure a common values file for multiple helm releases to the same namespace similar to:
```yaml
sidecar:
  dashboards:
    label: 'load-for-{{ .Release.Name }}'
```

This would allow users to not have to specify a unique `.sidecar.xyz.label(Value)` for each release's values file, but still retain the ability to dictate which releases should mount which configmaps by referencing the unique release names (or other templated value).

---

A note: I attempted to follow the CONTRIBUTING.md guidelines which specified that helm-docs should be generated--I attempted to do so and most of the README was significantly altered, so I instead manually inserted the README changes. If this is improper please let me know and I can follow whatever procedure is expected.


jkroepke (approved):
LGTM

Sheikh-Abubaker (approved):

Issue #3588, opened 2025-03-05
[grafana-sampling] Update flush interval for otelcol.connector.servicegraph


bradleypettit:
The otelcol.connector.servicegraph component currently has a default flush interval of 0s.

This can cause excessive metrics generation, and an upstream fix is already in progress here - https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37966.

We should explicitly set this value to prevent any issues while waiting for the upstream changes.

PR #3589, opened 2025-03-05
[grafana-sampling] Update default servicegraph metrics_flush_interval

bradleypettit:
Fixes https://github.com/grafana/helm-charts/issues/3588

The default flush interval is 0s which generates a lot of metrics. This change will set it to 60s by default, which will eventually be the default upstream. As part of this change it is also made available in the values so it can be customised.


rlankfo (approved):

petewall (approved):
LGTM!

PR #3592, opened 2025-03-06
[tempo-distributed] adds ability to set updateStrategy in the Ingester StatefulSet

ervandenbosch:
In our LGTM project we want to be able to set the 'updateStrategy' of the Tempo Ingester StatefulSet to 'onDelete' _without_ using Zone Aware Replication.

Currently, this is not possible while using the Tempo Distributed helm chart.

With the changes in this PR, the Tempo Ingester StatefulSet should default to updateStrategy 'rollingUpdate' with partition set to '0' and override its the updateStrategy depending on either the 'zoneAwareReplication' or 'statefulStrategy' values.


mapno (commented):
The changes to `tempo-distributed` look ok, but it contains many more changes and it's also missing a version bump.

ervandenbosch:
> The changes to `tempo-distributed` look ok, but it contains many more changes and it's also missing a version bump.

I removed the change to the readme's based on helm-docs and bumped the chart tempo-distributed just now :)

mapno (approved):

ervandenbosch:
@Sheikh-Abubaker  perhaps time to review?

PR #3601, opened 2025-03-13
[tempo-distributor] add scaledobject for querier

agrahamlincoln:
What this PR does

Adds a new ScaledObject resource that is optionally enabled to add KEDA autoscaling behavior to the querier component


Sheikh-Abubaker:
@agrahamlincoln, wouldn't this require keda to be installed in the cluster prior to setting up the ScaledObject resource ?

agrahamlincoln:
> @agrahamlincoln, wouldn't this require KEDA to be installed in the cluster prior to setting up the ScaledObject resource ?

yes, which is why this should be gated behind a keda.enabled flag `.Values.querier.autoscaling.keda.enabled` - however `querier.autoscaling.enabled` is not enabled by default anyways, so applying the helm chart with the default `values.yaml` will not enable this

PR #3607, opened 2025-03-17
[cloudcost-exporter] Add cloudcost-exporter Helm chart

nikimanoledaki:
ref https://github.com/grafana/cloudcost-exporter/issues/392

This PR adds the Helm chart of the cloudcost-exporter here so that we can benefit from the default Helm release workflow.

This helm Chart is a copy of what we created in the cloudcost-exporter, which has been reviewed and tested internally: https://github.com/grafana/cloudcost-exporter/tree/main/deploy/helm/cloudcost-exporter


nikimanoledaki:
Added an exclusion so that cloudcost-exporter is not tested here as part of the `chart-testing` action. 

Reasons:
* This would require IRSA, which would require adding an action to fetch secrets from Vault to get the OIDC path and Role ARN.
* Even so, the condition would not be met since the Service Account name  is not unique e.g. in [this run](https://github.com/grafana/helm-charts/actions/runs/13908120517/job/38915789747?pr=3607) it was `cloudcost-exporter-frlzwicnzo`, which changes with every kind cluster.
* This IRSA setup for GitHub Actions would be different from how a user would setup auth for cloudcost-exporter, which could be useful as an integration test, but in this case we are not testing this so this is redundant and adds complexity without any benefit
* There is an internal CD pipeline that would be triggered when there is a new Helm chart version so it can still tested that way

charts/cloudcost-exporter/.helmignore
Duologic, line 1: I often see these files, shouldn't most of these be part of people's personal `.gitignore` instead of in a repository?

Duologic: Ah, wait, that's `.helmignore`, same question tho.

nikimanoledaki: That's a good point. This `.helmignore` was generated as part of the initial Helm chart bootstrapping command (https://helm.sh/docs/chart_template_guide/helm_ignore_file/) (better safe than sorry etc!)

PR #3609, opened 2025-03-17
[synthetic-monitoring-agent] Update outdated chart & add extraEnv

hnajib-sym:
- Set to latest release of synthetic-monitoring-agent is [v0.34.4](https://github.com/grafana/synthetic-monitoring-agent/releases/tag/v0.34.4)

- Kubernetes requirement removed , doesn't make sense since chart is very basic.

- UID in Dockerfile is set to [12345](https://github.com/grafana/synthetic-monitoring-agent/blob/main/Dockerfile#L24) hence ```yaml runAsUser: 12345``` and ```yaml fsGroup: 12345```

> [!NOTE]  
> Pod fails with default UID 65534 , fsGroup updated to align 

- ```NET_RAW``` capabilities is required else pod fails with ```operation not permitted```
> [!NOTE]  
> synthetic-monitoring-agent Docker file is already add [cap_net_raw](https://github.com/grafana/synthetic-monitoring-agent/blob/main/Dockerfile#L15) 

> [!WARNING]  
> drop ALL take precedence and pod fails , so it was removed

- ```yaml extraEnv```  was added to able to pass GO PROXY environments variables : HTTP_PROXY & HTTPS_PROXY


zanhsieh:
@hnajib-sym 
Can you use helm-docs to regen README.md plz?

zanhsieh (approved):

Sheikh-Abubaker (approved):

Issue #3647, opened 2025-04-12
add support to set the env var RESOURCE_NAME for the sidecar images


CarstenSon:
It would be awesome if we could set the RESOURCE_NAME env var for the sidecar images to let that fetch only certain secrets/configmaps from a value within the Grafana helm chart

CarstenSon:
This might e.g. be helpful if you want the sidecar to only fetch certain resources with specific names and enables you to have a very strict RBAC-Role in place on k8s to have additional security in any given setup

PR #3649, opened 2025-04-12
feat: add possibility to set env var RESOURCE_NAME using grafana-helm chart values

CarstenSon:
add possibility to set RESOURCE_NAME env var from grafana helm-chart to make sidecars fetch only specific resources

closes https://github.com/grafana/helm-charts/issues/3647

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 8.11.4
+version: 8.12.0
 appVersion: 11.6.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -353,6 +353,10 @@ containers:
         value: "/etc/grafana/provisioning/alerting"
       - name: RESOURCE
         value: {{ quote .Values.sidecar.alerts.resource }}
+      {{- if .Values.sidecar.alerts.resourceName }}
+      - name: RESOURCE_NAME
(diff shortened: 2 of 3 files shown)


jkroepke (commented):
could you please move the description over the property in the values.yaml?

CarstenSon:
@jkroepke: I added an exemplary value below the description so I guess it should be fine now - the resource_name value is a different one than the resource one which might be a bit misleading :-)

jkroepke:
Ah, now I get it. But could you please define an empty default value?

Like

```
    # resourceName: comma separated list of resource names to be fetched/checked by this sidecar.
    # per default all resources of the type defined in {{ .Values.sidecar.plugins.resource }} will be checked.
    # This e.g. allows stricter RBAC rules which are limited to the resources meant for the sidecars.
    # resourceName: "secret/plugins-0,configmap/plugins-1"
    resourceName: ""
```

CarstenSon:
@jkroepke since the implementation in the sidecar image uses RESOURCE_NAME as an optional environment variable I would much rather have it as an optional value like this without a default value. 

The sidecar would otherwise check for no secrets/configmaps per default while the default in the sidecar container would be accessing all secrets/configmaps - whatever is set up in the RESOURCE env-var 

Or would it be the preferred version in cases like this to just set up the extra env-vars using the existing options with .Values.sidecars.<sidecarName>.env ?

CarstenSon:
@jkroepke : ah - now I got it.. the if condition will evaluate to false anyway as it's an empty string so you just have the advantage of autocompletion for jetbrains IDEs while the value is still ommited and thus the default behavior not changed - will change that.

CarstenSon:
@jkroepke: done and thx for the remarks :smile:  learned from it and will use that in the future :+1:

jkroepke (approved):

zalegrala (approved):

PR #3656, opened 2025-04-17
[rollout-operator] fix: add missing patch permission to apps/statefulsets

paulojmdias:
Based on tests, I think the `patch` permission to `apps/statefulsets` is missing.

```
level=warn ts=2025-04-17T09:02:31.196597988Z msg="unable to adjust desired replicas of StatefulSet" group=ingester err="statefulsets.apps \"loki-general-ingester-zone-b\" is forbidden: User \"system:serviceaccount:loki:loki-general-rollout-operator\" cannot patch resource \"statefulsets\" in API group \"apps\" in the namespace \"loki\""
level=debug ts=2025-04-17T09:02:31.196748498Z msg="reconciling StatefulSet" statefulset=loki-general-ingester-zone-a
level=debug ts=2025-04-17T09:02:31.196812368Z msg="reconciling StatefulSet" statefulset=loki-general-ingester-zone-b
level=debug ts=2025-04-17T09:02:31.196850638Z msg="reconciling StatefulSet" statefulset=loki-general-ingester-zone-c
level=info ts=2025-04-17T09:02:31.196885908Z msg="reconcile done"
```

Also, you can see this is being added on the [development](https://github.com/grafana/rollout-operator/blob/main/development/rollout-operator-role.yaml) manifests.

PR #3657, opened 2025-04-17
chore(github): Update codeowners check to run on latest ubuntu

56quarters:
It previously explicitly used ubuntu-20.04 which is EOL.

Related #3656


andyasp (approved):

narqo (approved):
🔥

PR #3669, opened 2025-04-25
[grafana] Add support for custom script and extra mounts in Grafana sidecar

refucktor:
## Background

- Grafana's chart uses [kiwigrid/k8s-sidecar](https://github.com/kiwigrid/k8s-sidecar) as default container image for sidecar and init containers.
- The lifecycle execution of the "script" is after the file has been written and before executing the HTTP call.
- There are some cases where the script is used for variable or string substitution for dynamic content.
- Using the approach with an init container, the script should only run once at initialization, otherwise the watch will always detect a difference in the file and rewrite perpetually

## Goals

- Allow using the script functionality from the Grafana init container (before only conf in the side container section)
- Allow using the "extraMounts" also in the init container configuration

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -230,6 +230,10 @@ initContainers:
       - name: SKIP_TLS_VERIFY
         value: "{{ . }}"
       {{- end }}
+      {{- if .Values.sidecar.datasources.script }}
+      - name: SCRIPT
+        value: "{{ .Values.sidecar.datasources.script }}"
+      {{- end }}
     {{- with .Values.sidecar.resources }}
     resources:
       {{- toYaml . | nindent 6 }}
@@ -241,6 +245,9 @@ initContainers:
     volumeMounts:
       - name: sc-datasources-volume
         mountPath: "/etc/grafana/provisioning/datasources"
+      {{- with .Values.sidecar.datasources.extraMounts }}
+      {{- toYaml . | trim | nindent 6 }}
+      {{- end }}
 {{- end }}
 {{- if and .Values.sidecar.notifiers.enabled .Values.sidecar.notifiers.initNotifiers }}
   - name: {{ include "grafana.name" . }}-init-sc-notifiers
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -230,9 +230,9 @@ initContainers:
       - name: SKIP_TLS_VERIFY
         value: "{{ . }}"
       {{- end }}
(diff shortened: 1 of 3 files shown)


charts/grafana/templates/_pod.tpl
```diff
@@ -230,6 +230,10 @@ initContainers:
       - name: SKIP_TLS_VERIFY
         value: "{{ . }}"
       {{- end }}
+      {{- if .Values.sidecar.datasources.script }}
```
jkroepke, line 233: Please add this to values.yaml as well, including a description like

> Absolute path to a script to execute after a configmap got reloaded. It runs before calls to REQ_URI. If the file is not executable it will be passed to sh. Otherwise it's executed as is. [Shebangs](https://en.wikipedia.org/wiki/Shebang_(Unix)) known to work are #!/bin/sh and #!/usr/bin/env python

refucktor: to be proactive I took some decisions from my previous questions. If the results are ok, just ignore the questions. Basically, I went with option A, using `with` control structure and bump the patch version 😃

charts/grafana/templates/_pod.tpl
```diff
@@ -241,6 +245,9 @@ initContainers:
     volumeMounts:
       - name: sc-datasources-volume
         mountPath: "/etc/grafana/provisioning/datasources"
+      {{- with .Values.sidecar.datasources.extraMounts }}
```
jkroepke, line 248: Is the value `sidecar.datasources.extraMounts` in the values.yaml present?

charts/grafana/templates/_pod.tpl
```diff
@@ -241,6 +245,9 @@ initContainers:
...
+      {{- toYaml . | trim | nindent 6 }}
```
jkroepke, line 249: what the reason for trim here?

charts/grafana/templates/_pod.tpl
jkroepke: option a is fine.

charts/grafana/templates/_pod.tpl
refucktor: @jkroepke yes it is

charts/grafana/templates/_pod.tpl
refucktor: Honestly it is widely used in other parts of the file, I tried to keep it consistent with the rest.
I guess the idea is to optimize the final output by removing leading & trailing whitespaces

jkroepke (approved):
LGTM 👍

Sheikh-Abubaker (approved):

PR #3674, opened 2025-04-28
Github actions: harden workflows

alexweav:
Address linter feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,8 @@ name: CI
 
 on: pull_request
 
+permissions: {}
+
 jobs:
   call-check-codeowners:
     uses: ./.github/workflows/check-codeowners.yaml
--- a/.github/workflows/validate-pr.yaml
+++ b/.github/workflows/validate-pr.yaml
@@ -8,6 +8,10 @@ on:
       - reopened
       - synchronize
 
+permissions:
+  pull-requests: read
+  contents: read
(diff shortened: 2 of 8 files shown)


.github/workflows/validate-pr.yaml
```diff
@@ -16,10 +20,10 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
+          persist-credentials: false
 
       - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.0.1
-
+        uses: helm/chart-testing-action@b0d4458c71155b54fcf33e11dd465dc923550009 # v2.0.1
```
alexweav, line 26: When pinning I didn't alter any versions, even if the version appeared to be a typo (all the other workflows use v2.1.0 not v2.0.1)

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on: pull_request
 
+permissions: {}
```
julienduchesne, line 5: You'll probably need `contents: read`, otherwise you can't checkout I think

julienduchesne (approved):

.github/workflows/ci.yml
julienduchesne: Oh, indeed 👍. I misread and thought those were actions, not actual workflows

PR #3676, opened 2025-04-29
feat: add overridable max_span_attr_byte for distributor

dd84ai:
## Description
adding overridable max_span_attr_byte for distributor config
https://grafana.com/docs/tempo/latest/configuration/#distributor
as according to documentation, this parameter is configurable

## Motivation:
It is desirable to be overridable to higher value like 20'0000 for the sake of seeing larger SQL queries in Tracing attributes
received from opentelemetry instrumentation psycopg2 auto instrumentation in python
https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-psycopg
For my team it is important feature to have regardless of extra involved RAM costs and disk usages and etc.

## Comment:
I already checked it works in deployment using `tempo-distributed-1.35.0` version
```
distributor:
  max_span_attr_byte: 20480
```
becomes visible in /conf/tempo.yaml inside distributor and traces appear having more full SQL queries attributes as desired.


dd84ai:
@QuentinBisson also bless you for giving a check and sorting out, even if the result ended in closing of it.
Nice to see some movement to be present now. Thank you. ^_^

PR #3684, opened 2025-05-01
[grafana] Explicitly drop all unused capabilities for init-chown-data and set readonlyRootFilesystem

jcpunk:
The trivy scanner recommends explicitly dropping all capabilities. The explicit `add` ensures `CHOWN` is still permitted.

Similarly, readonlyRootFilesystem is recommended.

PR #3685, opened 2025-05-01
Set init-chown-data readOnlyRootFilesystem

jcpunk:
Trivy recommends setting `readOnlyRootFilesystem` where possible.


jkroepke:
can we combine this and #3684 ?

PR #3698, opened 2025-05-05
[grafana] Add `DAC_READ_SEARCH` to `initChownData` and `readOnlyRootFilesystem`

jcpunk:
This should resolve the regression from https://github.com/grafana/helm-charts/pull/3684

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 8.15.0
+version: 8.15.1
 appVersion: 11.6.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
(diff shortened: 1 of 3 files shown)


Kimi450:
I tested the restart usecase explained in https://github.com/grafana/helm-charts/issues/3699 and this fixes that at least

1. Install the chart
`helm upgrade -i my-release grafana/grafana --set initChownData.enabled=true --set persistence.enabled=true --wait`
2. Once the pod is running, delete it
3. Observe that it cannot restart because the initcontainer fails with
```19:52:08 dell-e7440 [--0] ~  $ k logs -c init-chown-data my-release-grafana-846948d9f4-vgxm7
chown: /var/lib/grafana/csv: Permission denied
chown: /var/lib/grafana/png: Permission denied
chown: /var/lib/grafana/pdf: Permission denied
19:52:16 dell-e7440 [--0] ~  $
```
4. updated the deployment with the changes proposed here
5. new pod comes up

jkroepke:
> baseline policy

Generally, chown is a privileged operation. If you want to be compliant with baseline, just disable the initContainer. In modern Kubernetes environment, its generally not recommend to use that approach.

markussiebert:
But I don't understand - shouldn't the default be the hardened? I am only here, because of the issues. Never decided to activate the init container. Never had any psp issues until the cap of this pr was added. And chown alone doesn't violate the psp.

jkroepke:
But how the PSP can affect your deployment, if chmod container is not enabled on your deployment?

markussiebert:
The default is enable chown

jkroepke:
Thats the root issue. this should be not enabled by default.

jkroepke:
Closed - see https://github.com/grafana/helm-charts/issues/4087

PR #3705, opened 2025-05-07
grafana: image-renderer pod instance labels equals pod name

ebuildy:
Is a good practice that Label app.kubernetes.io/instance  match metadata.name.

According Polaris https://polaris.docs.fairwinds.com/checks/reliability/ "metadataAndInstanceMismatched" rule.


jkroepke:
Yikes, this is an painful breaking change that involes manual action for each end-user.

PR #3713, opened 2025-05-15
[grafana] Add support for envValueFrom in alerts sidecar container 

sangheee:
Hi,
I noticed that other sidecar containers support setting environment variables using both `value` and `valueFrom`, **but the alerts sidecar container currently does not.**
So I added support for it.

I would like to add environment variables like the example below.
```yaml
  alerts:
    enabled: true
    envValueFrom:
        self: 
            fieldRef:
                apiVersion: v1
                fieldPath: metadata.namespace
```

Could you please take a look?


jkroepke (approved):
LGTM

charts/grafana/ci/with-sidecars-envvaluefrom-values.yaml
```diff
@@ -14,6 +14,21 @@ extraObjects:
       var2: "dmFsdWUy"
 
 sidecar:
+  alerts:
+    enabled: true
```
Sheikh-Abubaker, line 18: should the alerts be enabled by default ?

jkroepke: Inside tests, (ci/values.yaml), yes.

Sheikh-Abubaker (approved):

PR #3715, opened 2025-05-17
[tempo-distributed] Remove replicas when autoscaling is enabled for compactor

j-jasmin:
This MR removes the `replicas` field from the compactor deployment when autoscaling is enabled. This prevents the deployment from being scaled back to the fixed replica count during rollouts, before the HPA/KEDA takes over again.


Sheikh-Abubaker (approved):

jkroepke (approved):

j-jasmin:
Could someone merge this? 🙇‍♀️ With every ArgoCD reconciliation it's throwing away the compactor pods which makes the metric for outstanding compaction blocks pretty unreliable

PR #3728, opened 2025-06-01
perf(chart): support ipv6 cluster on tempo-gateway in helm charts

kamontat:


charts/tempo-distributed/values.yaml
```diff
@@ -2105,6 +2105,7 @@ gateway:
 
         server {
           listen             8080;
+          listen             [::]:8080;
```
zalegrala, line 2108: Does this work if the pod only has ipv4?

kamontat:
@zalegrala You should checkout grafana/helm-charts#2953 as it a full loop of ipv6 supports

PR #3737, opened 2025-06-06
[tempo-distributed] Fixes #3736: use correct storageClassName for zone-aware ingester PVCs

### LINKED ISSUE #3736: [tempo-distributed] incorrect storageClass in volumeClaimTemplates when using zoneAwareReplication
**Describe the bug**  
When using `zoneAwareReplication` with different `storageClass` values per zone, the `storageClassName` in `volumeClaimTemplates` for `ingester` StatefulSets is not rendered correctly. Instead, it falls back to the global or empty value, ignoring the zone-specific `storageClass`.

**To Reproduce**  
Use the following values.yaml configuration:

```yaml
ingester:
  persistence:
    enabled: true
    size: 12Gi
  zoneAwareReplication:
    enabled: true
    topologyKey: topology.kubernetes.io/zone
    zones:
      - name: ru-2a
        nodeSelector:
          topology.kubernetes.io/zone: ru-2a
        storageClass: fast.ru-2a
      - name: ru-2b
        nodeSelector:
          topology.kubernetes.io/zone: ru-2b
        storageClass: fast.ru-2b
      - name: ru-2c
        nodeSelector:
          topology.kubernetes.io/zone: ru-2c
        storageClass: fast.ru-2c
```

After installing the Helm chart, inspect the StatefulSets for the ingesters. The `storageClassName` in `volumeClaimTemplates` does not reflect the zone-specific value (`fast.ru-2a`, etc.).

**Expected behavior**  
Each StatefulSet corresponding to a zone should use its respective `storageClass` as defined under `zones[].storageClass`.

**Actual behavior**  
All StatefulSets either use the global `storageClass`, a default value, or omit `storageClassName` entirely.

**Environment:**  
- Tempo chart version: `1.41.1`  

**Additional Context**  
The tempo-distributed/templates/ingester/statefulset-ingester.yaml template uses a $storageClass variable defined outside the range $zoneName, $rolloutZone, which causes it to ignore per-zone overrides from zoneAwareReplication.zones[].storageClass.

Proposed fix:
Use $rolloutZone.storageClass inside volumeClaimTemplates when rendering zone-specific StatefulSets.

shityakov:
🛠️ What this PR does
This PR fixes an issue with the tempo-distributed chart when using zone-aware replication for ingesters.
Specifically, it ensures that the correct storageClassName is passed into the volumeClaimTemplates for each zone.

Previously, the storageClassName value from zones[].storageClass was not used in the generated StatefulSets. This caused the PVCs to default to an incorrect or global storage class, which may lead to provisioning failures or incorrect zone placement.

✅ Changes made
Use the zone-specific storageClass value from the zone map when rendering the PVC block.

🔗 Related issue
Fixes #3736

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.3.0
+version: 10.3.1
 appVersion: 12.3.0
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -229,7 +229,7 @@ initContainers:
       {{- end }}
       {{- if .Values.sidecar.datasources.startupProbe }}
       - name: HEALTH_PORT
-        value: "{{ .Values.sidecar.datasources.startupProbe.httpGet.port | default "8080" }}"
+        value: "{{ .Values.sidecar.datasources.startupProbe.httpGet.port | default "8081" }}"
       {{- end }}
       {{- if .Values.sidecar.datasources.ignoreAlreadyProcessed }}
       - name: IGNORE_ALREADY_PROCESSED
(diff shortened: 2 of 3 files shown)


QuentinBisson (approved):
LGTM

PR #3741, opened 2025-06-09
[tempo-distributed] Remove support for Agent and add docs link to metamonitoring.

mattdurham:
Agent will be EOL in november and enterprise is removing agent for k8s-monitoring.


mattdurham:
This also felt prudent, I sort of hate directly link to enterprise docs from here. @knylander-grafana thoughts on how to make that better? Pinging @hedss and @petewall for some additional love.

petewall:
Just chiming in as [a chicken](https://en.wikipedia.org/wiki/The_Chicken_and_the_Pig), but this looks great to me!

knylander-grafana (commented):
Updates to readme look good. Thank you for doing this.  We may want to revisit the tempo-distributed Helm chart doc in the tempo repo to see if there are any updates we need.

PR #3746, opened 2025-06-11
[grafana] allow alerting and datasources to be templated strings

michaelruigrok:
For each key-value pair in alerting and datasources, if the value is a string, apply it as a single template

This allows you to programmatically generate the yaml structure itself. For instance, if you have several alerting rules with only a couple values changing, you can now create a named template that takes only the few values you care about as parameters.

### DIFF
--- a/charts/grafana/templates/image-renderer-service.yaml
+++ b/charts/grafana/templates/image-renderer-service.yaml
@@ -11,7 +11,7 @@ metadata:
     {{- end }}
   {{- with .Values.imageRenderer.service.annotations }}
   annotations:
-    {{- toYaml . | nindent 4 }}
+    {{- tpl (toYaml . | nindent 4) $ }}
   {{- end }}
 spec:
   type: ClusterIP
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.42.0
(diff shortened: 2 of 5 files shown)


charts/grafana/templates/_config.tpl
```diff
@@ -54,7 +57,10 @@ grafana.ini: |
 {{- end }}
 
 {{- range $key, $value := .Values.alerting }}
-{{- if (hasKey $value "file") }}
+{{- if kindIs "string" $value }}
```
jkroepke, line 60: Please put the conditon after the lastest else if, before else

jkroepke (changes requested):

charts/grafana/templates/_config.tpl
michaelruigrok: `hasKey` will cause a runtime error since `$value` would be a string and not a map. In that case I would have to wrap everything before the `else` in a `kindIs "map"` check. Do you still want me to change this? Or did you have another solution I've missed?

charts/grafana/templates/_config.tpl
```diff
@@ -54,7 +57,10 @@ grafana.ini: |
...
+{{ $key }}: |
+  {{- tpl $value $root | nindent 2 }}
+{{- else if (hasKey $value "file") }}
 {{ $key }}:
 {{- toYaml ( $files.Get $value.file ) | nindent 2 }}
```
michaelruigrok, line 65: @jkroepke This line should be indented, right?

jkroepke:
Closed - see https://github.com/grafana/helm-charts/issues/4087

PR #3748, opened 2025-06-12
[tempo] fix port 3200 alignment

KyriosGN0:
in https://github.com/grafana/helm-charts/pull/3745 i forgot that port 3100 is set also in the k8s service and the tempo config itself, this PR fixes it


mapno (approved):
Thank you 🙏

PR #3756, opened 2025-06-17
[grafana] Add default values for HTTProute backendRefs

### LINKED ISSUE #3755: ArgoCD sync status for HTTPRoute
We are deploying the Grafana Helm chart with using ArgoCD with httproute configured.

The helm Chart does not have the default values included in spec.rules.backendRefs, therefore ArgoCD will run into a sync issue.

erickuiper:
fixes #3755

Included the default values for group and kind.


jkroepke:
could you explain the reason?

erickuiper:
We are using ArgoCD to deploy this chart. Kubernetes adds these default
values after the installation, as these values are not part of the chart
ArgoCD will report an out of sync error.

Op di 17 jun 2025, 11:05 schreef Jan-Otto Kröpke ***@***.***>:

Sheikh-Abubaker (approved):

PR #3769, opened 2025-06-25
[grafana] Use grafana_com configuration name

makkes:
`grafana_com` has been the preferred name for some time now.

See
https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/#grafana_net for further justification.

> [!IMPORTANT]
> This is a breaking change for the chart and I'm not sure how these are handled in this repo. I'd be happy for any assistance if I missed documenting it properly or doing anything else that's necessary for breaking changes.


jkroepke (approved):
LGTM

PR #3771, opened 2025-06-25
[grafana] bump kiwigrid/k8s-sidecar image to 1.30.5

KyriosGN0:
Signed-off-by: AvivGuiser <avivguiser@gmail.com>

### DIFF
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -959,7 +959,7 @@ sidecar:
     # -- The Docker registry
     registry: quay.io
     repository: kiwigrid/k8s-sidecar
-    tag: 1.30.0
+    tag: 1.30.5
     sha: ""
   imagePullPolicy: IfNotPresent
   resources: {}
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 9.2.7
+version: 9.2.8
 appVersion: 12.0.2
 kubeVersion: "^1.8.0-0"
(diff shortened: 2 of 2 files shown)


jkroepke (approved):
LGTM

charts/grafana/values.yaml
```diff
@@ -959,7 +959,7 @@ sidecar:
     # -- The Docker registry
```
Sheikh-Abubaker, line 959: I'm having a limited scope over here, however I believe shouldn't this comment be [RedHat](https://quay.io/repository/kiwigrid/k8s-sidecar?tab=tags) registry instead of Docker ?

Sheikh-Abubaker: @jkroepke what are your thoughts ?

jkroepke: As I know, quay.io does not have such low rate limits as docker have

jkroepke (approved):

charts/grafana/values.yaml
Sheikh-Abubaker: @jkroepke, Thanks for the context and I am in sync on using quay.io instead of docker, however the comment is kind of misleading since docker registry isn't being used here, therefore my suggestion is to just change the comment to:
```suggestion
    # -- The Red Hat registry
```

Sheikh-Abubaker: That makes sense! Thanks for explaining!

booleanbetrayal:
@Sheikh-Abubaker / @KyriosGN0 - It looks like this commit broke dashboard loading. I've experienced this, as have others. See also: https://github.com/kiwigrid/k8s-sidecar/issues/400#issuecomment-3017699662

PR #3775, opened 2025-06-27
[grafana] Add additional default value for HTTRoute backend

erickuiper:
image

We deploy this chart using ArgoCD and the application remains out-of-sync due to a missing default value.

This PR aligns with the [httproute spec](https://gateway-api.sigs.k8s.io/reference/spec/#backendref)

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 9.2.8
+version: 9.2.9
 appVersion: 12.0.2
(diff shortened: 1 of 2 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 9.2.8
+version: 9.2.9
```
Sheikh-Abubaker, line 3:
```suggestion
version: 9.3.0
```

Sheikh-Abubaker (changes requested):

PR #3820, opened 2025-07-25
[tempo-distributed] fix missing namespace field

fengxsong:
fix namespace field missing issue when helm release is managed by ArgoCD.


fengxsong:
> @fengxsong, you can just tell argocd to create the namespace when the app is created, this is how it works for me

@KyriosGN0 we also use `helmfile` to manage releases. when a single helmfile contains multiple releases in different namespaces, the `spec.destination.namespace` parameter won't work.

PR #3821, opened 2025-07-28
[tempo-distributed] Fixing Tempo Registry Colcase Order

portswigger-amason:
Problem: Component-specific registry overrides weren't working despite being documented as overriding global registry.

  Root Cause: Template helper function had incorrect priority order in coalesce:
  `{{- $registry := coalesce .global.registry .component.registry .tempo.registry -}}`

  Fix: Changed priority to put component registry first:
  `{{- $registry := coalesce .component.registry .tempo.registry .global.registry -}}``

  Result:
  - Component registry now properly overrides global registry as documented
  - Fully backwards compatible - when component registry is null, falls back to global
  - Enables clean configuration like memcached.image.registry: public.ecr.aws

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.46.2
+version: 1.46.3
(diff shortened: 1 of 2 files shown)


portswigger-amason:
Could I please get some help in getting this merged. I would like to roll out the changes.

PR #3826, opened 2025-07-30
[tempo-distributed] Pass Namespace to Tempo Autoscaling

msgSend:
Some of the Tempo HPA templates are missing namespace attributes in the metadata which causes them to be deployed in the wrong namespace if Helm isn't used to directly install it into the desired namespace.

This just adds the namespace to all of them so that they will show up in the manifests.


Sheikh-Abubaker (commented):
Closing in favor of https://github.com/grafana/helm-charts/pull/4009

PR #3835, opened 2025-08-04
You should be able to configure empty tenant deletion

jacobcrawford:
The configuration options **empty_tenant_deletion_enabled** and **empty_tenant_deletion_age** have been supported for over a year in Tempo, but the helm chart did not expose a configuration option in the values file.


jacobcrawford:
@Sheikh-Abubaker can you take a look at this small PR? 😄

PR #3852, opened 2025-08-11
[grafana-sampling] add receiver parameters

rlankfo:
This adds the ability to set grpc server keepalive parameters to otlp receivers used in the sampling chart.


charts/grafana-sampling/values.yaml
```diff
@@ -28,8 +28,7 @@ sampling:
     # -- Percentage of successful requests to sample.
     percentage: 10
   failedRequests:
-    # -- Toggle sampling failed requests.
-    sample: false
+    # -- Toggle sampling failed requests. sample: false
```
petewall, line 31: was this intentional?

petewall (approved):

PR #3864, opened 2025-08-21
[grafana] Update _pod.tpl to override container name 

### LINKED ISSUE #3861: [Grafana] NameOverride ignored for container name
**Summary**

When nameOverride is specified in values, the deployment/pod name is correctly overridden, but the container name within the pod continues to use the chart name, creating an inconsistency.

**Current Behavior**

Deployment name: Uses grafana.name helper which respects nameOverride
Container name: Hardcoded to {{ .Chart.Name }} in _pod.tpl

This results in mismatched naming when nameOverride is configured.

**Expected Behavior**

Container name should match the deployment name when nameOverride is specified, maintaining consistency across Kubernetes resources.

**Proposed Solution**

- Update `_pod.tpl` container naming to `- name: {{ default .Chart.Name .Values.nameOverride  }}` to match the `grafana.name` helper convention or alternatively - name: {{ include "grafana.name" .  }}

brianjacksondev:
Closes #3861

Quick fix to standardize the naming of the primary container within the grafana pod using the `grafana.name` named template

### DIFF
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -967,7 +967,7 @@ containers:
       {{- toYaml . | trim | nindent 6 }}
       {{- end }}
 {{- end}}
-  - name: {{ include "grafana.name" . }}
+  - name: grafana
     {{- $registry := .Values.global.imageRegistry | default .Values.image.registry -}}
     {{- if .Values.image.sha }}
     image: "{{ $registry }}/{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}@sha256:{{ .Values.image.sha }}"
--- a/charts/grafana/templates/_pod.tpl
+++ b/charts/grafana/templates/_pod.tpl
@@ -967,7 +967,7 @@ containers:
       {{- toYaml . | trim | nindent 6 }}
       {{- end }}
 {{- end}}
-  - name: {{ .Chart.Name }}
+  - name: {{ include "grafana.name" . }}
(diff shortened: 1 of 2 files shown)


charts/grafana/templates/_pod.tpl
```diff
@@ -967,7 +967,7 @@ containers:
       {{- toYaml . | trim | nindent 6 }}
       {{- end }}
 {{- end}}
-  - name: {{ .Chart.Name }}
+  - name: {{ include "grafana.name" . }}
```
jkroepke, line 970: Whats the use-case? From my point of view, the even the hard-coded name grafana is fine.

jkroepke: Currently, I don't see any benefits, only down sides from end-user experiences.

But I'm open to change `{{ .Chart.Name }}` to just plain `grafana` as hard-coded value, similar how all other containers are named.

brianjacksondev: So this probably is more bothersome for my specific implementation of this chart.

I have a wrapper-style helm chart that pulls in this grafana chart through a helm dependency, and I give that chart an alias. The alias then defines the ` {{ .Chart.Name }}` (and therefore the container name) and I use nameOverride to reset it back to "grafana". 

The only resource that remains inconsistent is the container name - the deployments, configmaps, and other helm resources all incorporate the "grafana.name" helper (which uses the nameOverride); the only outlier I see is the primary container name itself. 

I would also be open to setting the container name to plan `grafana`; but would still prefer the usage of the helper - Let me know what you decide and I'll update accordingly

PR #3865, opened 2025-08-22
[pdc-agent] feat: support for pod service account

MarcStorm:
Support for running pod with a non-default service account.

Backwards compatible.


Sheikh-Abubaker:
@MarcStorm sure! I'll have a look

PR #3866, opened 2025-08-22
feat[pdc-agent]: support for readOnlyRootFilesystem

MarcStorm:
Allows to run on a read only root filesystem.

Disabled by default to be backwards compatible


Sheikh-Abubaker (commented):
Hey @MarcStorm, Thanks for your contribution, would you mind updating the docs ? so that we could proceed to merge this!

PR #3891, opened 2025-09-03
[grafana] Enable helm tpl for alert files

waterdrag0n:
File-based alert configurations don't support Helm templating, while inline alerts do.
Added this commit to enable file-based alerts with Helm templates like inline alerts.


charts/grafana/templates/_config.tpl
```diff
@@ -55,8 +55,8 @@ grafana.ini: |
 
 {{- range $key, $value := .Values.alerting }}
 {{- if (hasKey $value "file") }}
-{{ $key }}:
-{{- toYaml ( $files.Get $value.file ) | nindent 2 }}
+{{ $key }}: |
+{{- tpl ($files.Get $value.file) $root | nindent 2 }}
```
jkroepke, line 59: this is a breaking change, isn't it?

If the file contains somethink like

```
    apiVersion: 1
    contactPoints:
      - orgId: 1
        name: Slack channel
        receivers:
          - uid: default-receiver
            type: slack
            settings:
              # Webhook URL to be filled in
              url: ""
              # We need to escape double curly braces for the tpl function.
              text: '{{ template "default.message" . }}'
              title: '{{ template "default.title" . }}'
```

helm tries to evaluate the alertmanager templating.

jkroepke (changes requested):

charts/grafana/templates/_config.tpl
waterdrag0n: @jkroepke Thank you for the review. I understand your concern about the potential template conflicts, and I understand it's valid.

However, I'm confused about something:
The same issue already exists in the current implementation. (if the whole alert is written on values.yaml)
Looking at the current code, alerts defined directly in `values.yaml` already use the `tpl` function, which means they're already subject to the exact same template syntax conflicts you mentioned.

So i have 2 questions:
1. Is this current behavior for values-based alerts intentional? If template conflicts are problematic, how do values-based alerts currently handle this issue? Or am I missing something about how they're handled?

2. Should alert configuration behavior really depend on where it's defined? We're talking about the same logical configuration (alert rules, contactpoints...etc), just stored in different places - why should they behave differently?

If template conflicts are truly a breaking change concern, we should address it consistently across both methods.
Having identical alert configurations behave differently based on their location seems inconsistent.
Unless there's some defensive logic in the values-based approach that I'm missing? Thank you.

jkroepke: Please add an upgrade grade (https://github.com/grafana/helm-charts/blob/d5d7d3fd1c734d12675bd19a1f4a0a50294f5ede/charts/grafana/README.md?plain=1#L32) here

waterdrag0n: @jkroepke Added the upgrade guide. Please review, thanks!

charts/grafana/README.md
```diff
@@ -53,6 +53,11 @@ to `global.imageRegistry`. If you were not previously setting `global.image.regi
 is required on upgrade. If you were previously setting `global.image.registry`, you will
 need to instead set `global.imageRegistry`.
 
+### To 10.0.0
+
+Static alerting resources now support Helm templating. If you are using the `alerting` configuration,
+double curly braces must be escaped using `{{ "{{" }}` syntax to prevent unintended template processing.
```
jkroepke, line 59:
```suggestion
Static alerting resources now support Helm templating. This means that alerting resources loaded from external files (`alerting.*.files`) are now processed by the Helm template engine.

If you already use template expressions intended for Alertmanager (for example, `{{ $labels.instance }}`), these must now be escaped to avoid unintended Helm evaluation. To escape them, wrap the braces with an extra layer like this:

`{{ "{{" }} $labels.instance {{ "}}" }}`

This ensures the expressions are preserved for Alertmanager instead of being rendered by Helm.
```

jkroepke (approved):
LGTM

Issue #3867, opened 2025-08-25
Helm chart fails to render due to quoted variable in deployment


heytrav:
I've been seeing the following error in ArgoCD as well as in the CLI when I try to deploy our helm chart:

```
Error: YAML parse error on grafana/templates/deployment.yaml: error converting YAML to JSON: yaml: line 99: did not find expected key
```

I think I've isolated the problem to when helm is attempting to create the deployment (I marked the problematic line with `#ERROR`):

```
---
# Source: grafana/templates/deployment.yaml

apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-grafana
  namespace: monitoring
  labels:
    helm.sh/chart: grafana-9.3.5
    app.kubernetes.io/name: grafana
    app.kubernetes.io/instance: test
    app.kubernetes.io/version: "12.1.0"
spec:
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/name: grafana
      app.kubernetes.io/instance: test
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        helm.sh/chart: grafana-9.3.5
        app.kubernetes.io/name: grafana
        app.kubernetes.io/instance: test
        app.kubernetes.io/version: "12.1.0"
      annotations:
       .
       .
  
    spec:
      .
      .    
      .

      containers:
        - name: grafana-sc-alerts
          image: "quay.io/kiwigrid/k8s-sidecar:1.30.3"
          imagePullPolicy: IfNotPresent
          env:
            - name: METHOD
              value: WATCH
            - name: LABEL
              value: "grafana_alert"
            - name: FOLDER
              value: "/etc/grafana/provisioning/alerting"
            - name: RESOURCE
              value: "both"
            - name: NAMESPACE
              value: ""ALL""                      # ERROR
            - name: REQ_USERNAME
              valueFrom:
                secretKeyRef:
                  name: test-grafana
                  key: admin-user
            - name: REQ_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: test-grafana
                  key: admin-password
            - name: REQ_URL
              value: http://localhost:3000/api/admin/provisioning/alerting/reload
            - name: REQ_METHOD
              value: POST
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:


```

This appears to have been added in commit d64971f7.

discostur:
I just created a PR with a fix for that problem.

PR #3892, opened 2025-09-03
Fix double quote in NAMESPACE (3867)

discostur:
- fixes https://github.com/grafana/helm-charts/issues/3867

**Description:**
This PR fixes a YAML parsing error that occurs when rendering the Grafana Helm chart with sidecar searchNamespace values.

**Problem:**
When using searchNamespace: ["ALL"] or similar values in sidecar configurations, the template rendering produced invalid YAML with double quotes like value: ""ALL"", causing the error:

```bash
Error: YAML parse error on grafana/templates/deployment.yaml: error converting YAML to JSON: yaml: line 99: did not find expected key
```

**Solution:**
Updated the template syntax in _pod.tpl for all sidecar NAMESPACE environment variables from:

```bash
value: "{{ tpl (. | join ",") $root }}"
```
to
```bash
value: {{ tpl (. | join ",") $root | quote }}
```


jkroepke:
@discostur it may avoid the YAML error, but does solve the root cause.

1. The values is changed then to `"\"ALL\""` which can't be understand by the sidecar config.

discostur:
@jkroepke you are right - i'm sorry! i tested with an old version of the helm chart ... it's already fixed: 

https://github.com/grafana/helm-charts/commit/dff3f1062d3c18014eedc9825ce4ec4102a70ab2

Bug was introduced with 9.3.5 https://github.com/grafana/helm-charts/commit/d64971f74b13054e4c2566c4f90ae40017ab2514 but already fixed with 9.3.6. 

I will just closes this PR and report back in the original issue.

PR #3893, opened 2025-09-04
[tempo-distributed] Remove mem-ballast-size-mbs flags

### LINKED ISSUE #2887: Deprecate mem-ballast-size-mbs hardcoded args in Tempo distributed
Once Tempo is using Go v1.20+, the hardcoded arg `mem-ballast-size-mbs` in Tempo distributed chart templates isn't needed anymore. So the recommended way is using the environment variable `GOMEMLIMIT` instead.

E.g.:
https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/templates/ingester/statefulset-ingester.yaml#L62

cablespaghetti:
Fixes #2887

Removes mem-ballast-size-mbs flags that are hardcoded into the chart. If I understand correctly this is something that is not useful for modern versions of Go, and instead users can optionally set a GOMEMLIMIT environment variable depending on the resources they have given to each container.

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.55.0
+version: 1.56.0
 appVersion: 2.9.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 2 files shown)


joe-elliott:
@mattdurham we tried this recently internally and saw increased resource usage correct?

Sheikh-Abubaker (approved):

QuentinBisson:
@cablespaghetti thank you for the contribution. Could you please check check out the resource usage of the tempo components before and after this change to check if we need to update the README?

PR #3909, opened 2025-09-20
Adding commonLabels for tempo-distributed HC. With this we'll be able to add labels to all K8S resources the chart is creating.

tincho-burgos:
Enabling the tempo-distributed HelmChart to have commonLabels that will be added to all K8S created.

Signed-off-by: Martin Burgos <martin.burgos@qlik.com>

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -8,6 +8,8 @@ global:
     # Example:
     # pullSecrets: [ my-dockerconfigjson-secret ]
     pullSecrets: []
+  # -- Adding common labels to all K8S resources
+  commonLabels: {}
   # -- Overrides the priorityClassName for all pods
   priorityClassName: null
   # -- configures cluster domain ("cluster.local" by default)
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -93,6 +93,13 @@ app.kubernetes.io/managed-by: {{ .ctx.Release.Service }}
 {{- end -}}
 
 {{/*
+Common labels to add to all resources
+*/}}
+{{ - define "tempo.commonLabels" -}}
+{{- toYaml .Values.commonLabels | nindent 0 }}
+{{- end -}}
+
+{{/*
 Simple service selector labels
 */}}
 {{- define "tempo.selectorLabels" -}}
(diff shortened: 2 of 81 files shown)


tincho-burgos:
Can you please take a look at it? @Sheikh-Abubaker @BitProcessor @faustodavid @swartz-k
Rebasing main, bump chart and ran helm-docs

charts/tempo-distributed/values.yaml
```diff
@@ -8,6 +8,8 @@ global:
     # Example:
     # pullSecrets: [ my-dockerconfigjson-secret ]
     pullSecrets: []
+  # -- Adding common labels to all K8S resources
+  commonLabels: {}
```
Sheikh-Abubaker:
```suggestion
  # -- Adding custom labels to all K8S resources
  customLabels: {}
```

QuentinBisson:
@Sheikh-Abubaker in that case, would it not be better to define global.labels and global.podLabels instead of customLabels? This is a breaking change though

QuentinBisson:
First of, this would make this chart use the same patterns for label as the Loki and Mimir helm charts which provides the global.podLabels and global.podAnnotations fields.

Those are useful, specifically in the context of subcharts because those are also passed around.

For instance, I would find it useful to add tempo culture as a subchart. Following this global pattern would ensure that subcharts could use those labels as well

Sheikh-Abubaker:
@tincho-burgos Here's the trailing space, just get rid of it to fix failing space
https://github.com/grafana/helm-charts/blob/6434e0321a6ba7477e312f0594571b42d6d8f3ff/charts/tempo-distributed/values.yaml#L13C39-L14C3

tincho-burgos:
@Sheikh-Abubaker Done! Apologies for that, I missed the `excluding pods` part

tincho-burgos:
@Sheikh-Abubaker , I have added an if to bypass the error if it's empty.

`==> Linting charts/tempo-distributed
Error:  templates/: template: tempo-distributed/templates/query-frontend/service-query-frontend.yaml:7:8: executing "tempo-distributed/templates/query-frontend/service-query-frontend.yaml" at <include "tempo.labels" (dict "ctx" . "component" "query-frontend")>: error calling include: template: tempo-distributed/templates/_helpers.tpl:93:16: executing "tempo.labels" at <.Values.global.commonLabels>: nil pointer evaluating interface {}.global`

tincho-burgos:
@Sheikh-Abubaker it should work now.

Sheikh-Abubaker (approved):
LGTM!

PR #3914, opened 2025-09-25
[rollout-operator] Handle fullnameOverride in webhooks self-signed-cert

yetisage:
Add handling of fullnameOverride to the webhooks self-signed-cert, so the SAN matches the service name.


tcp13equals2 (approved):
LGTM

charleskorn (approved):

PR #3922, opened 2025-10-06
[grafana] Grafana Operator configuration support

Baarsgaard:
In the Grafana-Operator project we've have seen users having some difficulty configuring Grafana created via Helm.
As this is a documentation problem as well as a usability problem, we're now looking to close that gap.

One of the ideas is making it more convenient for users to get started with existing Grafana instances!
If you believe this absolutely does not belong here, you're more than welcome to decline this PR 😄.

I tried following the existing conventions I found in other templates, namely `deployment.yaml`, `_pod.tpl`, `service.yaml`, and `servicemonitor.yaml`, but I've likely overlooked something?

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.0.0
+version: 10.0.1
 appVersion: 12.1.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.
--- a/charts/grafana/README.md
+++ b/charts/grafana/README.md
@@ -284,6 +284,11 @@ This ensures the expressions are preserved for Alertmanager instead of being ren
 | `serviceMonitor.labels`                   | Labels for the servicemonitor passed to Prometheus Operator      |  `{}`                                |
 | `serviceMonitor.scrapeTimeout`            | Timeout after which the scrape is ended       | `30s`                                                   |
 | `serviceMonitor.relabelings`              | RelabelConfigs to apply to samples before scraping.     | `[]`                                      |
+| `grafanaOperator.enabled`                 | If true, a Grafana CR is created for a grafana operator                 | `false`     |                 |
+| `grafanaOperator.labels`                  | Additional Labels for `Grafana` CR `instanceSelector`s to match against | `{}`        |                 |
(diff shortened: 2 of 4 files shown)


Baarsgaard:
I did write an example for someone using the `kube-prometheus-stack` chart:
```yaml
grafana:
  extraObjects:
    # Will fail if the Grafana CRD is not installed
    - apiVersion: grafana.integreatly.org/v1beta1
      kind: Grafana
      metadata:
        name: "{{ .Release.Name }}-grafana"
        namespace: "{{ .Release.Namespace }}"
        labels:
          dashboards: grafana # Customize the labels if you plan to have multiple unique instance
      spec:
        external:
          url: "http://{{ .Release.Name }}-grafana.{{ .Release.Namespace }}.svc.cluster.local"
          adminUser:
            name: "{{ .Release.Name }}-grafana"
            key: admin-user
          adminPassword:
            name: "{{ .Release.Name }}-grafana"
            key: admin-password
```

As is shown, it's possible to use a subset of Helm templating in the values file, but it's limited to [Helm builtin objects](https://helm.sh/docs/chart_template_guide/builtin_objects/), mainly `.Release`, `.Chart`, and `.Capabilities`, but without flow control(`if` conditions).

But this is _only_ valid if the configuration is (near) default values, as compared to what is possible by upstreaming a proper template into the Grafana chart.

Which is why we wanted to see if you're open to the contribution.

jkroepke:
https://github.com/grafana/helm-charts/blob/b672274878f23b6817e406922775fbfee6352a1a/charts/grafana/templates/extra-manifests.yaml#L3

You can also write like


```
grafana:
  extraObjects:
    # Will fail if the Grafana CRD is not installed
    - |
      apiVersion: grafana.integreatly.org/v1beta1
      kind: Grafana
      metadata:
        name: "{{ .Release.Name }}-grafana"
        namespace: "{{ .Release.Namespace }}"
        labels:
          dashboards: grafana # Customize the labels if you plan to have multiple unique instance
      spec:
        external:
          url: "http://{{ .Release.Name }}-grafana.{{ .Release.Namespace }}.svc.cluster.local"
          adminUser:
            name: "{{ .Release.Name }}-grafana"
            key: admin-user
          adminPassword:
            name: "{{ .Release.Name }}-grafana"
            key: admin-password
```

to use if conditions.

PR #3927, opened 2025-10-10
[tempo-distributed] Add zone specific annotations and podAnnotations for ingester statefulsets

sringel:
Add zone specific annotations and podAnnotations for ingester statefulsets.


QuentinBisson:
@Sheikh-Abubaker can you take a look?

PR #3929, opened 2025-10-11
[synthetic-monitoring-agent] feat: add ability to enable features by environment variable

### LINKED ISSUE #3578: Allow users to specify features for synthetic-monitoring-agent
For the `synthetic-monitoring-agent`, a set of values can be provided to enable additional features of the agent. Currently it doesn't appear that this is possible as we can't pass in additional args or env vars to set this.

This is particularly impactful as if you haven't checked the docs, it's easy to miss that the `adhoc` feature is required in order to enable running the `Test` check from the Grafana UI.

tylerauerbeck:
Resolves #3578 

Allows user to enable features by environment variable

### DIFF
--- a/charts/tempo-distributed/Chart.yaml
+++ b/charts/tempo-distributed/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo-distributed
 description: Grafana Tempo in MicroService mode
 type: application
-version: 1.51.0
+version: 1.51.1
 appVersion: 2.9.0
 engine: gotpl
 home: https://grafana.com/docs/tempo/latest/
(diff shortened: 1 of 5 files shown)


Sheikh-Abubaker:
could you please elaborate your use case ?

tylerauerbeck:
Currently if you deploy the helm chart as is, when you go to run the `test` check of your synthetic check, it fails with something resembling a timeout error. However, what's actually causing this is that the `adhoc` feature isn't enabled. This change would allow you to enable the `adhoc` feature (or any other feature found [here](https://grafana.com/docs/grafana-cloud/testing/synthetic-monitoring/set-up/set-up-private-probes/#available-feature-flags)) when deploying the chart.

PR #3935, opened 2025-10-13
[tempo-distributed] Add action to the servicemonitor relabling for the cluster label

stuarta:
Currently the servicemonitor template is missing the action on the `targetLabel: cluster` relabelings.

Whilst this is valid, k8s adds in the default `action: replace` when the servicemonitor is deployed.

If you use ArgoCD to deploy tempo, with the servicemonitors enabled, this results in a delta between the desired state and the live state. Thus argo is constantly in a sync loop, trying to remediate the on-cluster state.

This PR adds action: replace to the servicemonitor template, allowing argocd to successfully apply the desired configuration.


QuentinBisson (approved):

QuentinBisson:
@stuarta Can you check the DCO? Your commits need to be signed

PR #3948, opened 2025-10-16
[grafana-mcp] use grafana/mcp-grafana instead of the docker image

KyriosGN0:
the new image registry is versioned, which means better version control of the mcp image, i bumped the chart version to v1 as its breaking change imo

### DIFF
--- a/charts/grafana-mcp/Chart.yaml
+++ b/charts/grafana-mcp/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana-mcp
-version: 0.2.1
+version: 0.2.0
 appVersion: 0.7.6
 kubeVersion: "^1.8.0-0"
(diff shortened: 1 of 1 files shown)


jkroepke:
> i bumped the chart version to v1 as its breaking change imo

Yes, it's an breaking change but breaking changes are allowed without bump the major, if major is below v1.

charts/grafana-mcp/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: grafana-mcp
-version: 0.1.4
-appVersion: latest
+version: 0.2.1
```
jkroepke:
```suggestion
version: 0.2.0
```

QuentinBisson (approved):

PR #3958, opened 2025-10-27
[rollout-operator] allow the image registry to be set

SAdams601:
Add an `image.registry` value so that the docker registry can be customized.


charts/rollout-operator/values.yaml
```diff
@@ -12,6 +12,7 @@ revisionHistoryLimit: 10
 
 image:
   repository: grafana/rollout-operator
+  registry: ""
```
charleskorn: Why is this needed? Couldn't the registry be added to `repository` instead? (eg. `us.gcr.io/foo/blah/rollout-operator`)

PR #3962, opened 2025-10-29
[grafana] Fix using `GF_SECURITY_ADMIN_USER__FILE`

### LINKED ISSUE #518: Error deploying when only GF_SECURITY_ADMIN_USER__FILE is set
**What happened:**
When I tried to deploy the chart with the following values:
  ```
adminUser: ""
  adminPassword: ""
  env:
    GF_SECURITY_ADMIN_USER__FILE: /path/to/user/file
    GF_SECURITY_ADMIN_PASSWORD__FILE: /path/to/password/file
```

The chart still expected the values populated in the secret created by the `.Values.admin.existingSecret` key, even on cases where it's not being created. 

**Why it happened:**
This happens because the conditional clauses regarding the GF_SECURITY_ADMIN_USER env var initialization are different from the ones in GF_SECURITY_ADMIN_PASSWORD.

**GF_SECURITY_ADMIN_USER**

`{{- if and (not .Values.env.GF_SECURITY_ADMIN_USER) (not .Values.env.GF_SECURITY_DISABLE_INITIAL_ADMIN_CREATION) }`

**GF_SECURITY_ADMIN_PASSWORD**

`{{- if and (not .Values.env.GF_SECURITY_ADMIN_PASSWORD) (not .Values.env.GF_SECURITY_ADMIN_PASSWORD__FILE) (not .Values.env.GF_SECURITY_DISABLE_INITIAL_ADMIN_CREATION) }}`

This misconfiguration can be found in `templates/_pod.tpl`, line 316 and 323. Any reason why they're different? If there's no reason, I can happily create a PR to fix this issue.

dguendisch:
Fixes #518 

As found in https://github.com/grafana/helm-charts/issues/518, using `GF_SECURITY_ADMIN_USER__FILE` to specify the admin user name to be read from a file rather than from an env variable, is currently broken because the generated pod will still try to populate an env var `GF_SECURITY_ADMIN_USER` from a secret.

The affected file `_pod.tpl` is already respecting the cases for the admin key, i.e. if a user is using `GF_SECURITY_ADMIN_PASSWORD__FILE`, then `_pod.tpl` will skip defining an env var that doesn't read from a file (`GF_SECURITY_ADMIN_PASSWORD`).

This PR adds the same logic to skip env var creation of `GF_SECURITY_ADMIN_USER` when the helm values already define `GF_SECURITY_ADMIN_USER__FILE`.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.1.3
+version: 10.1.4
 appVersion: 12.2.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.


jkroepke:
> though I'm not sure what your policy with incrementing the version is, let me know if you like to have it differently

personal feeling.

jkroepke (approved):
LGTM

Sheikh-Abubaker (approved):
@dguendisch Thanks for your contribution! LGTM!

PR #3963, opened 2025-10-29
[tempo-distributed] drop jaeger-ui related ports and config completely

zoomoid:
This no longer works, https://github.com/grafana/tempo/pull/3840 removed jaeger from tempo-query, changing with it the port that the server running in the container listens to:

- fixes config file used for running tempo-query; pointed to the config file of the main container, which contains config that the tempo-query pod cannot and does not parse anyway
- remove obsolete jaeger-related ports in favour of default port `7777` of tempo-query. Note that the container port isn't configurable right now. If you opt to set `ADDRESS` via `queryFrontend.query.extraEnvs` to something different, you cannot access the server anymore from outside the pod.
- remove conditional clause in toggle for `queryFrontend` Ingress. This no longer works for accessing Jaeger UI. Instead, you can repurpose the ingress, e.g. for configuring Grafana data sources
- change default of queryFrontend.service.port to HTTP port of query-frontend; the jaeger-related port previously used in here no longer works
- Remove tempo-query/jaeger-metrics-related service monitor. `tempo-query` doesn't export any metrics, there's nothing to scrape there, so this can be dropped completely in favour of the already existing service monitor for query-frontend's HTTP port

most of these changes probably also need to be reflected in the non-distributed charts as well, but I'm awaiting feedback here first.

### DIFF
--- a/charts/tempo-distributed/values.yaml
+++ b/charts/tempo-distributed/values.yaml
@@ -1005,8 +1005,8 @@ queryFrontend:
     # -- Docker image tag for the query-frontend image. Overrides `tempo.image.tag`
     tag: null
   service:
-    # -- Port of the query-frontend service
-    port: 3200
+    # -- Port of the tempo-query component in the queryFrontend service
+    port: 7777
     # -- Annotations for queryFrontend service
     annotations: {}
     # -- Labels for queryFrontend service
--- a/charts/tempo-distributed/templates/query-frontend/service-query-frontend.yaml
+++ b/charts/tempo-distributed/templates/query-frontend/service-query-frontend.yaml
@@ -29,7 +29,7 @@ spec:
       {{- end }}
     {{- if .Values.queryFrontend.query.enabled }}
     - name: tempo-query
-      port: 7777
+      port: {{ .Values.queryFrontend.service.port }}
       targetPort: tempo-query
       protocol: TCP
       appProtocol: grpc


charts/tempo-distributed/templates/query-frontend/ingress-query-frontend.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.queryFrontend.query.enabled .Values.queryFrontend.ingress.enabled -}}
+{{- if .Values.queryFrontend.ingress.enabled -}}
```
QuentinBisson: Is there any reason to change this?

charts/tempo-distributed/values.yaml
```diff
@@ -1006,7 +1006,7 @@ queryFrontend:
     tag: null
   service:
     # -- Port of the query-frontend service
-    port: 16686
+    port: 3200
```
QuentinBisson: Should it not be 7777?

zoomoid: okay, so getting back to this, the `service.port` properties are nested under `queryFrontend`, which has the static `3200` port for (i guess) just metrics via HTTP. The property's description also refers to *query-frontend*, not *tempo-query*, so this whole field seems a bit misplaced. I don't know if it's actually needed for something, but if it should just rectify the broken old port to the new (gRPC) port of tempo-query, we can set it to `7777` in the values and use that value in the template, sure.

However I think the confusion I had here is really indicative of drift of the upstream `tempo-query` service and the helm chart's use of it.

For now, I'll set this to 7777 and carry that to the service template.

PR #3984, opened 2025-11-01
[enterprise-logs] Mark chart as deprecated

jkroepke:


Sheikh-Abubaker:
@jkroepke what if we remove the minio dependency completely or set the version to something that actually exists ? anyways the chart is being deprecated so I don't really think it matters ?

jkroepke:
Maybe @dehaansa can help here.

(ref https://github.com/grafana/helm-charts/pull/3982)

PR #3989, opened 2025-11-05
[rollout-operator] Always set server-tls.self-signed-cert.dns-name

kimxogus:
In mimir helm chart with helm release name `mimir`, this should be `mimir-rollout-operator.NAMESPACE.svc`, but it's always `rollout-operator.NAMESPACE.svc` unless `.Values.fullnameOverride` is set.
This broke mimir's rollout operator and upgrading any statefulsets in the namespace.

### DIFF
--- a/charts/rollout-operator/Chart.yaml
+++ b/charts/rollout-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: rollout-operator
 description: "Grafana rollout-operator"
 type: application
-version: 0.37.0
+version: 0.37.1
 appVersion: v0.32.0
 home: https://github.com/grafana/rollout-operator
 kubeVersion: ^1.10.0-0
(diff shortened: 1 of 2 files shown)


dimitarvdimitrov:
> I've also removed default cert name defaults to chart fullname. Current default `certification` is too general name for this.

won't this break the existing installations? From what I saw the rollout-operator creates the certificate with that name itself. But this also means we need _some_ secret name. I'm not sure how this will work if you change the secret name during an upgrade.

At the very least, let's do this in a separate PR. Right now I think the priority should be to get the rollout-operator working correctly in mimir-distributed 6.0.0 and the secret name is only a cosmetic change

dimitarvdimitrov:
> i'm happy to merge this if you undo the secret name change. i think it's unnecessary for a bugfix PR

CI is failing because of the same change. The generated role is now invalid

kimxogus:
@dimitarvdimitrov Thank you for quick review and merge. I opened a separate PR #3990

PR #3990, opened 2025-11-05
[rollout-operator] Replace default webhooks.selfSignedCertSecretName with fullname

kimxogus:
Certificate secret's default name `certificate` is too general name, so I made it chart's fullname to match with the other resources.

Changing cert secret name doesn't break existing installations. It will recreate rollout-operator pod and cert secret will be recreated as well.

### DIFF
--- a/charts/rollout-operator/templates/deployment.yaml
+++ b/charts/rollout-operator/templates/deployment.yaml
@@ -53,7 +53,7 @@ spec:
           - -kubernetes.namespace={{ .Release.Namespace }}
           {{- if .Values.webhooks.enabled }}
           - -server-tls.enabled=true
-          - -server-tls.self-signed-cert.secret-name={{ .Values.webhooks.selfSignedCertSecretName | default (include "rollout-operator.fullname" . )  }}
+          - -server-tls.self-signed-cert.secret-name={{ include "rollout-operator.secretName" . }}
           - -server-tls.self-signed-cert.dns-name={{ include "rollout-operator.fullname" . }}.{{ .Release.Namespace }}.svc
           {{- end }}
           {{- range .Values.extraArgs  }}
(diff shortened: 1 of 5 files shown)


charts/rollout-operator/values.yaml
```diff
@@ -107,4 +107,4 @@ webhooks:
   # -- Validating and mutating webhook failure policy. `Ignore` or `Fail`.
   failurePolicy: "Fail"
   # -- Secret resource name for the TLS certificate to be used with the webhooks
```
dimitarvdimitrov: can you clarify what the default is if this value is empty?

charts/rollout-operator/templates/deployment.yaml
```diff
@@ -53,7 +53,7 @@ spec:
           - -kubernetes.namespace={{ .Release.Namespace }}
           {{- if .Values.webhooks.enabled }}
           - -server-tls.enabled=true
-          - -server-tls.self-signed-cert.secret-name={{ .Values.webhooks.selfSignedCertSecretName }}
+          - -server-tls.self-signed-cert.secret-name={{ .Values.webhooks.selfSignedCertSecretName | default (include "rollout-operator.fullname" . )  }}
```
dimitarvdimitrov: can `{{ .Values.webhooks.selfSignedCertSecretName | default (include "rollout-operator.fullname" . )  }}` maybe go in a named template in _helpers.tpl? that way we don't have to remember to update it in multiple places next time

dimitarvdimitrov (commented):
I left a couple of small comments

Perhaps my major question would be if this change would require manual upgrade path. If i understand correctly, the rollout operator would just ignore the existing secret and create a new one. There will be an orphaned secret in the namespace, but that shouldn't cause problems. Did i get this right? (it may be worth including these details in the PR description, since we don't have a changelog)

PR #4010, opened 2025-11-18
[loki-distributed] fix: Add missing metadata namespace

DrDJIng:
I have here added missing metadata namespaces for namespaced resources.

I think I got them all, and I think they are all namespaced resources, so let me know if I made a mistake.


DrDJIng:
Is there anything else I need to do to get this merged?

PR #4018, opened 2025-11-24
[pdc-agent] Expose token secret key name in values

holgerjh:
This PR adds the key `tokenSecretKey` to the `values.yaml`. It defines the lookup key used in the token secret defined by `tokenSecretName` and allows for a more dynamic configuration.


dafydd-t (dismissed):
Thanks for the PR!

- Can you bump the chart version please?
- After a merge / rebase & a reorder of `tokenSecretKey` & `tokenSecretName` in the README (the linting errors), I think we are good to go 👍

charts/pdc-agent/templates/deployment.yaml
```diff
@@ -58,7 +58,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ required "A secret name for the PDC agent token is required" .Values.tokenSecretName }}
-                  key: token
+                  key: {{ required "A secret key for the PDC agent token is required" .Values.tokenSecretKey | quote }}
```
asimpson:
```suggestion
                  key: {{ default "token" .Values.tokenSecretKey | quote }}
```

Would make upgrading work, e.g. `token` becomes the `tokenSecretKey` for existing users.

asimpson (changes requested):
1 more change, thanks for your patience 😎

asimpson (approved):
👍 now we're cooking!

beejeebus (approved):

PR #4025, opened 2025-11-25
[grafana] enforce explicit privileged false on grafana containerSecurityCo…

blame19:
…ntext

While the default settings are already secure, some linting/sec tools will throw warning if privileged: false is not explicitly set (see: [kubeaudit](https://github.com/Shopify/kubeaudit)). 
This small change doesn't affect grafana's capabilities in the slightest, but fixes a small irritant for the teams aiming for 'clean' security context audits.


jkroepke (approved):
LGTM

Issue #4045, opened 2025-12-09
when set as init containers, alerts and notifier sidecars have hardcoded `method`


sylvainOL:
It prevents them to be used as sidecar containers

Issue #4033, opened 2025-12-02
grafana: stray u in init-sc-dashboard startupProbe


bootc:
See:
https://github.com/grafana/helm-charts/blame/3c02be4de19c654ee4bc61849ec2d3d1fe03c857/charts/grafana/templates/_pod.tpl#L444

The stray trailing `u` causes template rendering failures and should be removed.

The problem was introduced in #3949.

QuentinBisson:
It was fixed with https://github.com/grafana/helm-charts/pull/4047

Issue #4031, opened 2025-12-01
[grafana] sidecar does not load datasources in version 10.3.0


mladBlum:
After updating the Grafana Helm chart from 10.2.0 to 10.3.0, datasource provisioning done by the kiwigrid/k8s-sidecar stopped working. The datasources are still correctly copied into the grafana pod, but they do not appear in Grafana anymore. Dashboard provisioning works fine.

I saw that the kiwigrid/k8s-sidecar was updated to 2.1.2. So I downgraded it manually, and the datasources appeared again in Grafana.

The logs of the datasource sidecar look like this:

```
{"time": "2025-12-01T16:44:20.627241+00:00", "level": "INFO", "msg": "Starting collector"}
{"time": "2025-12-01T16:44:20.628500+00:00", "level": "INFO", "msg": "No folder annotation was provided, defaulting to k8s-sidecar-target-directory"}
{"time": "2025-12-01T16:44:20.628861+00:00", "level": "INFO", "msg": "Loading incluster config..."} 
{"time": "2025-12-01T16:44:20.629657+00:00", "level": "INFO", "msg": "Unique filenames will not be enforced."}
{"time": "2025-12-01T16:44:20.629880+00:00", "level": "INFO", "msg": "5xx response content will not be enabled."}
{"time": "2025-12-01T16:44:20.630136+00:00", "level": "INFO", "msg": "Performing initial list-based sync before starting watch."}                                                                                                                      
{"time": "2025-12-01T16:44:20.630315+00:00", "level": "INFO", "msg": "Skipping initial request to external endpoint."} 
{"time": "2025-12-01T16:44:20.630488+00:00", "level": "INFO", "msg": "Loading incluster config..."}
{"time": "2025-12-01T16:44:20.631590+00:00", "level": "INFO", "msg": "Performing list-based sync on secret resources: {'namespace': 'analytic'}"}
{"time": "2025-12-01T16:44:20.698447+00:00", "level": "INFO", "msg": "Writing /etc/grafana/provisioning/datasources/mongodb.yaml (binary)"}
{"time": "2025-12-01T16:44:20.706552+00:00", "level": "INFO", "msg": "Loading incluster config..."}             
{"time": "2025-12-01T16:44:20.711725+00:00", "level": "INFO", "msg": "Performing list-based sync on configmap resources: {'namespace': 'analytic'}"}
[...]

PR #4046, opened 2025-12-09
[grafana] Make sidecar containers working

### LINKED ISSUE #4039: [grafana] Sidecars use the same health probe port which causes conflicts with more than 1 sidecar enabled
In the implementation of native sidecars each loader sidecar is default configured with the same health probe port. With health probes and native sidecar enabled this causes a failure to start due to ports being in use. Ideally each sidecar has it's own non-conflicting health probe port configuration as the default, for example 8081, 8082, 8083, 8084. Startup probe examples should also be modified to use the corresponding default health port.

ref: https://github.com/grafana/helm-charts/pull/3949

https://github.com/grafana/helm-charts/blob/4036bdc6f0de3eecd25de4d4265c14e4ef4eeb2f/charts/grafana/templates/_pod.tpl#L141-L144

https://github.com/grafana/helm-charts/blob/4036bdc6f0de3eecd25de4d4265c14e4ef4eeb2f/charts/grafana/templates/_pod.tpl#L230-L233

https://github.com/grafana/helm-charts/blob/4036bdc6f0de3eecd25de4d4265c14e4ef4eeb2f/charts/grafana/templates/_pod.tpl#L368-L371

https://github.com/grafana/helm-charts/blob/4036bdc6f0de3eecd25de4d4265c14e4ef4eeb2f/charts/grafana/templates/_pod.tpl#L457-L460

sylvainOL:
When trying to use sidecars containers as init containers with `restartPolicy` set to `Always` (a.k.a. [native sidecar](https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/)), several small issues are occuring:

* every sidecar containers tries to bind the same port unless you set a startingProbe (#4039)
* dashboard init container as a `u` in its definition (#4033)
* `METHOD` is hardcoded for alerts and notifier containers, preventing them to be usable in sidecar mode (#4045)

This PR resolves the three issues by:

* setting a default health port different for each sidecar
* removing the not needed `u`
* allowing to choose `METHOD` on alerts and notifier containers when `restartPolicy` is defined and set to `Always`

Please note that the behaviour of all 4 sidecar containers is not exactly homogeneous and I just tried to replicate their behavior of what's actually done when started as regular container

closes #4039, #4033, #4045 and possible #4031

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.4.3
+version: 10.5.0
 appVersion: 12.3.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.


QuentinBisson (approved):

sdesbure:
How sh*t you're right :/

`default` values works only if the end of  path doesn't exists or is null but here `.Values.sidecar.alerts.startupProbe` is nil by default :( and so we have a attempt to retrieve a key into a nil dict and it it makes the error you see

I'll make a fix tomorrow for that

Sorry

PR #4067, opened 2025-12-30
[grafana] Set proper server.domain in grafana.ini even if Gateway API is used

kofuk:
In the current grafana chart, there is an issue where grafana.ini's server.domain is not set correctly when using route instead of ingress. This PR resolves this issue by referencing the hostnames value configured in route.

### DIFF
--- a/charts/grafana/values.yaml
+++ b/charts/grafana/values.yaml
@@ -80,12 +80,12 @@ deploymentStrategy:
 readinessProbe:
   httpGet:
     path: /api/health
-    port: 3000
+    port: grafana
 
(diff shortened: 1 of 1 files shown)


charts/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.4.1
+version: 10.4.2
```
jkroepke:
```suggestion
version: 10.4.3
```

jkroepke (approved):

PR #4068, opened 2026-01-05
[grafana] Update kiwigrid/k8s-sidecar tag to 2.2.1

jnoordsij:
The v2 release of this image came paired with a very significant memory usage increase (see https://github.com/kiwigrid/k8s-sidecar/issues/462), which has subsequently been fixed (see https://github.com/kiwigrid/k8s-sidecar/pull/490) and released (see https://github.com/kiwigrid/k8s-sidecar/releases/tag/2.2.0).

Therefore I propose to update the tag to use the latest release.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.5.0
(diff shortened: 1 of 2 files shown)


jnoordsij:
@QuentinBisson done!

PR #4069, opened 2026-01-05
[tempo] Make `metrics_generator` storage paths configurable

gjed:
Make the `metrics_generator` storage paths configurable through `values.yaml`, fixing hardcoded `/tmp` paths that caused metrics loss on pod restarts.

**Why**
TraceQL metrics with `local-blocks` processor require persistent storage paths, but hardcoded `/tmp` paths lose data on pod restart.

**Changes**
Adds configurable `storage`, `traces_storage`, `processor`, and `registry` sections under `tempo.metricsGenerator` matching official Tempo documentation structure.

**Backward Compatibility**
Backward compatible: defaults preserve existing behavior, all new options are optional.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.4.3
+version: 10.5.5
 appVersion: 12.3.1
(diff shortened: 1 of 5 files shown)


Sheikh-Abubaker (approved):
LGTM!

PR #4080, opened 2026-01-08
[grafana] fix(unified): set default index_path in chart

RafaelPaulovic:
Environment variables take precedence over config files, setting `index_path` this way guarantees that the ephemeral storage path is used even when a custom config file is provided.

The setting remains configurable via the `grafana.ini` section.

Follow-up from: https://github.com/grafana/helm-charts/pull/4073


jkroepke (approved):
LGTM

PR #4082, opened 2026-01-12
[grafana] Allow customization of hostUsers

jcpunk:
Kubernetes has supported the `hostUsers` flag for a while now. This PR adds the flag to grafana, but leaves it as the current default of `true`.

Folks investigating user namespaces will benefit from having a simple place to flip this on/off.

### DIFF
--- a/charts/tempo/Chart.yaml
+++ b/charts/tempo/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: tempo
 description: Grafana Tempo Single Binary Mode
 type: application
-version: 1.24.2
+version: 1.24.3
 appVersion: 2.9.0
 engine: gotpl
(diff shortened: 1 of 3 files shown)


jkroepke (commented):
Let align with https://github.com/prometheus-community/helm-charts/pull/6474

charts/grafana/values.yaml
```diff
@@ -1502,6 +1503,8 @@ imageRenderer:
   # image-renderer deployment serviceAccount
   serviceAccountName: ""
   automountServiceAccountToken: false
+  # image-renderer deployment hostUsers
+  hostUsers: NULL
```
jkroepke:
```suggestion
  hostUsers: ~
```

jkroepke (approved):
LGTM

PR #4083, opened 2026-01-12
[tempo] Add full volumeClaimTemplate spec when persistence is enabled

a1kemist:
When deploying the Tempo helm chart with argocd and persistence is enabled, there is a diff between the templated resources and what is applied on the server:

image

---

This causes argocd to mark the resource as out of sync and can issues with sync policies. This PR adds the full spec when persistence is enabled in order to avoid the out of sync issue.


Sheikh-Abubaker (approved):
Looks reasonable!

Issue #3833, opened 2025-08-03
Existing PVC


Schwitzd:
If set `persistence.volumeName`, the default pvc is still created ended up with two volumes.

I'm using the latest Grafana chart

PR #4086, opened 2026-01-13
[grafana] fix: Use PV name when supplied, rather than looking up existing PVC v…

aimmac23:
…olume name (#3833)

This appears to have been a regression introduced when adding PVC volume name lookups. The "volumeName" field is still in values.yaml, leading to confusion.

### DIFF
--- a/charts/grafana/Chart.yaml
+++ b/charts/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: grafana
-version: 10.5.6
+version: 10.5.7
 appVersion: 12.3.1
 kubeVersion: "^1.8.0-0"
 description: The leading tool for querying and visualizing time series and metrics.


charts/grafana/templates/pvc.yaml
```diff
@@ -25,9 +25,13 @@ spec:
   resources:
     requests:
       storage: {{ .Values.persistence.size | quote }}
+  {{- if .Values.persistence.volumeName }}
+  volumeName: {{ .Values.persistence.volumeName }}
+  {{- else }}
   {{- if and (.Values.persistence.lookupVolumeName) (lookup "v1" "PersistentVolumeClaim" (include "grafana.namespace" .) (include "grafana.fullname" .)) }}
   volumeName: {{ (lookup "v1" "PersistentVolumeClaim" (include "grafana.namespace" .) (include "grafana.fullname" .)).spec.volumeName }}
   {{- end }}
+  {{- end }}
```
jkroepke:
```suggestion
```

PR #4118, opened 2026-01-29
[pdc-agent] add podAnnotations support

wjmolina:
Add `podAnnotations` support to allow setting annotations on the pod template. This is useful for service discovery, metrics scraping (for example, Prometheus or Grafana Agent annotations), and other pod-level metadata.


asimpson (approved):
Thanks for the PR! 👍

asimpson:
@wjmolina could you run `helm docs` and commit any changes? This keeps the docs in sync.

PR #4133, opened 2026-02-20
[rollout-operator] add deployment annotations + allow not to create crds

DrFaust92:
add deployment annotations + allow not to create crds

for clusters with instances of rollout operator, manage CRDs once


DrFaust92:
Good point! You're right that Helm doesn't template files in the `crds/` directory.

However, a common pattern used by other charts (e.g., [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/charts/crds)) is to use a **CRDs subchart** that:
1. Keeps plain CRD YAMLs in `crds/crds/` for initial install (no templating)
2. Uses a pre-upgrade Job in `crds/templates/upgrade/` to handle CRD updates on subsequent upgrades

I can refactor this PR to follow a similar approach — moving the CRDs into `templates/` with the conditional flag, or adopting the subchart pattern. What would you prefer?

tcp13equals2:
> Good point! You're right that Helm doesn't template files in the `crds/` directory.
> 
> However, a common pattern used by other charts (e.g., [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/charts/crds)) is to use a **CRDs subchart** that:
> 
> 1. Keeps plain CRD YAMLs in `crds/crds/` for initial install (no templating)
> 2. Uses a pre-upgrade Job in `crds/templates/upgrade/` to handle CRD updates on subsequent upgrades
> 
> I can refactor this PR to follow a similar approach — moving the CRDs into `templates/` with the conditional flag, or adopting the subchart pattern. What would you prefer?

Hi @DrFaust92 - sorry for a delayed reply. Re your suggested refactor, sure I would be interested in seeing how this looks. Thank you!

DrFaust92:
@tcp13equals2 force-pushed a refactor — went with option 1 (move CRDs into `templates/`) since the chart only has 2 CRDs and the upgrade-job/subchart pattern felt like overkill here. Summary of changes:

- Both CRDs moved from `crds/` → `templates/` so Helm actually templates them.
- `crds.install` (default `true`) gates rendering for clusters that manage CRDs out of band (e.g. shared rollout-operator).
- `crds.keep` (default `true`) adds `helm.sh/resource-policy: keep` so CRDs survive `helm uninstall`.
- Deployment-level `annotations` added.
- Rebased onto current main, bumped chart to 0.48.0, regenerated README.

Verified `helm lint` clean and `helm template` produces the expected output for all four combinations of `crds.install` / `crds.keep`. Indentation issue you flagged on `replica-templates-...yaml` is gone now since both CRDs were rewritten.

DrFaust92:
Update — refactored again to use the **kube-prometheus-stack subchart pattern** as you suggested (without the upgrade Job, since rollout-operator only has 2 CRDs and the subchart pattern alone gives us the toggle):

- Added a `crds` subchart at `charts/rollout-operator/charts/crds/` containing plain (untemplated) CRD YAMLs in its own `crds/` dir.
- Parent chart declares it as a dependency with `condition: crds.enabled` (default `true`).
- Set `crds.enabled=false` to skip CRD installation entirely (subchart not loaded).
- `annotations` field added for deployment-level annotations.
- Bumped chart to 0.48.0; commit signed-off (DCO).
- Tweaked `.gitignore` to un-ignore `charts/rollout-operator/charts/` so the in-tree subchart sources can be committed (otherwise caught by the repo-wide `/charts/*/charts` rule for built tarballs).

**Non-breaking for existing installs**: their CRDs are already in cluster from the previous top-level `crds/` directory, and Helm leaves CRDs in `crds/` alone on upgrade. New users get them via the subchart `crds/` (same install-once behavior).

Verified `helm lint` clean and `helm template --include-crds` renders 2 CRDs by default and 0 with `--set crds.enabled=false`.

tcp13equals2:
Thanks @DrFaust92 - I will review this week. Thank you!

tcp13equals2:
@DrFaust92 - I did some testing with these changes today and it performs correctly.

Could you update the README to reflect these changes and I would also suggest adding some more commentary to the `charts/crds/values.yaml` explaining when `crds.enabled` should be toggled. The CRDs are always needed for the rollout-operator to run, but an operator may choose to disable this flag for certain reasons ...

Also note that there is a lint error.

DrFaust92:
> @DrFaust92 - I did some testing with these changes today and it performs correctly.
> 
> Could you update the README to reflect these changes and I would also suggest adding some more commentary to the `charts/crds/values.yaml` explaining when `crds.enabled` should be toggled. The CRDs are always needed for the rollout-operator to run, but an operator may choose to disable this flag for certain reasons ...
> 
> Also note that there is a lint error.

Done, waiting for CI

tcp13equals2 (approved):

PR #4158, opened 2026-04-22
fix(tempo-distributed): recreate job superfluous quotes, wrong rbac

cwrau:
This removes the superfluous and wrong quotes; kubectl doesn't accept them

Additionally; this fixes the RBAC needed for the job

### DIFF
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -389,7 +389,7 @@ spec:
             - pvc
             - --namespace={{ $newStatefulSet.metadata.namespace }}
             - {{ printf "%s-%s-%d" $template $newStatefulSet.metadata.name $index }}
-            - --type='json'
+            - --type=json
             - '-p=[{"op": "replace", "path": "/spec/resources/requests/storage", "value": "{{ $size }}"}]'
           {{- end }}
         {{- end }}
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -441,6 +441,7 @@ rules:
     {{- end }}
     verbs:
       - patch
+      - get
   {{- end }}
 ---
 apiVersion: rbac.authorization.k8s.io/v1
--- a/charts/tempo-distributed/templates/_helpers.tpl
+++ b/charts/tempo-distributed/templates/_helpers.tpl
@@ -430,7 +430,7 @@ rules:
       - delete
   {{- if $templates }}
   - apiGroups:
-      - v1
+      - ""
     resources:
       - persistentvolumeclaims
     resourceNames:


charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -430,7 +430,7 @@ rules:
       - delete
   {{- if $templates }}
   - apiGroups:
-      - v1
```
electron0zero, line 433: any reasons this `v1` was removed?

charts/tempo-distributed/templates/_helpers.tpl
```diff
@@ -441,6 +441,7 @@ rules:
     {{- end }}
     verbs:
       - patch
+      - get
```
electron0zero, line 444: also this, can you explain why this was added?

charts/tempo-distributed/templates/_helpers.tpl
cwrau: Now that you mention this; that's weird

I tried it and got 

```
persistentvolumeclaims "storage-loki-0" is forbidden: User "system:serviceaccount:monitoring:loki-recreate" cannot get resource "persistentvolumeclaims" in API group "" in the namespace "monitoring" 
```

so I thought pvc are in no apiGroup.

But now that I checked they really are in the `v1` group...

cwrau: Using the RBAC in this PR worked

charts/tempo-distributed/templates/_helpers.tpl
cwrau: Same log as https://github.com/grafana/helm-charts/pull/4158/changes#r3123193816; the patch action apparently needs a `get` beforehand

charts/tempo-distributed/templates/_helpers.tpl
cwrau: I tried this right now with a fresh SA and impersonation and get the same results;

apigroup `v1` is apparently wrong and `get` is required

cwrau:
```
$ k -n default --as system:serviceaccount:default:test patch pvc/test --type json -p '[{"op":"replace","path":"/spec/resources/requests/storage","value":"3Gi"}]' || k describe role test
Error from server (Forbidden): persistentvolumeclaims "test" is forbidden: User "system:serviceaccount:default:test" cannot get resource "persistentvolumeclaims" in API group "" in the namespace "default"
Name:         test
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources                  Non-Resource URLs  Resource Names  Verbs
  ---------                  -----------------  --------------  -----
  persistentvolumeclaims.v1  []                 [test]          [patch]
---
# I added the get permission
---
$ k -n default --as system:serviceaccount:default:test patch pvc/test --type json -p '[{"op":"replace","path":"/spec/resources/requests/storage","value":"3Gi"}]' || k describe role test
Error from server (Forbidden): persistentvolumeclaims "test" is forbidden: User "system:serviceaccount:default:test" cannot get resource "persistentvolumeclaims" in API group "" in the namespace "default"
Name:         test
Labels:       <none>
Annotations:  <none>
PolicyRule:
  Resources                  Non-Resource URLs  Resource Names  Verbs
  ---------                  -----------------  --------------  -----
  persistentvolumeclaims.v1  []                 [test]          [patch get]
---
# I changed the apiGroup to `""`
---
$ k -n default --as system:serviceaccount:default:test patch pvc/test --type json -p '[{"op":"replace","path":"/spec/resources/requests/storage","value":"3Gi"}]' || k describe role test
persistentvolumeclaim/test patched
```

PR #4163, opened 2026-05-19
Update update-helm-repo.yaml workflow to use grafana/shared-workflows/actions/create-github-app-token

alsoba13:
This is obviously a breaking change, as we are changing the `vault_repo_secret_name` secret by a `github_app_name` input, but all the consumers should upgrade anyway to fix their workflows.

Example of use:
https://github.com/grafana/pyroscope/pull/5171


zerok (dismissed):

narqo (approved):
+1 to unblock the reviews, while we are searching for the owner of DCO

alsoba13:
solved by #4164

PR #4168, opened 2026-05-22
fix(tempo): use tempo-query config for query frontend sidecar

noashinu:
The tempo-query sidecar in query-frontend was started with the main Tempo config:
-config=/conf/tempo.yaml

This is incorrect because tempo-query has its own generated configuration file:
/conf/tempo-query.yaml

The argument has been changed to:
-config=/conf/tempo-query.yaml

This ensures that the tempo-query container reads the configuration generated from
queryFrontend.query.config instead of the main Tempo configuration.


Sheikh-Abubaker (commented):
Hey @noashinu Thanks for your contribution, I recommend checking out the deprecation notice in this chart's [README](https://github.com/grafana/helm-charts/tree/main/charts/tempo-distributed) and feel free to open a similar PR on the community fork

PR #4169, opened 2026-05-26
Sign the index.yaml commit pushed to gh-pages

pracucci:
## Summary

Replace `cr index --push` with `cr index` (no push) + `planetscale/ghcommit-action`, so the resulting commit on `gh-pages` is signed by GitHub's web-flow key and shows as **Verified**.

## Why

`cr index --push` calls `git commit` + `git push` internally (see `pkg/git/git.go` in `helm/chart-releaser`), producing an unsigned commit. This breaks for consumers of this reusable workflow whose `gh-pages` branch has "Require signed commits" branch protection enabled, and weakens the audit/provenance story for any consumer that requires Verified commits across the artifact-publishing chain.

## How

- Reorder the checkouts so the gh-pages branch is checked out first, at the workspace root. `actions/checkout` with no `path:` empties the target directory before cloning, so doing it second would wipe the `source/` subdir checked out earlier. With it first, the source checkout into `path: source` happens into a sibling subdir and both coexist.
- Drop the now-unneeded `path: helm-charts` checkout (which previously held the default-branch clone that `cr index --push` worked from) and its `Configure Git for helm-charts` step. cr will still find `origin/gh-pages` because the workspace-root checkout is on that branch with full history.
- Drop `--push` from the `cr index` invocation so it only regenerates `index.yaml` locally at `${CR_INDEX_PATH}/index.yaml`; copy the file to the workspace-root `index.yaml` so it appears as a modification to the gh-pages working tree.
- Use `planetscale/ghcommit-action` with `file_pattern: index.yaml` to commit just that one file via `createCommitOnBranch`.

Token, permissions, and the source-tree checkout used for chart packaging are unchanged.

## Tracking the upstream improvement

Filed helm/chart-releaser#625 asking whether the `cr` maintainers would take a `--sign-commits` flag that does this same thing inside the tool. If that lands, this workflow can drop the manual file-copy + ghcommit-action and go back to one-line cr usage.

## Test plan

Tested end-to-end in https://github.com/pracucci/helm-charts-test (sandbox, no secrets):

1. **Sandbox set up** by cloning this PR's workflow into the test repo, replacing the three hardcoded `grafana/helm-charts` refs with `pracucci/helm-charts-test`.
2. **Tiny chart added** at `charts/test-chart/` (Chart.yaml, single ConfigMap template, values.yaml) plus matching `cr.yaml` / `ct.yaml` at the root.
3. **Caller workflow added** at `.github/workflows/release.yml` invoking the reusable workflow on `workflow_dispatch`, passing `helm_repo_token: ${{ secrets.GITHUB_TOKEN }}`.
4. **`gh-pages` seeded** as an orphan branch with a minimal `index.yaml`.
5. **Chart version bumped** so `ct list-changed` detects a diff vs the root commit (without a bump it sees nothing and the release job is skipped).
6. **Workflow triggered** with `gh workflow run release.yml` and watched green end-to-end.
7. **Bot commit verified** at https://github.com/pracucci/helm-charts-test/commit/ba6b2f439d on `gh-pages`: `verification.verified: true`, `reason: valid`; committer `GitHub <noreply@github.com>`, author `github-actions[bot]`; only `index.yaml` modified (+13/-2); content matches what `cr index` produced.

### Notable findings

- **`ct list-changed` saw no diff** initially because the chart was added in the same commit as `since` — fix: bump chart version on a follow-up commit.
- **`Parse Chart.yaml` step lost `source/`** because `actions/checkout` at the workspace root empties the target before cloning (regardless of `clean:`). Fix in this PR: reorder checkouts so gh-pages runs first; source then lands into the empty `source/` subdir untouched.

### DIFF
--- a/.github/workflows/update-helm-repo.yaml
+++ b/.github/workflows/update-helm-repo.yaml
@@ -154,6 +154,22 @@ jobs:
             echo "AUTHTOKEN=${APP_TOKEN}" >> $GITHUB_ENV
           fi
 
+      # The cr tool requires a checkout of the target repository so that it can `git worktree
+      # add origin/<pages-branch>`. We check the gh-pages branch out at the workspace root so
+      # that planetscale/ghcommit-action (which runs from $GITHUB_WORKSPACE) can later commit
+      # the regenerated index.yaml via the GitHub API.
+      #
+      # This step runs BEFORE the source checkout because actions/checkout with no `path:`
+      # empties the target directory before cloning, which would otherwise wipe source/.
+      - name: Checkout helm-charts gh-pages
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6
+        with:
+          fetch-depth: 0
+          repository: grafana/helm-charts
+          ref: gh-pages
+          token: ${{ env.AUTHTOKEN }}
+          persist-credentials: false
+
       - name: Checkout
         uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6 # zizmor: ignore[artipacked] without this ignore comment, zizmor would complain that persist-credentials is not explicitly set to false. We need it set to true (default) to be able to push the release tags later on in this workflow
         with:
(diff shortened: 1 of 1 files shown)


krajorama (approved):
LGTM

